Issue #342, opened 2015-01-13
Merge the remaining "And now I have the following committed files" steps


kevgo:
Merge them into the preceding "And now I have the following commits" steps. 

This is blocking on #341

PR #362, opened 2015-01-18
Remove unnecessary verification step

kevgo:
@allewun @charlierudolph 

Removes the last occurrences of the now unnecessary verification steps.

Implements #342


allewun:
:ship:

Issue #287, opened 2014-12-24
Initial configuration isn't robust


allewun:
* catching blank main branch name isn't working correctly
* can set a bogus main branch name

kevgo:
Hey @allewun can this ticket be closed now after your glorious PR last week got merged?

PR #372, opened 2015-01-28
Robust initial configuration

allewun:
Implements #287 

**Initial setup changes:**
* Ensures branch existence before assigning to main and non-feature
* Ensure non-feature branches can't be set to main branch
* Moved all configuration into `git town config --setup`
  * It allows configuration without running a command as a side-effect
* Show simple message if Git Town if not configured, and easily begin setup if desired. Ex:
```
> git sync
Git Town hasn't been configured for this repository.
Please run 'git town config --setup'.
Would you like to do that now? y/n

...
```

**Configuration changes:**
* Added `git town config --reset` to completely remove all Git Town configurations for a given repo

@kevgo @charlierudolph


features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,73 @@
...
+    When I run `git town config --setup` and enter "nonexistent"
+    Then I see
+      """
+      Please enter the name of the main dev branch (typically 'master' or 'development'):
+
+        Error
+        There is no branch named 'nonexistent'
+
+      """
+    And Git Town is still not configured for this repository
```
charlierudolph: I think this should be a different interface. If I enter a non existent branch, it displays an inline error and allows me to enter the name again.

```
Please enter the name of the main dev branch (typically 'master' or 'development'):
nonexistent # user types
error: There is no branch named 'nonexistent' # displayed in red
master # user types
# ...
```

features/git-town/configuration/configuration_reset.feature
```diff
@@ -0,0 +1,36 @@
...
+  As a user wishing to remove all Git Town settings from a repository
+  I want to be able to cleanly reset the Git Town configuration with a single command
+  So that I can continue working on actual work and stay productive
+
+
+  Scenario: everything is configured
+    Given I have configured the main branch name as "main"
+    And my non-feature branches are configured as "qa" and "staging"
+    When I run `git town config --reset`
+    Then I see "Your Git Town settings have been reset for this repository"
```
charlierudolph: Along the lines of successful commands produce no output, I would vote that this produces no output.

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,73 @@
...
+
+      Git Town supports non-feature branches like 'release' or 'production'.
+      These branches cannot be shipped and do not merge master when syncing.
+      Please enter the names of all your non-feature branches as a comma separated list.
+      Example: 'qa, production'
+
+        Error
+        There is no branch named 'nonexistent'
+
+      """
```
charlierudolph: Just letting you know, I want to change this interface so instead of entering multiple branches as a comma delimited string, just let the user add one branch per line and entering an empty line means the user is done. This can totally be done in a separate PR but may be simpler to just convert this over in this PR.

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,73 @@
+Feature: Initial configuration
+
+  Background:
+    Given I haven't configured Git Town yet
+
+
+  Scenario Outline: Running Git Town commands while Git Town is unconfigured
+    When I run `<COMMAND>` while allowing errors
+    Then I see
```
kevgo: Once Charlie's PR is in, please change this to

```cucumber
When I run `<COMMAND>`
Then I see the error

File: features/git-town/configuration/configuration_initial.feature [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+    | git kill           |
+    | git pr             |
+    | git prune-branches |
+    | git repo           |
+    | git ship           |
+    | git sync           |
+    | git sync-fork      |
+
+
+  Scenario: Enter non-existent main branch
```
kevgo: I think these are two unrelated features that warrant to be in their own files.

```cucumber
Feature: Automatically running the configuration wizard if Git Town is unconfigured


Feature: git town: configuration wizard enforces an existing branch for the main branch
  Scenario: valid branch provided
  Scenario: non-existing branch provided
  ...

File: features/git-town/configuration/configuration_initial.feature (hunk 1 above)
Comment (kevgo) [inline features/git-town/configuration/configuration_initial.feature]: You are right, @charlierudolph. But this is such an extreme edge case that happens only once over the lifetime of the Git Town installation that we don't have to over-engineer this. A simple exit with a clean error message and the possibility to re-run the last command on the command line should be enough here.

File: features/git-town/configuration/configuration_initial.feature [hunk 6]
```diff
@@ -0,0 +1,73 @@
...
+      Please enter the name of the main dev branch (typically 'master' or 'development'):
+
+        Error
+        There is no branch named 'nonexistent'
+
+      """
+    And Git Town is still not configured for this repository
+
+
+  Scenario: Enter valid main branch and non-existent non-feature branch
```
kevgo: Scenario: user enters a valid main branch and non-existing non-feature branch

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,73 @@
...
+    Given I have a branch named "master"
+    When I run `git town config --setup` and enter "master" and "nonexistent"
+    Then I see
+      """
+      Please enter the name of the main dev branch (typically 'master' or 'development'):
+
+      Git Town supports non-feature branches like 'release' or 'production'.
+      These branches cannot be shipped and do not merge master when syncing.
+      Please enter the names of all your non-feature branches as a comma separated list.
+      Example: 'qa, production'
```
kevgo: Let's wrap this paragraph in a regex, to make this test more robust. The particular wording of the explanation of feature branches doesn't matter here, and even the slightest syntax adjustment causes a false negative here. 

```cucumber
Then I see
  """
  Please enter the name of the main dev branch (typically 'master' or 'development'):

  .*

    Error
     There is no branch named 'nonexistent'

  """
```

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,73 @@
...
+
+        Error
+        There is no branch named 'nonexistent'
+
+      """
+    And the main branch name is now configured as "master"
+    And my non-feature branches are still not configured
+
+
+    Scenario: Enter valid main branch and valid non-feature branches
```
kevgo: Why is this scenario indented?

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,73 @@
...
+      Given I have branches named "dev" and "qa"
+      When I run `git town config --setup` and enter "dev" and "qa"
+      Then the main branch name is now configured as "dev"
+      And my non-feature branches are now configured as "qa"
+
+
+    Scenario: Enter valid main branch and invalid non-feature branches
```
kevgo: Unindent this scenario as well

features/git-town/configuration/configuration_reset.feature
```diff
@@ -0,0 +1,36 @@
+Feature: resetting the configuration
+
+  As a user wishing to remove all Git Town settings from a repository
+  I want to be able to cleanly reset the Git Town configuration with a single command
+  So that I can continue working on actual work and stay productive
```
kevgo: This user story makes little sense to me. Here is my take:

```cucumber
As a user having messed up the Git Town configuration beyond repair
I want an easy way to cleanly reset the Git Town configuration completely
So that I can start over and configure it right this time.

File: features/git-town/configuration/non_feature_branches/add_branch.feature [hunk 11]
```diff
@@ -24,6 +24,15 @@ Feature: add a branch to the non-feature branches configuration
       """
 
 
+  Scenario: adding a branch that is already set as the main branch
+    Given I have configured the main branch name as "staging"
+    When I run `git town non-feature-branches --add staging` while allowing errors
+    Then I see
+      """
+      error: 'staging' is already set as the main branch
```
kevgo: This is a cool feature!

features/support/run_helpers.rb
```diff
@@ -30,8 +30,8 @@ def git_town_command? command
 end
 
 
-def output_of command
-  run(command).out.strip
+def output_of command, allow_failures: false
+  run(command, allow_failures: allow_failures).out.strip
```
kevgo: All these changes are better reverted before you merge Charlies commits, otherwise you'll run into pretty substantial merge conflicts. Once his changes are in, I recommend you do a `git reset master`, then a re-commit of everything you want to keep, and reverting the rest.

features/support/town_helpers.rb
```diff
@@ -14,12 +14,12 @@ def delete_non_feature_branches_configuration
 
 
 def main_branch_configuration
-  output_of 'git config --get git-town.main-branch-name'
+  output_of 'git config --get git-town.main-branch-name', allow_failures: true
```
kevgo: revert

features/support/town_helpers.rb
```diff
@@ -14,12 +14,12 @@ def delete_non_feature_branches_configuration
...
 end
 
 
 def non_feature_branch_configuration
-  output_of 'git config --get git-town.non-feature-branch-names'
+  output_of 'git config --get git-town.non-feature-branch-names', allow_failures: true
```
kevgo: revert

src/helpers/configuration.sh
```diff
@@ -20,30 +20,17 @@ if [[ $@ =~ --bypass-automatic-configuration ]]; then
...
-    echo_error "Please try again."
-    exit_with_error
+# Show configuration prompt
+if [[ "$(is_git_town_configured)" == false ]]; then
+  echo "Git Town hasn't been configured for this repository."
+  echo "Please run 'git town config --setup'."
+  echo "Would you like to do that now? y/n"
+
+  read setup_git_town_now
+  if [[ "$setup_git_town_now" == y ]]; then
```
kevgo: Does the `y` not have to be wrapped in double quotes here?

src/helpers/configuration_helpers.sh
```diff
@@ -21,6 +21,9 @@ function add_non_feature_branch {
   elif [ "$(is_non_feature_branch "$branch_name")" = true ]; then
     echo_inline_error "'$branch_name' is already a non-feature branch"
     exit 1
+  elif [ "$branch_name" == "$main_branch_name" ]; then
+    echo_inline_error "'$branch_name' is already set as the main branch"
+    exit 1
```
kevgo: Boo on the use of magic numbers here. Use `exit_with_error` instead

src/helpers/configuration_helpers.sh
```diff
@@ -52,6 +69,22 @@ function get_configuration {
 }
 
 
+# Returns whether or not Git Town is configured
+function is_git_town_configured {
+  if [ -n "$main_branch_name" ]; then
```
kevgo: Shouldn't you also check that the non-feature branches are at least set to `''` or something here? What if I forget to provide non-feature branches. Does Git Town think everything is peachey, and just runs?

src/helpers/configuration_helpers.sh
```diff
@@ -70,6 +103,70 @@ function remove_non_feature_branch {
...
+  if [[ -z "$main_branch_input" ]]; then
+    echo_error_header
+    echo_error "You have not provided the name for the main branch."
+    echo_error "This information is necessary to run this script."
+    echo_error "Please try again."
+    exit_with_error
+  fi
+
+  ensure_has_branch "$main_branch_input"
+  store_configuration main-branch-name "$main_branch_input"
```
kevgo: This is really cool code. Nicely done!

src/helpers/configuration_helpers.sh
```diff
@@ -70,6 +103,70 @@ function remove_non_feature_branch {
...
+#
+function setup_configuration {
+  setup_configuration_main_branch
+  echo
+  setup_configuration_non_feature_branches
+}
+
+
+function setup_configuration_main_branch {
+  # Ask and store main-branch-name
```
kevgo: I would move the comment above the method

src/helpers/configuration_helpers.sh
```diff
@@ -70,6 +103,70 @@ function remove_non_feature_branch {
...
+    exit_with_error
+  fi
+
+  ensure_has_branch "$main_branch_input"
+  store_configuration main-branch-name "$main_branch_input"
+}
+
+
+function setup_configuration_non_feature_branches {
+  # Ask and store non-feature-branch-names
```
kevgo: Move the comment above the method

src/helpers/configuration_helpers.sh
```diff
@@ -70,6 +103,70 @@ function remove_non_feature_branch {
...
+  echo "Git Town supports non-feature branches like 'release' or 'production'."
+  echo "These branches cannot be shipped and do not merge $main_branch_name when syncing."
+  echo "Please enter the names of all your non-feature branches as a comma separated list."
+  echo "Example: 'qa, production'"
+  read non_feature_input
+
+  if [[ -n "$non_feature_input" ]]; then
+    ensure_has_branches "$non_feature_input" &&
+    ensure_valid_non_feature_branches "$non_feature_input"
```
kevgo: Since this line only executes if the previous one works, you should indent it with 2 spaces to mark it as dependent on the previous one.

src/helpers/configuration_helpers.sh
```diff
@@ -70,6 +103,70 @@ function remove_non_feature_branch {
...
+
+  if [ $? -eq 0 ]; then
+    store_configuration non-feature-branch-names "$non_feature_input"
+  fi
+}
+
+
+# Reset git town configuration for repository if flag is passed,
+# otherwise show the config
+function show_or_setup_or_reset_config {
```
kevgo: Better name: `run_config_operation`

src/helpers/git_helpers/branch_helpers.sh
```diff
@@ -56,6 +56,16 @@ function ensure_has_branch {
 }
 
 
+# Exits if any of the branches do not exist
```
kevgo: **does** not exist

kevgo:
This is pretty much good to go from my end if all the comments are addressed.

Sorry for the late review, I've been down under on RMS... :tired_face:

features/git-town/configuration/configuration_reset.feature
kevgo: Hmm, I could get behind that. No error message means command successful. If we don't trust the empty output of the command, how could we trust any output of it?

features/git-town/configuration/configuration_initial.feature
kevgo: Any particular reason for that @charlierudolph?

features/git-town/configuration/configuration_initial.feature
allewun: Agreed, would also like to change the wording of these prompts in a new PR soon

src/helpers/configuration.sh
allewun: @charlierudolph has previously recommended not quoting strings unless necessary (I still think it looks weird), but it works. Same as `if [ ... == true ] ...`

src/helpers/configuration_helpers.sh
allewun: magic

src/helpers/configuration_helpers.sh
allewun: good point

src/helpers/configuration_helpers.sh
allewun: I had some bash issues with these that I didn't quite work out yet. I'm not sure why the `ensure`s don't terminate immediately, hence the requirement that I compare the exit status `$?` in the next line.

Maybe something to do with the scope of the `exit_with_error` within `ensure_has_branch` within [`ensure_has_branches`](https://github.com/Originate/git-town/blob/aw-robust-config/src/helpers/git_helpers/branch_helpers.sh#L64).

features/git-town/configuration/configuration_initial.feature
charlierudolph: I disagree that is an "extreme edge case". This is the first interaction users will have with Git Town. I'd rather have this be fully functional and intuitive and not make me have to rerun commands. I can implement this in a separate PR if @allewun doesn't want to add it to this.

features/git-town/configuration/configuration_initial.feature
charlierudolph: Branch names can have commas in them and I think its more intuitive to enter one per line.

src/helpers/configuration_helpers.sh
charlierudolph: Part of this is exit_with_error adds a new line (a stupid implementation of exit_with_error by me). I will rename that function so exit_with_error only exits and we have another function for adding a newline and exiting with error

allewun:
@charlierudolph can you explain what this does?

```ruby
  expect(unformatted_last_run_output.strip).to include(message), %(
    ACTUAL
    ***************************************************
    #{@last_run_result.out.gsub '\n', "\n"}
    ***************************************************
    EXPECTED TO INCLUDE
    ***************************************************
    #{message.gsub '\n', "\n"}
    ***************************************************
  ).gsub(/^ {4}/, '')
```

charlierudolph:
So a normal expectation would just hae been
```ruby
expect(unformatted_last_run_output.strip).to include(message)
```
And this just adds a custom error message by passing the extra argument.

features/git-town/configuration/configuration_initial.feature
allewun: Let's save it for a new PR, don't want to clutter this one.

features/git-town/configuration/configuration_reset.feature
allewun: Ok (though I don't necessarily always agree with the convention).

features/git-town/configuration/configuration_reset.feature
allewun: Wouldn't the user just use `git town config --setup` if that was the reason? I agree my "So that..." feels weak though

allewun:
Ready for another review

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,50 @@
...
+    Given I haven't configured Git Town yet
+
+
+  Scenario: user enters non-existent main branch
+    When I run `git town config --setup` and enter "nonexistent"
+    Then I get the error
+      """
+        Error
+        There is no branch named 'nonexistent'
+      """
```
charlierudolph: These error checks can be done in one line
```cucumber
Then I get the error "There is no branch named 'nonexistent'"
```

features/step_definitions/run_steps.rb
```diff
@@ -14,12 +14,14 @@
 
 
 When(/^I run `(.+?)` and enter "(.+?)"$/) do |command, input|
-  @result = run command, input: input
+  inputs = Kappamaki.from_sentence(input)
+  @result = run command, inputs: inputs
```
charlierudolph: Can input stay singular if you just join the inputs with a newline?

features/step_definitions/town_steps.rb
```diff
@@ -46,3 +46,13 @@
...
 end
+
+
+Then(/^my non\-feature branches are still not configured$/) do
+  expect(non_feature_branch_configuration.split(',')).to be_empty
+end
+
+
+Then(/^Git Town is (?:no longer|still not) configured for this repository$/) do
+  expect(run('git config --get-regex git-town || true').out).to eql ''
```
charlierudolph: How about pushing this out into town helpers

```ruby
Then(/^Git Town is (?:no longer|still not) configured for this repository$/) do
  expect(git_town_configurationt).to be_empty
end

def git_town_configuration
  array_output_of 'git config --get-regex git-town || true'
end
```
Using array output as if returns multiple lines if thing are actually in there. Also keeps all the `|| true` limited to the town_helpers file.
```
# when configured
$ git config --get-regex git-town
git-town.main-branch-name master
git-town.non-feature-branch-names release
```

src/helpers/configuration.sh
```diff
@@ -20,30 +20,15 @@ if [[ $@ =~ --bypass-automatic-configuration ]]; then
...
+# Show configuration prompt
+if [[ "$(is_git_town_configured)" == false ]]; then
+  echo "Git Town hasn't been configured for this repository."
+  echo "Please run 'git town config --setup'."
+
+  if prompt_yn "Would you like to do that now? y/n"; then
+    echo
+    setup_configuration
+  else
+    exit 0
```
charlierudolph: Shouldn't this `exit_with_error` as we aren't doing anything because they refuse to setup git-town

src/helpers/configuration_helpers.sh
```diff
@@ -52,6 +69,22 @@ function get_configuration {
...
+    echo true
+  else
+    echo false
+  fi
+}
+
+
+# Remove all Git Town configuration
+function remove_all_configuration {
+  git config --remove-section git-town > /dev/null 2>&1
```
charlierudolph: What is the output you are redirecting?

src/helpers/terminal_helpers.sh
```diff
@@ -3,6 +3,19 @@
 # Helper methods for writing to the terminal.
 
 
+# http://stackoverflow.com/a/12202793/3418047
+function prompt_yn () {
+  while true; do
```
charlierudolph: Please name the argument
```sh
local prompt=$1
```

src/helpers/terminal_helpers.sh
```diff
@@ -3,6 +3,19 @@
...
+    read -p "$1 " yn
```
charlierudolph: Please name the argument
```sh
local prompt=$1
```

src/helpers/configuration_helpers.sh
allewun: If it doesn't exist, git prints an textual error:
```
> git config --remove-section git-town
fatal: No such section!
```

src/helpers/configuration.sh
allewun: It's not exactly an error, though I don't feel strongly about this

features/step_definitions/run_steps.rb
allewun: I'm not sure what you're asking

features/git-town/configuration/configuration_initial.feature
allewun: Cool

src/helpers/configuration.sh
charlierudolph: The command was not successful though as we aren't doing anything.

features/step_definitions/run_steps.rb
charlierudolph: Can you do the following instead?
```ruby
input = Kappamaki.from_sentence(input).join("\n")
@result = run command, input: input
```

allewun: It doesn't seem to work, but it also feels hackier. Why do it that way, when semantically we have multiple inputs?

charlierudolph: Just to avoid a loop if possible. I'm fine with it as is.

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,50 @@
...
+    And the main branch name is now configured as "master"
+    And my non-feature branches are still not configured
+
+
+  Scenario: user enters valid main branch and valid non-feature branches
+    Given I have branches named "dev" and "qa"
+    When I run `git town config --setup` and enter "dev" and "qa"
```
kevgo: How do you feel about making it more explicit which input belongs to which step? 

```cucumber
When I run `git town config --setup` and enter the main branch name "dev" and the non-feature branches "qa" 
```

This would also allow to change the interface of the config setup command, or the order in which branch names are queried, without having to make a shotgun refactoring of all these scenarios.

features/git-town/configuration/configuration_initial.feature
```diff
@@ -0,0 +1,50 @@
...
+    Then the main branch name is now configured as "dev"
+    And my non-feature branches are now configured as "qa"
+
+
+  Scenario: user enters valid main branch and invalid non-feature branches
```
kevgo: Scenario: user enters the existing main branch as feature branches

features/git-town/configuration/configuration_reset.feature
kevgo: Okay, what is the reason for this command to exist then? Why would I ever reset the Git Town configuration if not to provide new configuration right afterwards? One cannot use Git Town without configuration, right?

features/step_definitions/run_steps.rb
kevgo: I'd say that words like "sand", "water", or "input" have no meaningful plural. They imply multiple individual things.

features/step_definitions/town_steps.rb
kevgo: +1

features/support/run_helpers.rb
```diff
@@ -63,12 +63,12 @@ def result_has_shell_error? result
...
 
 
-def run_shell_command command, input
+def run_shell_command command, inputs
   result = OpenStruct.new(command: command, location: Pathname.new(Dir.pwd).basename)
   command = "#{shell_overrides}; #{command} 2>&1"
 
   status = Open4.popen4(command) do |_pid, stdin, stdout, _stderr|
-    stdin.puts input if input
+    inputs.each { |input| stdin.puts input }
```
kevgo: Let's rename `inputs` to `input` and make it a [casual API](http://blog.originate.com/blog/2014/06/26/casual-apis/).
* if a string is given, use the string
* if an array is given, use each string inside the array

src/helpers/configuration.sh
kevgo: Yeah, lets use a different error code here: 1 for command failed, 2 for user aborted. This allows scripts to distinguish these situations.

src/helpers/configuration_helpers.sh
```diff
@@ -45,6 +48,20 @@ function add_or_remove_non_feature_branches {
 }
 
 
+# Ensure that non-feature branches don't contain main branch
+function ensure_valid_non_feature_branches {
+  local branches=$1
+
+  split_string "$branches" ',' | while read branch; do
```
kevgo: Cool solution!

src/helpers/configuration_helpers.sh
kevgo: Other people will have the same question as the reviewer. Please add a comment that explains this motivation for redirecting the output here.

src/helpers/configuration_helpers.sh
```diff
@@ -70,6 +103,66 @@ function remove_non_feature_branch {
...
+function setup_configuration_non_feature_branches {
+  echo "Git Town supports non-feature branches like 'release' or 'production'."
+  echo "These branches cannot be shipped and do not merge $main_branch_name when syncing."
+  echo "Please enter the names of all your non-feature branches as a comma separated list."
+  echo "Example: 'qa, production'"
+  read non_feature_input
+
+  if [[ -n "$non_feature_input" ]]; then
+    ensure_has_branches "$non_feature_input" || exit_with_error
+    ensure_valid_non_feature_branches "$non_feature_input" || exit_with_error
```
kevgo: Cool error checking here

src/helpers/terminal_helpers.sh
```diff
@@ -3,6 +3,19 @@
 # Helper methods for writing to the terminal.
 
 
+# http://stackoverflow.com/a/12202793/3418047
```
kevgo: The externally linked site might go away in the future. Please also provide a summary of the comment here, and use the URL only as a link to read more about this issue, or as proof for your claims here.

kevgo:
This is a really cool change, and its pretty much good to go. @allewun lets finish this up and then :sheep: it!

src/helpers/terminal_helpers.sh
allewun: This was mainly for crediting the original author, is that necessary?

kevgo: No, that's not necessary. The copyright for that source code is actually with StackOverflow. So I suggest you create your own code based on the ideas outlined in the article, and remove the link.

features/git-town/configuration/configuration_reset.feature
allewun: Perhaps the user doesn't wish to use Git Town anymore, and doesn't want any remnants of it polluting the repo?

The reason I added this was actually to make it easier to test my changes regarding initial configuration.

kevgo: Okay, lets go with uninstalling Git Town configuration as the user value.

> As a user no longer using Git Town on a repository
> I want to be able to cleanly remove all Git Town configuration from my Git repo
> So that my repository is left in a clean state after the uninstallation.

features/git-town/configuration/configuration_initial.feature
allewun: done

features/support/run_helpers.rb
allewun: This seems to complicate things. We already have `Kappamaki.from_sentence` filtering the inputs for us.

allewun:
All comments addressed

features/step_definitions/run_steps.rb
```diff
@@ -72,8 +77,20 @@
 
 
 Then(/^I see$/) do |output|
+  actual = unformatted_last_run_output.strip
+  expect(actual).to eql "#{output}"
```
charlierudolph: Couldn't this just be
```ruby
expect(actual).to eql output
```

features/support/run_helpers.rb
```diff
@@ -63,17 +63,18 @@ def result_has_shell_error? result
 end
 
 
-def run_shell_command command, input
+def run_shell_command command, inputs
   result = OpenStruct.new(command: command, location: Pathname.new(Dir.pwd).basename)
   command = "#{shell_overrides}; #{command} 2>&1"
+  kill = inputs.pop && true if inputs.last == '^C'
```
charlierudolph: I'm confused how killing works. Can you please add a comment as to what this does? 
Also is the `&& true` necessary? Wouldn't kill just end up being `'^C'` which is truthy anyway.

charlierudolph:
Just that last couple comments. Then ready to :ship:

features/support/run_helpers.rb
allewun: If the last input is "^C", I set a `kill` flag that prevents the result from erroring. Basically we shouldn't have an error in this situation, since "the user" is choosing to abort the command.

This isn't a true process kill as you can tell. I *was* able to figure out how to kill the child process, but I couldn't get it to have any meaningful `result`.

PR #395, opened 2015-02-21
“Show me my repo” step

kevgo:
@charlierudolph @allewun 

While investigating bugs, I found this step useful at times.


allewun:
This is cool. Would it make sense to distinguish this step visually from normal steps, perhaps by making it all caps? Or surrounding the text with brackets?

kevgo:
I think that's not necessary. I halts the spec indefinitely, so it can never accidentally be left in a PR. It's really only meant to be in there for a few minutes. Making it all caps would make it hard to remember how to write it.

allewun:
:ship:

PR #408, opened 2015-02-28
Rename expected_steps to expected_commands

kevgo:
@charlierudolph @allewun 

More descriptive variable name. We are checking for commands here, not steps.


allewun:
:ship:

Issue #392, opened 2015-02-20
Clean up output options


kevgo:
```ruby
Then(/^the output begins with "(.*)"$/) do |output|
  expect(unformatted_last_run_output).to start_with output
end


Then(/^the output contains "(.*)"$/) do |output|
  expect(unformatted_last_run_output).to include output
end
```

seem redundant to 

```ruby
Then(/^I see "(.*)"$/) do |string|
  step 'I see', string
end

Then(/^I see$/) do |string|
  expect(unformatted_last_run_output).to include(string)
end
```

PR #410, opened 2015-02-28
Dry up output verification

kevgo:
@allewun @charlierudolph 

I think these tests are still specific enough.

Implements #392


allewun:
It's not exactly the same, but I'm fine with this cleanup. Maybe was too specific initially. :+1:

Issue #376, opened 2015-02-03
Rename run_command


kevgo:
@charlierudolph suggested:

* renaming `run_command_outside_git ` to `run_command`
* renaming `run_command` to `run_git_command` and `print_git_command`

kevgo:
blocked on #358

PR #411, opened 2015-02-28
Improve naming of command runners

kevgo:
@charlierudolph @allewun 

This came out of a code review.

Implements #376

### DIFF
--- a/src/helpers/terminal_helpers.sh
+++ b/src/helpers/terminal_helpers.sh
@@ -165,7 +165,7 @@ function run_git_command {
 
 
 # Run a command, prints command and output
-function run_command_outside_git {
+function run_command {
   local cmd="$*"
   echo_header "$cmd"
   eval "$cmd" 2>&1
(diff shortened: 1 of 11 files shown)


src/helpers/terminal_helpers.sh
```diff
@@ -157,7 +157,7 @@ function prompt_yn {
 
 
 # Run a command, prints command and output
-function run_command {
+function run_git_command {
   local cmd="$*"
   print_command "$cmd"
```
allewun: I think `print_command` should be renamed to `print_git_command` since it prepends the git branch.

src/helpers/terminal_helpers.sh
```diff
@@ -157,7 +157,7 @@ function prompt_yn {
 
 
 # Run a command, prints command and output
```
allewun: Update comment so it's not identical to that of `run_command`

allewun:
:ship: after addressing comments

src/helpers/terminal_helpers.sh
kevgo: Good catch. I gave the comments some love.

src/helpers/terminal_helpers.sh
kevgo: Good idea. Done.

PR #440, opened 2015-03-22
HoundCI support

kevgo:
@charlierudolph @allewun 

This is mostly a test right now. I think this would be very useful, in at least 10 of my PRs I ran into style issues that I didn't catch. Let's try this. Hopefully it allows us to use CI more efficiently. If it sucks, we can take it out again at any time.

implements #436

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,2 @@
+ruby:
+  config_file: .ruby-style.yml
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,10 @@ Style/MethodDefParentheses:
   EnforcedStyle: require_no_parentheses
 
 
+Style/StringLiterals:
+  EnforcedStyle: single_quotes
+
+
 Metrics/AbcSize:
   Max: 19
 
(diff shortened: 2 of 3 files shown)


allewun:
I don't understand why Rubucop isn't sufficient. If Hound is finding things that rubocop misses, isn't that just a configuration change in rubocop.yml?

kevgo:
Hound uses Rubocop. The problem is that if there is a Rubocop problem, you won't know until CircleCI fails. Then you have to go on Github, click through to the PR, click through to CircleCI, scroll down, and see what's up. Hound on the other hand tells you pretty instantly about style violations, as easily readable comments, that are also sent out as emails (at least to me). This mostly saves clickery and cognitive overhead.

allewun:
Hmm I see.. in my workflow I run the specs locally and to get immediate feedback, so Hound seems like it would be even slower. Good to go from my end though.

kevgo:
On my (6 year old) machine at home, the full spec runs > 10 min (since I have only 2 cores here), so I let the CI do all the testing work for me. I only run individual specs locally, or smaller blocks of specs. Hound will be a great help for this.

kevgo:
Apologies for merging this prematurely, but Hound is pretty disruptive if we keep using it unconfigured. I'm happy to do any changes mentioned here later, but for now I'm going to merge this PR so that we can see how it feels like if used properly.

PR #441, opened 2015-03-22
Update Ruby dependencies

kevgo:
@charlierudolph @allewun 

Just the regular dependency refresh to stay up to date


allewun:
:ship:

PR #452, opened 2015-03-25
Make "y" the default answer for Git Town prompts

allewun:
Resolves issue #443 

@kevgo @charlierudolph


features/git-town/configuration/configuration_automatic.feature
```diff
@@ -27,13 +27,31 @@ Feature: Automatically running the configuration wizard if Git Town is unconfigu
     | git sync-fork      |
 
 
+  Scenario Outline: Proceeding to configuration upon initial config prompt (using prompt's default)
```
kevgo: I don't understand the title for this scenario outline. Do you mean "Chosing the default option for the choice whether to run the set up wizard"?

features/git-town/configuration/configuration_automatic.feature
```diff
@@ -27,13 +27,31 @@ Feature: Automatically running the configuration wizard if Git Town is unconfigu
...
+    When I run `<COMMAND>` and enter "" and "^C"
+    Then I see "Git Town hasn't been configured for this repository."
```
kevgo: I think you should verify that you see the second step in the wizard here, which would prove that the wizard got started and you got past the initial question.

features/git-town/configuration/configuration_automatic.feature
```diff
@@ -27,13 +27,31 @@ Feature: Automatically running the configuration wizard if Git Town is unconfigu
...
+    And I see "Please specify the main dev branch"
```
kevgo: What exactly the question is is irrelevant here. What we really care about is that you got past the question whether you want to run the wizard, and see the first question of the wizard now. 

```cucumber
And I see the first question of the setup wizard
```

src/helpers/configuration.sh
```diff
@@ -24,7 +24,7 @@ fi
 if [[ "$(is_git_town_configured)" == false ]]; then
   echo "Git Town hasn't been configured for this repository."
   echo "Please run 'git town config --setup'."
-  echo -n "Would you like to do that now? [y/n] "
+  echo -n "Would you like to do that now?"
```
kevgo: I would keep it as "Would you like to do that now? [Y/n] "

features/git-town/configuration/configuration_automatic.feature
allewun: Yeah this title isn't the clearest, will fix that up

features/git-town/configuration/configuration_automatic.feature
allewun: I like that!

src/helpers/configuration.sh
allewun: I thought it made more sense to have the `[Y/n]` echoed in the `prompt_yn` function, since the text directly relates to the `case` statement's pattern matching. I don't feel that strongly about this though.

kevgo: The user should see `Y/n` somewhere, so that he knows that he has to just hit Enter to run the wizard. Where that happens I don't care.

features/git-town/configuration/configuration_automatic.feature
allewun: I'm not sure it's unnecessary, as there are 2 implicit steps here. 

1. the user types a git town command, hits enter, then sees an initial config prompt
2. then the user types "y" or "n" or "", hits enter, then sees the first question of the config

Buuut as I type this now, I see that it should probably be split up huh? One scenario outline that tests for the "Git Town hasn't been configured..." message. Another scenario outline that does what this one here does

features/step_definitions/town_steps.rb
```diff
@@ -56,3 +56,8 @@
 Then(/^Git Town is (?:no longer|still not) configured for this repository$/) do
   expect(git_town_configuration).to be_empty
 end
+
+
```
kevgo: Please `git sync` this branch to make these messages go away.

features/step_definitions/town_steps.rb
```diff
@@ -56,3 +56,8 @@
...
+Then(/^I see the first question of the configuration wizard$/) do
+  step %Q(I see "Please specify the main dev branch")
```
kevgo: Hound is right here. Good :poodle:

features/step_definitions/town_steps.rb
allewun: Done

src/helpers/configuration.sh
allewun: I think we're talking about different things. I had only moved the location of the Y/n in code, it would still show the user the same way. I mistakenly removed Y/n in the spec. Added that back.

features/step_definitions/town_steps.rb
allewun: is more appropriate :)

allewun:
@kevgo all comments addressed

src/helpers/configuration.sh
kevgo: Cool. Yeah, the spec should still show the Y/n thing.

features/git-town/configuration/configuration_automatic.feature
kevgo: Yes. I think we already have a spec that verifies that the config wizard comes up if the configuration is missing. At least we should have one. This particular spec only describes the default value for the first prompt. So it can assume that the config wizard works without testing it again, and just verifies that hitting Enter on the first prompt does the right thing.

features/git-town/configuration/configuration_automatic.feature
```diff
@@ -9,10 +9,43 @@ Feature: Automatically running the configuration wizard if Git Town is unconfigu
     Given I haven't configured Git Town yet
 
 
-  Scenario Outline: Proceeding to configuration upon initial config prompt
+  Scenario Outline: Seeing a configuration prompt when running a Git Town command
```
kevgo: Scenario Outline: All Git Town commands show the configuration wizard if running unconfigured

features/git-town/configuration/configuration_automatic.feature
```diff
@@ -9,10 +9,43 @@ Feature: Automatically running the configuration wizard if Git Town is unconfigu
...
+    | git kill           |
+    | git prune-branches |
+    | git pull-request   |
+    | git repo           |
+    | git ship           |
+    | git sync           |
+    | git sync-fork      |
+
+
+  Scenario Outline: Explicitly proceeding to configuration wizard upon seeing config prompt
```
kevgo: Scenario Outline: Starting the wizard by answering "y" to the question whether to start it

kevgo: You also need a scenario outline that specifies what happens if the user answers "n" here

features/git-town/configuration/configuration_automatic.feature
```diff
@@ -9,10 +9,43 @@ Feature: Automatically running the configuration wizard if Git Town is unconfigu
...
+    | git kill           |
+    | git prune-branches |
+    | git pull-request   |
+    | git repo           |
+    | git ship           |
+    | git sync           |
+    | git sync-fork      |
+
+
+  Scenario Outline: Implicitly proceeding to configuration wizard upon seeing config prompt
```
kevgo: Scenario Outline: Starting the wizard by choosing the default answer to the question whether to start it.

src/helpers/configuration.sh
kevgo: I don't see this Y/n added back to the spec.

features/git-town/configuration/configuration_automatic.feature
allewun: [This](https://github.com/Originate/git-town/blob/aw-prompt-default/features/git-town/configuration/configuration_automatic.feature#L64)?

features/git-town/configuration/configuration_automatic.feature
allewun: I'll call it prompt instead, since we haven't reached the wizard yet:

> Git Town hasn't been configured for this repository.
> Please run 'git town config --setup'.
> Would you like to do that now? [Y/n]

src/helpers/configuration.sh
allewun: It's been DRYed up [here](https://github.com/Originate/git-town/blob/aw-prompt-default/features/step_definitions/town_steps.rb#L61-L65).

features/git-town/configuration/configuration_automatic.feature
kevgo: Yep, exactly that!  :)

src/helpers/configuration.sh
kevgo: Cool

kevgo:
:ship:

allewun:
Is there anyway to DRY up the tables? All 4 scenarios here have the same tables.

kevgo:
I'm not aware of a non-hacky way to DRY up the examples in a scenario outline, and I can imagine that the cucumber guys would say that you shouldn't, since it makes the examples a lot less readable.

PR #467, opened 2015-04-01
Show nightly version

allewun:
@kevgo @charlierudolph 

What do you think... useless or not? I frequently wonder exactly what revision I'm using, esp as it's been a while since 0.5.0.. though I tend follow it up with a `git sync` in `git-town/` anyway.

Maybe useful for those that installed via the manual instructions?

### DIFF
--- a/src/git-town
+++ b/src/git-town
@@ -31,7 +31,12 @@ function show_help {
 
 
 function show_version {
-  echo "Git Town 0.5.0$(gittown_dev_version)"
+  local version="0.5.0"
+  if [ "$(is_git_town_installed_manually)" == true ]; then
+    version="$version $(git_town_revision)"
+  fi
+
+  echo "Git Town $version"
 }
 
 
--- a/src/git-town
+++ b/src/git-town
@@ -31,7 +31,7 @@ function show_help {
 
 
 function show_version {
-  echo "Git Town 0.5.0$(gittown_nightly_version)"
+  echo "Git Town 0.5.0$(gittown_dev_version)"
 }
 
 
--- a/src/git-town
+++ b/src/git-town
@@ -31,7 +31,7 @@ function show_help {
 
 
 function show_version {
-  echo "Git Town 0.5.0"
+  echo "Git Town 0.5.0$(gittown_nightly_version)"
 }
 
 
--- a/src/helpers/environment.sh
+++ b/src/helpers/environment.sh
@@ -18,11 +18,12 @@ function ensure_git_repository {
 
 
 function gittown_dev_version {
-  local base_dir=$( dirname "${BASH_SOURCE[0]}" )
-  local git_hash=$(git -C "$base_dir" rev-parse --short HEAD)
-  local git_date=$(git -C "$base_dir" --no-pager show -s --format=%cD HEAD | cut -d ' ' -f2-4)
+  local base_dir=$(dirname "${BASH_SOURCE[0]}")
 
   if ! git -C "$base_dir" remote -v > /dev/null 2>&1 | grep "https://github.com/Homebrew/homebrew.git (fetch)"; then
+    local git_hash=$(git -C "$base_dir" rev-parse --short HEAD)
+    local git_date=$(git -C "$base_dir" --no-pager show -s --format=%cD HEAD | cut -d ' ' -f2-4)
+
     echo " (${git_date}, ${git_hash})"
   fi
 }
--- a/src/helpers/environment.sh
+++ b/src/helpers/environment.sh
@@ -9,17 +9,6 @@ function is_git_repository {
 }
 
 
-function is_gittown_nightly {
-  local base_dir=$( dirname "${BASH_SOURCE[0]}" )
-
-  if git -C "$base_dir" remote -v > /dev/null 2>&1 | grep "https://github.com/Homebrew/homebrew.git (fetch)"; then
-    echo false
-  else
-    echo true
-  fi
-}
-
-
 function ensure_git_repository {
   if [ "$(is_git_repository)" == false ]; then
     echo_inline_error "This is not a git repository."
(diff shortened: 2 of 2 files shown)


kevgo:
We don't have nightly builds. What's the use case for this? 

Also, can you please provide an example output?

allewun:
"Nightly" might be the wrong term. For people who've installed git town manually, the version they use likely has a different feature set than what the release notes indicates, even though `git town version` would print 0.5.0.

allewun:
And just to clarify, if Git Town is installed via homebrew, the output remains unchanged:
```
> git town version
Git Town 0.5.0
```

charlierudolph:
So all you want is the date and commit sha to show in the version?
That seems fine to me. I would prefer a more readable version
```
> git town version
Git Town 0.5.0 (Mar 31, 2015 - b7fe18d) 
```
I think adding the date and sha always would be fine too.

Along that line we need to release more. I'll start preparing a release now actually

allewun:
Updated:

kevgo:
I'm still somehow YAGNI on this. I'd like to caution us to keep Git Town as simple as possible, and only add things for which there is a real user demand, and which solve a real problem. If that's the case for you @allewun then I'm fine merging this. But if not, I'd rather leave it out. Up to you.

allewun:
Completely agree about keeping things simple. This is a fairly low-impact change though. If it causes any issues later on, I have no issue with ripping it out. For the moment, I see it being marginally useful, so let's give it a shot.

src/helpers/environment.sh
```diff
@@ -16,6 +17,17 @@ function ensure_git_repository {
 }
 
 
+function gittown_dev_version {
+  local base_dir=$( dirname "${BASH_SOURCE[0]}" )
```
charlierudolph: remove space padding inside parens

src/helpers/environment.sh
```diff
@@ -16,6 +17,17 @@ function ensure_git_repository {
...
+  local git_hash=$(git -C "$base_dir" rev-parse --short HEAD)
+  local git_date=$(git -C "$base_dir" --no-pager show -s --format=%cD HEAD | cut -d ' ' -f2-4)
+
+  if ! git -C "$base_dir" remote -v > /dev/null 2>&1 | grep "https://github.com/Homebrew/homebrew.git (fetch)"; then
```
charlierudolph: Why the output redirection? And shoudln't this whole thing be in a subshell?

src/helpers/environment.sh
```diff
@@ -16,6 +17,17 @@ function ensure_git_repository {
 }
 
 
+function gittown_dev_version {
+  local base_dir=$( dirname "${BASH_SOURCE[0]}" )
+  local git_hash=$(git -C "$base_dir" rev-parse --short HEAD)
+  local git_date=$(git -C "$base_dir" --no-pager show -s --format=%cD HEAD | cut -d ' ' -f2-4)
```
charlierudolph: Dont calculate these unless you need them (move then to the inside of the if statement.

src/helpers/environment.sh
allewun: Why should it be in a subshell? The output redirection is in case the user isn't using Git 1.8.5 and above and the [`-C`](https://github.com/git/git/blob/5fd09df3937f54c5cfda4f1087f5d99433cce527/Documentation/RelNotes/1.8.5.txt#L115-L116) flag isn't supported.

Related to that, have we been considering the version of Git that users will have?

charlierudolph: All calls where we want to capture the output should be in a subshell.As then the output is returned and we can compare it

allewun: I'm not capturing the output here, I'm only seeing if grep returned successfully.

* success => git-town was installed via homebrew
* failure => git-town was installed some other way

charlierudolph: I suggest against that. Instead you can have grep output the count of matches. This is what we have been doing throughout the repository. 
```sh
if [ "$(git -C "$base_dir" remote -v | grep -c "https://github.com/Homebrew/homebrew.git (fetch)")" == 0 ]; then
  # ...
fi
```

allewun: Alrighty. TBH I don't see any benefit doing that, but for consistency I'll change it

allewun: I guess the negation might be hard to follow

charlierudolph: I disliked the negation and the output redirection.

src/helpers/environment.sh
```diff
@@ -16,6 +17,18 @@ function ensure_git_repository {
 }
 
 
+function gittown_dev_version {
+  local base_dir=$(dirname "${BASH_SOURCE[0]}")
+
+  if ! git -C "$base_dir" remote -v > /dev/null 2>&1 | grep "https://github.com/Homebrew/homebrew.git (fetch)"; then
+    local git_hash=$(git -C "$base_dir" rev-parse --short HEAD)
+    local git_date=$(git -C "$base_dir" --no-pager show -s --format=%cD HEAD | cut -d ' ' -f2-4)
```
charlierudolph: `--no-pager` should be unnecessary as the output is a single line.

charlierudolph: I'd also prefer a space between `-f` and `2-4`

allewun: done

src/helpers/environment.sh
```diff
@@ -16,6 +17,19 @@ function ensure_git_repository {
 }
 
 
+function gittown_dev_version {
+  local base_dir=$(dirname "${BASH_SOURCE[0]}")
+
```
charlierudolph: remove one of these empty lines

src/git-town
```diff
@@ -31,7 +31,7 @@ function show_help {
 
 
 function show_version {
-  echo "Git Town 0.5.0"
+  echo "Git Town 0.5.0$(gittown_dev_version)"
```
charlierudolph: This looks like dev_version is always printed when thats not the case. How about
```sh
local version="0.5.0"
if [ "$(is_manual_install)" == "true" ]; then
  version="$version ($revision_info)"
fi
echo "Git Town $version"
```

src/helpers/environment.sh
allewun: which one? :D

kevgo:
Alright, this looks good from my end. :ship:

charlierudolph:
:ship:

Issue #463, opened 2015-03-31
DRY up verification of uncommitted files


kevgo:
* store the name of the uncommitted file in a variable when it is created
* verify the existence of the uncommitted file through `And I don't have my uncommitted file`.

PR #482, opened 2015-04-06
 DRY up uncommitted file steps

### LINKED ISSUE #455: DRY up uncommitted file steps
There are a million steps with `I have an uncommitted file with name: "uncommitted" and content: "stuff"`. Dry those up into `I have an uncommitted file`.

ricmatsui:
Closes #455 #463

### DIFF
--- a/features/step_definitions/files_steps.rb
+++ b/features/step_definitions/files_steps.rb
@@ -1,5 +1,10 @@
+Given(/^I have an uncommitted file$/) do
+  @uncommitted_file_name = create_uncommitted_file
+end
+
+
 Given(/^I have an uncommitted file with name: "(.+?)" and content: "(.+?)"$/) do |name, content|
-  IO.write name, content
+  create_uncommitted_file file_name: name, file_content: content
 end
 
 
@@ -11,8 +16,18 @@ end
 
 
 
+Then(/^I (?:still|again) have my uncommitted file$/) do
+  verify_uncommitted_file file_name: @uncommitted_file_name
+end
+
+
 Then(/^(?:now I|I still) have the following committed files$/) do |files_data|
   files_data.diff! files_in_branches
+end
+
+
+Then(/^I don't have my uncommitted file$/) do
+  expect(uncommitted_files).to_not include @uncommitted_file
 end
 
 
@@ -26,7 +41,6 @@ Then(/^I don't have any uncommitted files$/) do
 end
 
 
-Then(/^I (?:still|again) have an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |file_name, content|
-  expect(uncommitted_files).to eql [file_name]
-  expect(IO.read file_name).to eql content
+Then(/^I (?:still|again) have an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |name, content|
+  verify_uncommitted_file file_name: name, file_content: content
 end
--- a/features/step_definitions/files_steps.rb
+++ b/features/step_definitions/files_steps.rb
@@ -4,7 +4,7 @@ end
 
 
 Given(/^I have an uncommitted file with name: "(.+?)" and content: "(.+?)"$/) do |name, content|
-  create_uncommitted_file file_name: name, file_content: content
+  create_uncommitted_file name: name, content: content
 end
 
 
@@ -17,7 +17,7 @@ end
 
 
 Then(/^I (?:still|again) have my uncommitted file$/) do
-  verify_uncommitted_file file_name: @uncommitted_file_name
+  verify_uncommitted_file name: @uncommitted_file_name
 end
 
 
@@ -42,5 +42,5 @@ end
 
 
 Then(/^I (?:still|again) have an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |name, content|
-  verify_uncommitted_file file_name: name, file_content: content
+  verify_uncommitted_file name: name, content: content
 end
--- a/features/step_definitions/files_steps.rb
+++ b/features/step_definitions/files_steps.rb
@@ -27,7 +27,7 @@ end
 
 
 Then(/^I don't have my uncommitted file$/) do
-  expect(uncommitted_files).to_not include @uncommitted_file
+  expect(uncommitted_files).to_not include @uncommitted_file_name
 end
 
 
--- a/features/step_definitions/files_steps.rb
+++ b/features/step_definitions/files_steps.rb
@@ -41,6 +41,6 @@ Then(/^I don't have any uncommitted files$/) do
 end
 
 
-Then(/^I (?:still|again) have an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |name, content|
+Then(/^my workspace (?:still|again) has an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |name, content|
   verify_uncommitted_file name: name, content: content
 end
--- a/features/step_definitions/files_steps.rb
+++ b/features/step_definitions/files_steps.rb
@@ -41,6 +41,7 @@ Then(/^I don't have any uncommitted files$/) do
 end
 
 
-Then(/^my workspace (?:still|again) has an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |name, content|
+Then(/^my workspace (?:still|again) has an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) \
+    do |name, content|
   verify_uncommitted_file name: name, content: content
 end
--- a/features/support/files_helpers.rb
+++ b/features/support/files_helpers.rb
@@ -25,27 +25,22 @@ end
 
 def create_uncommitted_file options = {}
   options.reverse_merge! default_uncommitted_file_attributes
-  IO.write options[:file_name], options[:file_content]
-  options[:file_name]
+  IO.write options[:name], options[:content]
+  options[:name]
 end
 
 
-def default_uncommitted_file_attributes
-  {
-    file_name: 'uncommitted_file',
-    file_content: 'default uncommitted content'
-  }
-end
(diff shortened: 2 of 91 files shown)


features/support/files_helpers.rb
```diff
@@ -23,6 +23,32 @@ def files_in_branches
 end
 
 
+def create_uncommitted_file options = {}
+  options.reverse_merge!(default_uncommitted_file_attributes)
```
kevgo: Omit parentheses if not necessary, like here.

features/support/files_helpers.rb
```diff
@@ -23,6 +23,32 @@ def files_in_branches
...
+  IO.write options[:file_name], options[:file_content]
+  options[:file_name]
+end
+
+
+def default_uncommitted_file_attributes
+  {
+    file_name: "default_uncommitted_file_name_#{SecureRandom.urlsafe_base64}",
```
kevgo: What do you think about naming the uncommitted file "uncommitted_file" by default? I don't see a need for the random hash, since we won't create two uncommitted files at the same time. And if the file name ever matters in a test (because we want to test for collisions), the test must specify it explicitly anyways.

features/support/files_helpers.rb
```diff
@@ -23,6 +23,32 @@ def files_in_branches
...
+    file_content: 'default uncommitted content'
+  }
+end
+
+
+def verify_uncommitted_file options = {}
+  if options.key? :file_name
+    options.reverse_merge!(default_uncommitted_file_attributes)
```
kevgo: Remove parentheses

features/support/files_helpers.rb
```diff
@@ -23,6 +23,32 @@ def files_in_branches
...
+    expect(uncommitted_files).to include options[:file_name]
+    expect(IO.read options[:file_name]).to eql options[:file_content]
+  else
+    fail 'Cannot verify uncommitted file without file_name'
```
kevgo: Here you enforce that `options` must have a `file_name` key, but the default value for this parameter is an empty hash. This means that calling this method without an explicit parameter doesn't work, and you should remove the default value for it.

```ruby
def verify_uncommitted_file options
```

kevgo:
Nice work @ricmatsui! Looks very good to me! :+1:

features/support/files_helpers.rb
ricmatsui: Good point.

features/support/files_helpers.rb
ricmatsui: Yeah that works too. I had just included it to be consistent with the default commit attributes, and it also made it clear which tests had been relying on the default file name of "uncommitted_file"

ricmatsui:
Thanks, all comments addressed.

features/step_definitions/files_steps.rb
```diff
@@ -1,5 +1,10 @@
+Given(/^I have an uncommitted file$/) do
+  @uncommitted_file_name = create_uncommitted_file
+end
+
+
 Given(/^I have an uncommitted file with name: "(.+?)" and content: "(.+?)"$/) do |name, content|
-  IO.write name, content
+  create_uncommitted_file file_name: name, file_content: content
```
kevgo: Looking at this again, I think you can shorten the parameter names a bit. The method name makes it pretty clear that this is about files.

```ruby
create_uncommitted_file name: name, content: content
```

What do you think?

features/step_definitions/files_steps.rb
```diff
@@ -11,11 +16,21 @@
 
 
 
+Then(/^I (?:still|again) have my uncommitted file$/) do
+  verify_uncommitted_file file_name: @uncommitted_file_name
```
kevgo: Same here, the parameters could be shortened.

features/step_definitions/files_steps.rb
```diff
@@ -11,11 +16,21 @@
...
+end
+
+
 Then(/^(?:now I|I still) have the following committed files$/) do |files_data|
   files_data.diff! files_in_branches
 end
 
 
+Then(/^I don't have my uncommitted file$/) do
```
kevgo: Looking at this again, this is a chance to improve the name of this step definition as well. It was never that great to begin with. "I don't have my uncommitted file" is technically not true, its still there, just stashed away. What do you think about `Then my uncommitted file is( still)? stashed away`?

features/step_definitions/files_steps.rb
```diff
@@ -26,7 +41,6 @@
 end
 
 
-Then(/^I (?:still|again) have an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |file_name, content|
-  expect(uncommitted_files).to eql [file_name]
-  expect(IO.read file_name).to eql content
+Then(/^I (?:still|again) have an uncommitted file with name: "([^"]+)" and content: "([^"]+)"$/) do |name, content|
```
kevgo: What about `Then my workspace (?:still|again) has an uncommitted file with name: "([^"]+)" and content: "([^"]+)"`? Sorry about the shotgun refactoring this requires. If its too much work, we can also file a separate ticket for this, and I'll do it very efficiently using Vim.

features/support/files_helpers.rb
```diff
@@ -23,6 +23,32 @@ def files_in_branches
...
+def default_uncommitted_file_attributes
+  {
+    file_name: 'uncommitted_file',
+    file_content: 'default uncommitted content'
+  }
+end
+
+
+def verify_uncommitted_file options
+  if options.key? :file_name
```
kevgo: Looking at this again, I realize that this `if` clause is unnecessary and can be removed.

* you reverse-merge a file_name into the options hash on the next line anyways if one is missing
* manually checking that certain parameters have certain keys is manually implementing a static type system. Ruby as a dynamically typed language shouldn't need to do this. If we wanted type checking, we would use a statically typed language. 
* if you really wanted to keep this check in, I suggest a guard clause:

  ```ruby
def verify_uncommitted_file options
  return fail 'Cannot verify uncommitted file without file_name'  unless options.key? :file_name
  options.reverse_merge! default_uncommitted_file_attributes
  ...
```

So I would try to take this if-clause out and see if it still works. I think it will. 

```ruby
def verify_uncommitted_file options
  options.reverse_merge! default_uncommitted_file_attributes
  expect(uncommitted_files).to include options[:file_name]
  expect(IO.read options[:file_name]).to eql options[:file_content]
end
```

features/support/files_helpers.rb
```diff
@@ -23,6 +23,32 @@ def files_in_branches
...
 
 
+def create_uncommitted_file options = {}
+  options.reverse_merge! default_uncommitted_file_attributes
+  IO.write options[:file_name], options[:file_content]
+  options[:file_name]
+end
+
+
+def default_uncommitted_file_attributes
```
kevgo: Good idea to dry this up by separating these atttributes into their own place! You could also make this a constant.

```ruby
DEFAULT_UNCOMMITTED_FILE_ATTRIBUTES = {
  file_name: 'uncommitted_file',
  file_content: 'default uncommitted content'
}.freeze
```

features/step_definitions/files_steps.rb
ricmatsui: Sure, works for me too. I'll change it in the default attributes to match.

features/step_definitions/files_steps.rb
ricmatsui: Well it doesn't check that it's stashed away, so the definition would need to change to match. This is related to #480 's step, so what do you think of switching to `still stashed away` after that PR?

features/step_definitions/files_steps.rb
ricmatsui: Sure, no problem. Don't worry, I'm using find+sed which is maybe a few keystrokes slower :) . Are you using a plugin to do something like this in Vim?

features/support/files_helpers.rb
ricmatsui: Yeah you're right. What do you think of something like this?
```
def verify_uncommitted_file name, options = {}
  options[:name] = name
  ....
end

File: features/support/files_helpers.rb (hunk 10 above)
Comment (ricmatsui) [inline features/support/files_helpers.rb]: Cool.

File: features/support/files_helpers.rb (hunk 9 above)
Comment (ricmatsui) [inline features/support/files_helpers.rb]: Never mind, just take out the if-clause.

File: features/step_definitions/files_steps.rb (hunk 8 above)
Comment (kevgo) [inline features/step_definitions/files_steps.rb]: Wow, that's pretty hardcore! I'm using [ag.vim](https://github.com/rking/ag.vim) to find all matching locations. After changing the first result, and going to the next result, I use a custom key combination to re-apply the last change and go to the next search result.

```viml
function ReplaceAndGoToNext()
  :normal! .
  :cnext
endfunction
nnoremap <silent> <C-]> :call ReplaceAndGoToNext()<CR>
```

features/step_definitions/files_steps.rb
kevgo: Yeah, that's a good point. Let's do this later!

kevgo:
Nice work! :ship: it!

features/step_definitions/files_steps.rb
ricmatsui: Haha well StackOverflow is hardcore then! I think it's a nice approach to remember because it works anywhere. Like:
```bash
find features/ -type f -exec sed -i 's/find/replace/g' '{}' \;
```

I see, that's a good way of doing it and a nice plugin, thanks! And it's definitely fast.

kevgo: Thanks for sharing, I wanted to ask you for your solution, but then got stuck in an emergency situation at work for the last 2 days. Having such a global command-line search and replace solution sounds very useful to have! The boilerplate for this has always made me not use such a solution regularly. But it should be super easy to write an easier to use bash wrapper, like `gr foo bar`, which globally replaces all occurrences of "foo" with "bar" in the current directory or its children. I'll add that to my tool belt!

PR #491, opened 2015-04-15
Debugging through tags

kevgo:

Debugging through an environment variable is often cumbersome, not only because I use Fish shell, but also because I have to change the command called when debugging.

Adding a tag to the Cucumber specs is often easier, so this PR adds support for that, and also cleans up some no longer needed infrastructure for handling the debug flags.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,14 +63,15 @@ about problems.
 
 ## Debugging
 
-To see the output of the Git commands run in tests, you can either
-* set the `DEBUG_COMMANDS` environment variable while running your specs:
+To see the output of the Git commands run in tests, you can either set the
+`DEBUG_COMMANDS` environment variable while running your specs:
 
-  ```bash
-  $ DEBUG_COMMANDS=true cucumber <filename>[:<lineno>]
-  ```
+```bash
+$ DEBUG_COMMANDS=true cucumber <filename>[:<lineno>]
+```
 
-* add a `@debug-commands` flag to the respective Cucumber spec
+Alternatively, you can also add a `@debug-commands` flag to the respective
+Cucumber spec:
 
   ```cucumber
   @debug-commands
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -64,6 +61,28 @@ and notifies you via email and through status badges in pull requests
 about problems.
 
 
+## Debugging
+
+To see the output of the Git commands run in tests, you can either set the
```
charlierudolph: Remove `either`. You aren't listing the two options in a single sentence.

CONTRIBUTING.md
```diff
@@ -64,6 +61,28 @@ and notifies you via email and through status badges in pull requests
...
+Cucumber spec:
+
+  ```cucumber
+  @debug-commands
+  Scenario: foo bar baz
+    Given ...
+  ```
+
+For even more detailed output, you can use the `DEBUG` variable or tag
+in a similar fashion.
```
charlierudolph: Let explain what DEBUG actually does - prints every shell command executed and its output.

charlierudolph:
Very cool

allewun:
I like this. Can we make the debugging output stand out a bit from the other cucumber output? How about this (I prepended each line with an `@` symbol)?

screen shot 2015-04-15 at 6 30 29 pm

```ruby
def print_result result
  puts ''
  puts "#{result.location}$ #{result.command}".gsub(/^/, '@ ')
  puts "#{result.out}".gsub(/^/, '@ ')
  puts ''
end
```

charlierudolph:
Nice idea. I think we can dry up the use of gsub
```ruby
def print_result result
  puts ''
  puts %(
    #{result.location}$ #{result.command}
    #{result.out}"
  ).gsub(/^/, '@ ')
  puts ''
end

CONTRIBUTING.md
kevgo: Removed.

CONTRIBUTING.md
kevgo: Good idea. Added.

kevgo:
I have extracted @allewun's idea into #494

kevgo:
All comments addressed

charlierudolph:
:ship:

PR #498, opened 2015-04-20
rename debugging environment variables

allewun:
I think these variables should be namespaced. I just went ahead and made the PR since it's so small rather than create an issue first.

* `DEBUG_COMMANDS` → `GIT_TOWN_DEBUG_COMMANDS`
* `DEBUG` → `GIT_TOWN_DEBUG`

@kevgo @charlierudolph @ricmatsui

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,13 +64,13 @@ about problems.
 ## Debugging
 
 To see the output of the Git commands run in tests, you can set the
-`DEBUG_COMMANDS` environment variable while running your specs:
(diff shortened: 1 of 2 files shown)


kevgo:
Why do you think these variables should be namespaced? That makes them much harder to use in my opinion, and I don't see a scenario where there would be unintended conflicts.

allewun:
True. The issue was if the environment variables were already defined in the shell by some other process or set in a *.rc file for use by another process, they would be consumed unintentionally by git town.

On second thought, this is probably more trouble than it's worth.

PR #501, opened 2015-04-22
Save previous branch

allewun:
This implements #65 

This table summarizes the different cases handled:

| Previous Branch (A) | Current Branch (B) | Result        |
|:--------------:|:-------------:|---------------|
| keep         | keep        | A --> B (A is now the previous branch, B is the current)      |
| keep         | delete      | A --> main    |
| keep         | new (C)     | B --> C       |
| delete       | keep        | main --> B    |
| delete       | delete      | main --> main |
| delete       | new         | N/A           |

For any given git town command, the previous branch may have either been 1) deleted or 2) untouched, and the current branch may have been 1) unchanged, 2) deleted, 3) changed to a new branch.

After `run_steps` finishes, a new function `set_previous_branch`, which is overridden by each command, runs. It will stash any open changes, do 2 checkouts, then pop the stash if necessary. The 2 checkouts will insert the desired previous branch into the checkout history so that `git checkout -` works as expected.

@kevgo @charlierudolph @ricmatsui


features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Set the main branch as the previous branch after running a Git Town commmand that deletes the initial previous branch
+
+  (see ../same_current_branch/previous_branch_same.feature)
+
+
+  Scenario: checkout previous branch after a git-prune-branches that deletes previous and current branches
+    Given I have feature branches named "previous" and "current"
+    And I am on the "previous" branch
+    And I switch to the "current" branch
```
kevgo: And I checkout the "current" branch

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,12 @@
...
+    And I run `git prune-branches`
+    When I checkout my previous git branch
```
kevgo: What about "When I run `git checkout -`" to make it more clear what you do here? Alternatively you could do both: "When I run `git checkout -` to checkout my previous Git branch"

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,36 @@
+Feature: Allow checking out previous git branch to work correctly after running a Git Town commmand that deletes the current branch
```
kevgo: Feature: Allow checking out **the correct** previous **G**it branch ~~to work correctly~~ after running a Git Town commmand that deletes the current branch

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,36 @@
...
+  Scenario: checkout previous branch after git-ship deletes current branch
+    Given I have feature branches named "previous", "current"
+    And the following commit exists in my repository
+      | BRANCH  | LOCATION | FILE NAME    | FILE CONTENT    |
+      | current | remote   | feature_file | feature content |
+    And I am on the "previous" branch
+    And I switch to the "current" branch
+    And I run `git ship -m "feature done"`
+    When I checkout my previous git branch
+    Then I end up on the "previous" branch
```
kevgo: These scenarios seem to all to the same thing. You could either do a Scenario Outline, or - since you need to make sure that each command works - integrate these specs into the most appropriate normal Git Town command spec.

features/shared/checkout_previous_branch_after_success/new_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,26 @@
...
+    Then I end up on the "current" branch
+
+
+  Scenario: checkout previous branch after git-hack
+    Given I have feature branches named "previous" and "current"
+    And I am on the "previous" branch
+    And I switch to the "current" branch
+    And I run `git hack new`
+    When I checkout my previous git branch
+    Then I end up on the "current" branch
```
kevgo: What do you think about unifying these two steps into 

> Then running \`git checkout -\` correctly takes me back to the "current" branch

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,58 @@
+ Feature: Allow checking out previous git branch to work correctly after running a Git Town commmand that leaves the user on the same branch
+
+  As a developer running `git checkout -` after running a Git Town command
+  I want to end up on the expected previous branch
+  So that I can consistently and effectively use git's commands
```
kevgo: Uppercase "Git"

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
allewun: They do the same thing, but the scenarios don't necessarily have the same setup.

Off-topic: I think I'm seeing a bug in github's rendering here:
screen shot 2015-04-23 at 10 04 48 am

features/step_definitions/branch_steps.rb
```diff
@@ -6,6 +6,11 @@
 end
 
 
+Given(/^I switch to the "(.+?)" branch$/) do |branch_name|
+  step %(I am on the "#{branch_name}" branch)
```
kevgo: Nice and clean!

features/step_definitions/branch_steps.rb
```diff
@@ -54,6 +59,13 @@
 
 
 
+When(/^I checkout my previous git branch$/) do
+  step 'I run `git checkout -`'
```
kevgo: This is also pretty tight!

src/helpers/helpers.sh
```diff
@@ -43,6 +43,7 @@ if [ "$(is_git_repository)" == true ]; then
   export INITIAL_BRANCH_NAME=$(get_current_branch_name)
   export INITIAL_DIRECTORY=$(pwd)
   export INITIAL_OPEN_CHANGES=$(has_open_changes)
+  export INITIAL_PREVIOUS_BRANCH_NAME=$(get_previous_branch_name)
```
charlierudolph: I think `PREVIOUS_BRANCH_NAME` is sufficient. Initial doesn't add anything to me

src/git-ship
```diff
@@ -26,6 +26,20 @@ function preconditions {
...
+  # deleting previous branch
+  if [ "$target_branch_name" == "$INITIAL_PREVIOUS_BRANCH_NAME" ]; then
+    set_branch_as_previous_branch "$MAIN_BRANCH_NAME"
+  # deleting current branch
+  elif [ "$target_branch_name" == "$INITIAL_BRANCH_NAME" ]; then
+    set_branch_as_previous_branch "$INITIAL_PREVIOUS_BRANCH_NAME"
+  # deleting some other branch
+  else
+    set_branch_as_previous_branch "$INITIAL_PREVIOUS_BRANCH_NAME"
+  fi
```
charlierudolph: These two last branches result in the same functionality, can't they be combined?

src/git-prune-branches
```diff
@@ -22,8 +22,23 @@ function should_delete_branch {
...
+  else
+    set_branch_as_previous_branch "$INITIAL_PREVIOUS_BRANCH_NAME"
+  fi
+}
+
+
 function steps {
-  local initial_branch_deleted=false
+  prune_branches_initial_branch_deleted=false
+  prune_branches_initial_previous_branch_deleted=false
```
charlierudolph: why are these namespace? initial_branch_deleted and initial_previous_branch_deleted (i think it should just be previous_branch_deleted) is sufficient.

src/git-prune-branches
```diff
@@ -22,8 +22,23 @@ function should_delete_branch {
...
+  # previous branch was deleted
+  if [ "$prune_branches_initial_previous_branch_deleted" = true ]; then
+    set_branch_as_previous_branch "$MAIN_BRANCH_NAME"
+  # current branch was deleted
+  elif [ "$prune_branches_initial_branch_deleted" = true ]; then
+    set_branch_as_previous_branch "$INITIAL_PREVIOUS_BRANCH_NAME"
+  # some other branch was deleted
+  else
+    set_branch_as_previous_branch "$INITIAL_PREVIOUS_BRANCH_NAME"
+  fi
```
charlierudolph: can these last two branches be combined?

src/git-kill
```diff
@@ -12,31 +12,47 @@ function branch_to_kill {
...
+function set_previous_branch {
+  # previous branch was deleted
+  if [ "$target_branch_name" == "$INITIAL_PREVIOUS_BRANCH_NAME" ]; then
+    set_branch_as_previous_branch "$MAIN_BRANCH_NAME"
+  # current branch was deleted
+  elif [ "$target_branch_name" == "$INITIAL_BRANCH_NAME" ]; then
+    set_branch_as_previous_branch "$INITIAL_PREVIOUS_BRANCH_NAME"
+  # some other branch was deleted
+  else
+    set_branch_as_previous_branch "$INITIAL_PREVIOUS_BRANCH_NAME"
```
charlierudolph: Can these last two branches be combined?

src/git-kill
```diff
@@ -12,31 +12,47 @@ function branch_to_kill {
...
-    ensure_has_branch "$target_branch"
+  if [ "$target_branch_name" != "$INITIAL_BRANCH_NAME" ]; then
+    ensure_has_branch "$target_branch_name"
+  fi
+}
+
+
+function set_previous_branch {
+  # previous branch was deleted
+  if [ "$target_branch_name" == "$INITIAL_PREVIOUS_BRANCH_NAME" ]; then
```
kevgo: Based on your description, I think the new "previous branch" is always the existing previous branch, except if that branch gets deleted by the current Git Town command. In this case its always main. 

If I understand this correctly, you could automate this fully, without customizations in each Git Town command.

src/helpers/script_helpers.sh
```diff
@@ -153,6 +154,13 @@ function run_steps {
 }
 
 
+# Sets the previous branch after the git town command completes
+# This should be overridden when necessary
+function set_previous_branch {
+  true
+}
```
charlierudolph: How about about this just returns the branch name that needs to be set as the previous branch? Should help remove some duplication. (may mean renaming the function)

src/git-kill
allewun: Hm I think you're right. Simply check for the existence of the previous branch... if it exists use that, otherwise use $main.

src/helpers/git_helpers/branch_helpers.sh
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/env bash
 
 
+# Checkout a branch
+function checkout_branch_silently {
```
kevgo: I'd implement the "silent" aspect as a parameter to a `checkout_branch` function.

```bash
function checkout_branch {
  $silent=$1
```

Then `run_command` also gets such a parameter, and we are good to go.

src/helpers/git_helpers/open_changes_helpers.sh
```diff
@@ -40,12 +40,24 @@ function restore_open_changes {
 }
 
 
+# Unstashes changes silently (no console output)
+function restore_open_changes_silently {
```
kevgo: Same here, make "silently" a parameter to the normal `restore_open_changes` method.

src/git-kill
charlierudolph: Except for the cases where the branch changes though right? git hack and git extract should have the initial branch as the previous branch.

src/helpers/terminal_helpers.sh
```diff
@@ -181,6 +181,13 @@ function run_command {
 }
 
 
+# Run a command silently (won't print to console)
+function run_command_silently {
```
kevgo:
```bash
function run_command {
  local silent=$1
  local cmd="$*"
  if [ "$silent" == "silent" ]; then
    eval "$cmd" > /dev/null 2>&1
  else
    echo_header "$cmd"
    eval "$cmd" 2>&1
  fi
}
```

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
kevgo: Wow, that's scary!

src/git-kill
kevgo: Yes, if the Git Town command ends up on a different branch than $INITIAL_BRANCH_NAME, then use $INITIAL_BRANCH_NAME as the previous branch. This behavior is consistent across commands as well.

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
allewun: Bug reported

kevgo:
No offense to your fine feature here, gunslinger, but I'd like to vote for a config flag to disable it if wanted. This feature slows everything down with up to 10 extra Git commands (getting the previous branches, getting the current branch, stashing, checking out branches etc). I don't need this functionality, so I don't want to pay the speed penalty for it on every command.

This should be done in a separate PR, and I'm happy to have the feature active by default, with me having to go the extra mile to disable it.

allewun:
Where did you get 10 from? I only count 7 in the worst case.

```bash
 git status | grep -c "rebase in progress"
 git rev-parse --abbrev-ref HEAD
 git status --porcelain | wc -l | tr -d ' '
 git stash -u
 git checkout one
 git checkout two
 git stash pop
```

No offense taken, it's a valid point that I didn't consider. However doing a quick test, these 7 commands only add ~0.11 sec to the execution time (maybe add a tiny bit more for the bash function overhead). I suppose that's noticeable, but are you really issuing so many git town commands that it's a significant penalty?

kevgo:
I was just guessing, and under-estimated your nifty implementation! I actually guessed 7 first, but then got carried away... Anyways, let's merge this in first without flag, then see how things feel, then decide about the config setting.

src/git-prune-branches
allewun: These are global variables now and `initial_branch_deleted` seems generic enough that it might be defined elsewhere.

charlierudolph: It may be used in another git town command but shouldn't be used in any helper. Also what about the idea of instead having the global PREVIOUS_BRANCH_NAME simply be modified anytime we need to set the pervious branch to something other than its initial value.

src/helpers/terminal_helpers.sh
charlierudolph: That doesn't work. We are expecting the command to be all the arguments (though I'm uncertain if thats still needed). Maybe we can have the command be the first argument and silent be the second (optional)?

src/helpers/helpers.sh
allewun: I think it's much more clear knowing that it's the initial state, and there's parity with `INITIAL_BRANCH_NAME`

src/helpers/terminal_helpers.sh
allewun: I don't think that works either, since `$*` grabs everything else

allewun:
all comments have been addressed

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,36 @@
...
+  (see ../same_current_branch/previous_branch_same.feature)
+
+
+  Scenario: checkout previous branch after git-kill keeps the previous branch intact and deletes the current branch
+    Given I have feature branches named "previous" and "current"
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    And I run `git kill`
+    When I run `git checkout -` to checkout my previous Git branch
+    Then I end up on the "previous" branch
```
charlierudolph: I think this is more straightfoward with the additional step so we know where we are after the command ends
```cucumber
When I run `git kill`
Then I end up on the "main" branch
When I run `git checkout -` to checkout my previous Git branch
Then I end up on the "previous" branch
```

charlierudolph: I also think `to checkout my previous Git branch` is unnecessary.

src/helpers/script_helpers.sh
```diff
@@ -98,6 +98,30 @@ function remove_step_files {
...
+    # current branch is new
+    else
+      set_branch_as_previous_branch "$INITIAL_BRANCH_NAME"
+    fi
+
+  # previous branch is deleted
+  else
+    set_branch_as_previous_branch "$MAIN_BRANCH_NAME"
+  fi
+}
```
charlierudolph: I think this function should be split in 2
```sh
function restore_proper_previous_branch {
  set_branch_as_previous_branch "$(get_expected_previous_branch)"
}


function get_expected_previous_branch {
  if [ "$(has_branch "$INITIAL_PREVIOUS_BRANCH_NAME")" = true ]; then

    # current branch is unchanged
    if [ "$(get_current_branch_name)" = "$INITIAL_BRANCH_NAME" ]; then
      echo "$INITIAL_PREVIOUS_BRANCH_NAME"

    # current branch is deleted
    elif [ "$(has_branch "$INITIAL_BRANCH_NAME")" = false ]; then
      echo "$INITIAL_PREVIOUS_BRANCH_NAME"

    # current branch is new
    else
      echo "$INITIAL_BRANCH_NAME"
    fi

  # previous branch is deleted
  else
    echo "$MAIN_BRANCH_NAME"
  fi
}
```

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
allewun: good idea

src/helpers/git_helpers/branch_helpers.sh
```diff
@@ -148,6 +164,24 @@ function remote_only_merged_branches {
 }
 
 
+function set_branch_as_previous_branch {
+  local desired_previous_branch=$1
+  local current_branch="$(get_current_branch_name)"
```
charlierudolph: How about an early exit here
```sh
if [ "$desired_previous_branch" = "$(get_previous_branch_name)" ]; then
  return
fi
```

allewun:
please have another look

src/helpers/helpers.sh
ricmatsui: Yeah, I also think the parity with `INITIAL_` is important if only to keep the naming consistent with the purpose of recording what the inital state was.

ricmatsui:
I think this looks good. The performance should not be a problem unless your branches are significantly different because only then, I think, would it require a significant number of writes to disk to checkout.

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Allow checking out the correct previous Git branch after running a Git Town commmand that deletes the previous and current branches
+
+  (see ../same_current_branch/previous_branch_same.feature)
+
+
+  Scenario: checkout previous branch after a git-prune-branches that deletes previous and current branches
+    Given I have feature branches named "previous" and "current"
+    And I am on the "previous" branch
+    And I checkout the "current" branch
```
kevgo: How do you feel with expressing the fact that "previous" is the previous Git branch directly here? That's what we try to do in Cucumber.

> Given I have feature branches named "previous" and "current"
> And I am on the "current" branch with "previous" as the previous Git branch
> When I run `git prune-branches`

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,13 @@
...
+    When I run `git prune-branches`
+    Then I end up on the "main" branch
+    When I run `git checkout -`
+    Then I end up on the "main" branch
```
kevgo: In a similar fashion, this would convert to:

> Then I end up on the "main" branch
> And my new previous Git branch is now "main"

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Allow checking out the correct previous Git branch after running a Git Town commmand that deletes the previous and current branches
```
kevgo: Feature: setting "main" as the previous Git branch after running a Git Town command that deletes the previous and current branch

?

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Allow checking out the correct previous Git branch after running a Git Town commmand that keeps the previous branch intact and deletes the current branch
```
kevgo: Feature: Git Town commands that delete the current branch keep the previous Git branch as-is

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,39 @@
...
+
+  (see ../same_current_branch/previous_branch_same.feature)
+
+
+  Scenario: checkout previous branch after git-kill keeps the previous branch intact and deletes the current branch
```
kevgo: Scenario: git-kill

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,39 @@
...
+    Given I have feature branches named "previous" and "current"
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git kill`
+    Then I end up on the "main" branch
+    When I run `git checkout -`
+    Then I end up on the "previous" branch
+
+
+  Scenario: checkout previous branch after a git-prune-branches keeps the previous branch intact and deletes the current branch
```
kevgo: Scenario: git prune-branches

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,39 @@
...
+      | previous | local    | previous_file | previous content |
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git prune-branches`
+    Then I end up on the "main" branch
+    When I run `git checkout -`
+    Then I end up on the "previous" branch
+
+
+  Scenario: checkout previous branch after git-ship keeps the previous branch intact and deletes the current branch
```
kevgo: Scenario: git ship

features/shared/checkout_previous_branch_after_success/new_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Allow checking out the correct previous Git branch after running a Git Town commmand that creates a new current branch intact and leaves the previous branch intact
```
kevgo: Feature: Git Town commands that create a new branch set the current branch as the previous Git branch

features/shared/checkout_previous_branch_after_success/new_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  (see ../same_current_branch/previous_branch_same.feature)
+
+
+  Scenario: checkout previous branch after git-extract
```
kevgo: Scenario: git extract

features/shared/checkout_previous_branch_after_success/new_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,28 @@
...
+      |         |          | refactor commit | refactor_file |
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git extract refactor` with the last commit sha
+    Then I end up on the "refactor" branch
+    When I run `git checkout -`
+    Then I end up on the "current" branch
+
+
+  Scenario: checkout previous branch after git-hack
```
kevgo: Scenario: git hack

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Allow checking out the correct previous Git branch after running a Git Town commmand that leaves the current branch intact and deletes the previous branch
```
kevgo: Feature: deleting the previous Git branch sets main as the new previous Git branch

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,39 @@
...
+
+  (see ./previous_branch_same.feature)
+
+
+  Scenario: checkout previous branch after git-kill leaves current branch intact and deletes the previous branch
```
kevgo: Scenario: git kill

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,39 @@
...
+    Given I have feature branches named "previous" and "current"
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git kill previous`
+    Then I end up on the "current" branch
+    When I run `git checkout -`
+    Then I end up on the "main" branch
+
+
+  Scenario: checkout previous branch after git-prune-branches leaves current branch intact and deletes the previous branch
```
kevgo: Scenario: git prune-branches deletes the previous branch

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_deleted.feature
```diff
@@ -0,0 +1,39 @@
...
+      | current | local    | current_file | current content |
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git prune-branches`
+    Then I end up on the "current" branch
+    When I run `git checkout -`
+    Then I end up on the "main" branch
+
+
+  Scenario: checkout previous branch after git-ship leaves current branch intact and deletes the previous branch
```
kevgo: Scenario: shipping the previous branch

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,63 @@
+Feature: Allow checking out the correct previous Git branch after running a Git Town commmand that leaves the current and previous branches intact
```
kevgo: Feature: Git Town commands that don't change the current and previous branch preserve Git history

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,63 @@
...
+
+  As a developer running `git checkout -` after running a Git Town command
+  I want to end up on the expected previous branch
+  So that I can consistently and effectively use Git's commands
```
kevgo: So that Git Town does not interfere with my productive use of the Git command history

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,63 @@
...
+
+
+  Scenario: checkout previous branch after git-kill leaves current and previous branches intact
```
kevgo: Scenario: killing a branch not in the immediate Git history

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,63 @@
...
+    Given I have feature branches named "previous", "current", and "victim"
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git kill victim`
+    Then I end up on the "current" branch
+    When I run `git checkout -`
+    Then I end up on the "previous" branch
+
+
+  Scenario: checkout previous branch after git-prune-branches leaves current and previous branches intact
```
kevgo: Scenario: pruning a branch not in the immediate Git history

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,63 @@
...
+      | current  | local    | current_file  | current content  |
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git prune-branches`
+    Then I end up on the "current" branch
+    When I run `git checkout -`
+    Then I end up on the "previous" branch
+
+
+  Scenario: checkout previous branch after git-ship leaves current and previous branches intact
```
kevgo: Scenario: shipping a branch not in the immediate Git history

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,63 @@
...
+      | feature | remote   | feature_file | feature content |
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git ship feature -m "feature done"`
+    Then I end up on the "current" branch
+    When I run `git checkout -`
+    Then I end up on the "previous" branch
+
+
+  Scenario: checkout previous branch after git-sync leaves current and previous branches intact
```
kevgo: Scenario: git sync

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_same.feature
```diff
@@ -0,0 +1,63 @@
...
+    Given I have feature branches named "previous" and "current"
+    And I am on the "previous" branch
+    And I checkout the "current" branch
+    When I run `git sync`
+    Then I end up on the "current" branch
+    When I run `git checkout -`
+    Then I end up on the "previous" branch
+
+
+  Scenario: checkout previous branch after git-sync-fork leaves current and previous branches intact
```
kevgo: Scenario: git sync-fork

src/helpers/git_helpers/branch_helpers.sh
```diff
@@ -148,6 +166,30 @@ function remote_only_merged_branches {
 }
 
 
+function set_branch_as_previous_branch {
```
kevgo: How about naming this method `set_previous_branch`?

src/helpers/helpers.sh
kevgo: +1 on keeping the naming consistent, and distinguishing the initial previous branch from the current previous branch later.

src/helpers/script_helpers.sh
```diff
@@ -117,6 +145,15 @@ function run {
     preconditions "$@"
     steps > "$STEPS_FILE"
     run_steps "$STEPS_FILE" undoable
+
+    # Apply global previous-branch restoration
+    # only if we're at the git root.
+    # Any branch restoration logic for the other
+    # case needs to be handled in a custom way by
+    # the command itself
```
kevgo: Since https://github.com/Originate/git-town/commit/16c074ff7fd47de7040b694b0ba2a3227d52839b we are in the Git root during execution. Can't you run this stuff while being in the git root?

src/git-sync
```diff
@@ -31,6 +31,13 @@ function preconditions {
 }
 
 
+function set_previous_branch {
+  if [ "$IN_SUB_FOLDER" = false ]; then
+    set_initial_previous_branch_as_previous_branch
```
kevgo: Could this crazy long method (which isn't defined in this PR) be replaced with `set_previous_branch $INITIAL_PREVIOUS_BRANCH_NAME` or something comparable?

allewun: Yes you're right (see line 88). This was left behind from a previous commit. Will remove the function.

src/helpers/git_helpers/branch_helpers.sh
allewun: works for me

src/helpers/script_helpers.sh
charlierudolph: That was only for `git sync`. And those got added to the steps definitions so by this time we will be back to the initial folder. Looking at this now I believe this actually needs to be added to the steps for each command instead of out here as otherwise we may not get unexpected behavior when there are conflicts.

features/shared/checkout_previous_branch_after_success/deleted_current_branch/previous_branch_deleted.feature
allewun: How does this sound to you:
> Feature: when the previous and current branches are deleted during a Git Town command, the main branch becomes the new previous branch"

Your suggestion suggests that deleting the branches has a direct consequence of updating the branch history, but I see it as more of a side-effect.

features/shared/checkout_previous_branch_after_success/new_current_branch/previous_branch_same.feature
allewun: Same as https://github.com/Originate/git-town/pull/501#discussion_r29312041

So in this case:
> Feature: when a new branch is created during a Git Town command, the current branch becomes the new previous branch

features/shared/checkout_previous_branch_after_success/same_current_branch/previous_branch_deleted.feature
allewun: Same as https://github.com/Originate/git-town/pull/501#discussion_r29312041

So in this case:
> Feature: when the previous branch is deleted during a Git Town command, the main branch becomes the new previous branch

src/helpers/script_helpers.sh
allewun: What do you propose then? The current solution is robust enough that all tests still pass.

And if we have conflicts, I think the `restore_proper_previous_branch` won't be reached as `run_steps` will exit before that.

charlierudolph: I'm addressing something that has not been tested, which is that restoring the previous branch works as expected after aborting / continuing when they are conflicts. Unsure if we can actually handle aborting correctly with our current architecture, but continuing would be possible if the steps for resetting the previous branch were added to the steps file.

I see two ways of accomplishing this.

1) We go back to your original approach of updating each command with custom logic. This allows nice optimization for specific commands but does muddy each commands logic.

2) We add steps to the steps file after the command has written its basic steps. To avoid the extra stash push and pop, it might make sense to also pull these out of each command's logic and into the architecture. I'm thinking of something like
```sh
# src/helpers/script_helpers.sh
function run {
  if [ "$1" = "--abort" ]; then
    ensure_abortable
    abort_command
  elif [ "$1" = "--continue" ]; then
    ensure_continuable
    ensure_no_conflicts
    continue_command
  elif [ "$1" = "--skip" ]; then
    ensure_skippable
    skip_command
  elif [ "$1" = "--undo" ]; then
    ensure_undoable
    undo_command
  else
    remove_step_files
    preconditions "$@"
    pre_steps > "$STEPS_FILE" # This adds the logic for stashing 
    steps >> "$STEPS_FILE"
    post_steps >> "$STEPS_FILE" # This adds the logic for restoring the previous branch and then popping the stash
    run_steps "$STEPS_FILE" undoable
  fi
}
```
Each command may have some configuration like should stash and what are the possible previous branches to set to (if they exist)

kevgo: I like the approach of (2), and think we should go even further with this idea. Almost every command has `echo_if_true "stash_open_changes" "$INITIAL_OPEN_CHANGES"`. And a lot of them need to preserve the Git history. These commands just provide a clean, standardized workspace to run the actual command logic safely. We should make these things dedicated, configurable services that are automatically run by the GT runtime.

```bash
# src/helpers/script_helpers.sh
function run {
  if [ "$1" = "--abort" ]; then
    ensure_abortable
    abort_command
  elif [ "$1" = "--continue" ]; then
    ensure_continuable
    ensure_no_conflicts
    continue_command
  elif [ "$1" = "--skip" ]; then
    ensure_skippable
    skip_command
  elif [ "$1" = "--undo" ]; then
    ensure_undoable
    undo_command
  else
    
    # Cleanup
    remove_step_files

    # Prepare
    preconditions "$@"
    cd_git_root                 
    save_git_history
    stash_open_changes

    # Run core command logic
    steps >> "$STEPS_FILE"
    run_steps "$STEPS_FILE" undoable

    # Restore environment
    restore_git_history
    unstash_open_changes
    cd_original_folder
  fi
}
```

Commands can configure whether they want any of the GT environment services:

```bash
# src/git-sync-fork
export stash_open_changes=false   # thanks, but I don't need changes stashed away, thank you
export preserve_git_history=true  # yes, I change Git history, please preserve it
```

allewun: Sorry it's been a while. At a quick glance, I really like this idea. I'll chew on it when I get back to finishing this feature.

charlierudolph:
@allewun this has been stagnant for a while. Based on the conversation this looks like it will require a bigger overall change to the architecture. How do you feel about the following: 
* close this PR
* open an issue about moving stashing and change to git root as part of the infrastructure
* mark #65 as blocked on by the issue above

charlierudolph:
#501 is already an issue actually

allewun:
Agreed, will probably be best to tackle this fresh. The architecture seems to have changed a bit since I've last looked.

Issue #529, opened 2015-05-28
Support for nested feature branches


kevgo:
> As a user working on a feature that depends on another feature currently under review
I want to be able to cut feature branches off other feature branches and sync them properly
So that my development pace is not slowed down by having to wait until the features that my new features depend on ship.

* [x] `git hack my-feature dependent-feature` cuts the `my-feature` branch off the `dependent-feature` branch, not the main branch
* [x] `git sync dependent-feature` first merges the main branch into the `dependent-feature` branch, then the `dependent-feature` branch into the `my-feature` branch.
* [x] `git ship dependent-feature` ships the given branch, then adjusts all dependent branches like `my-feature` to be cut directly off the main branch
* [x] `git rename-branch dependent-feature` properly updates the branch name in the dependency list
* [x] `git new-pull-request` sets the branch to merge into to `dependent-feature` instead of main
* [x] `git-prune-branches` removes pruned branches from the branch hierarchy
* [x] `git-kill` points child branches of the killed branch to the parent of the killed branch

kevgo:
We want to provide a robust and safe solution for our branching model:
* don't allow shipping branches which depend on unshipped branches
* when shipping a branch that has dependent branches, update the dependent branches to be dependent on main now.
* sync branches in the order in which they were cut

In order to achieve that, we need to know the exact hierarchy of branches.

kevgo:
Preliminary research indicates that Git treats all branches as equal: branch A can be seen as cut from branch B, or vice versa. In our branching model feature branches are not equal. If feature branch B is cut from feature branch A, this relationship is not reciprocal, i.e. feature branch A shouldn't be seen as cut from feature branch B. 

`git reflog -g <branch-name> | grep 'branch: Created from' | cut -d ' ' -f 6` sometimes prints the name of the branch from which the given branch was cut. Sometimes it just prints `HEAD` though, which isn't helpful. 
Also, the reflog seems only available on the local machine, and it only spans a certain number of days before it gets garbage collected.

Git does therefore not reliably provide the needed exact and reliable hierarchy of branches. 
Achieving this seems only possible if we keep a dedicated log of that information on our side.

kevgo:
Currently, the branch hierarchy list needs to answer the following questions: 
* what is the parent for branch X?
* what are the dependents branches for branch X?

We have discussed two ways to implement our branch hierarchy list:

1. create a new orphaned branch called `gt-data` (or something similar) that contains files which describe the branch hierarchy list.
[...]

PR #534, opened 2015-06-10
Feature branch nesting for git-sync

kevgo:

Since I had 8 consecutive hours flying back from Germany 2 days ago, here comes the big one: support for nested feature branches!

Since this is such a massive feature, I am dividing and conquering it in the smallest possible increments that keep GT functional. This change makes `git-sync` work with nested branches. If the parent branch is unknown, it asks the user. It stores the information in the Git config. I added copious developer documentation that should explain how this works in detail.

Btw its amazing how useful TDD was here. After a few drawings on paper, I was able to sketch out at least the gist of which commands this should run in a feature spec upfront. This together with the `@debug-commands` flag was super useful to debug the code until it worked. Good times.  :)

implements #529


features/step_definitions/branch_configuration_steps.rb
```diff
@@ -0,0 +1,7 @@
+Given(/^Git Town has no branch hierarchy information for "(.*?)"$/) do |branch_names|
+  Kappamaki.from_sentence(branch_names).each do |branch_name|
+    run_shell_command "git config --unset git-town.branches.parent.#{branch_name}"
+    run_shell_command "git config --unset git-town.branches.parents.#{branch_name}"
+  end
+end
+
```
kevgo: fixed

kevgo:
FYI I have added verification for the branch information known to Git Town in https://github.com/Originate/git-town/commit/cc0954e438d7ada19752a1605e9eb70ae112214c

allewun:
I'm late to this, but I think "parents" is a bit misleading. With the way you've defined hierarchical branches, branches can only have one parent. "Ancestors" seems to be more fitting, what do you think?

kevgo:
You are right. I'll change it. Good catch!

PR #543, opened 2015-06-15
Remove specs without open changes

kevgo:

Removes some leftover specs without open changes.

implements #506


kevgo:

PR #549, opened 2015-06-19
Nested feature branch support for git-prune-branches

kevgo:

This is the last command to update

implements #529


kevgo:

PR #560, opened 2015-06-22
Disable Rubocop line length checking

kevgo:

Disables line length checking in Rubocop.
Some step definitions are getting really long.
I know that there is a way to disable this only in certain directories, but that didn't work on my machine, 
so this is a temporary fix until we figure out how to do this. I don't want this to hold up the review process.


kevgo:
This is the last PR in this series, and it passes the tests.

PR #572, opened 2015-07-06
Perennial branches

kevgo:

Reincarnation of #571 (which was closed after renaming the branch)


RELEASE_NOTES.md
```diff
@@ -101,7 +101,7 @@
 * **git sync:** pushes tags to the remote when running on the main branch
   ([#68](https://github.com/Originate/git-town/issues/68),
    [71b607](https://github.com/Originate/git-town/commit/71b607988c00e6dfc8f2598e9b964cc2ed4cfc39))
-* **non-feature branches:** cannot be shipped and do not merge main when syncing
+* **perennial branches:** cannot be shipped and do not merge main when syncing
```
allewun: Is it a good idea to rewrite history? Or do you think the impact is minimal (small userbase right now) and it's better just to remove all references to `non-feature`?

kevgo: Good point. Reverted.

allewun:
We should also remember to note in the readme that there was a name change

kevgo:
We have no CHANGELOG, and we'll compile the release notes when we make a new release (in the next couple of days).

PR #582, opened 2015-07-10
initial website commit

redouglas:
@kevgo @charlierudolph 

Here is my first stab at the website. Most content is brought over from the other site Kevin set up so it may be slightly out of date by now.

To run:
- Install Harp (npm install -g harp)
- cd into `/website` and do `harp server`
- access the site by going to `http://localhost:9000`

I intentionally left the internal site pages as markdown files to make things easier for you guys to edit. This of course also forces the layout to be a bit simplistic. See what you think. 

@kevgo I will need a script from you for the home page terminal script (or if you want you can do it yourself) All that's there now is a CSS mockup. 

This PR doesn't deal with the gh-pages branch quite yet. I would like to get the site to a place where you guys like it before bothering with the publishing.


website/_harp.json
```diff
@@ -0,0 +1,17 @@
+{
+  "globals" : {
+    "site_name"   : "Git Town",
+    "site_description" : "Git Town.",
+    "site_keywords"    : "git, gittown, town",
```
kevgo: Beautiful keywords!  :)

kevgo:
Great, thanks for doing this @redouglas! If you git-:ship: it, I'll add the tooling around it and fine-tune the text.

redouglas:
@kevgo Shipped - please ping me when it comes time to package for deployment. I will need to do a cursory check that all the URLs work properly. Also - don't forget about the front-page hero code! The bash shell isn't doing anything fancy yet. :smile:

kevgo:
This looks so cool! One point for you, @redouglas, you are a master of the craft! Can we add the screencast video to where the bash code is right now? In the future we can add more animations around what the individual commands do. Here is the embed code:

```html
<iframe src="https://drive.google.com/a/originate.com/file/d/0BxFxxboeZOPlcjRiVjV5b0c4Qmc/preview" width="640" height="480"></iframe>
```

PR #694, opened 2015-10-24
Include rename-branch git town help command list

mjhm:
@kevgo @charlierudolph


kevgo:
Ha, good catch, thanks!

PR #696, opened 2015-10-28
added tests

aricahunter:
git town config --setup will now still prompt you to enter in values for configuration parameters if values are already set. For main-branch-name it'll show you the current value (if there is one) and if you press enter it'll just use the current value. For perennial-branch-name, it'll show you the current value (if there is one) and if you press enter, it'll change it to be none.

### DIFF
--- a/features/git-town/configuration/already_set.feature
+++ b/features/git-town/configuration/already_set.feature
@@ -1,8 +1,8 @@
 Feature: listing the configuration
 
-  As a user configuring Git Town,
+  As a user running the Git Town configuration wizard,
   I want to see the existing configuration values
-  and be able to change them.
+  So that I can change it more effectively
 
 
   Background:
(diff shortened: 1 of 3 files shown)


features/git-town/configuration/already_set.feature
```diff
@@ -0,0 +1,34 @@
+Feature: listing the configuration
+
+  As a user configuring Git Town,
```
kevgo: As a user running the Git Town configuration wizard

features/git-town/configuration/already_set.feature
```diff
@@ -0,0 +1,34 @@
...
+  I want to see the existing configuration values
+  and be able to change them.
```
kevgo: So that I can change it more effectively

features/git-town/configuration/already_set.feature
```diff
@@ -0,0 +1,34 @@
...
+
+  Background:
+    Given I have branches named "production" and "qa"
+
+
+  Scenario: everything is configured
+    Given I have configured the main branch name as "main"
+    And my perennial branches are configured as "qa"
+  	When I run `git town config --setup`
+      Then I see
```
kevgo: Wow, what's up with the indentation in this scenario?  :)

features/git-town/configuration/already_set.feature
```diff
@@ -0,0 +1,34 @@
...
+        """
+
+
+  Scenario: empty input
+    Given I have configured the main branch name as "main"
+    And my perennial branches are configured as "qa"
+    When I run `git town config --setup` and enter "" and ""
+    And my repo is configured with the main branch as "main"
+    And my repo is configured with no perennial branches
+
```
kevgo: Please add other scenarios for non-empty input here.

kevgo:
:ship:

Issue #739, opened 2016-08-23
Using ssh with multiple keys to clone repos disable use of 'new-pull-request' and 'repo' commands.


avaz:
Hi,

I have different accounts in Github and Bitbucket and to access them I use ssh with multiple keys. So in my ssh config file I have something like:

```
Host bitbucket-as-account2
  HostName bitbucket.org
  User git
  IdentityFile /Users/anderson/.ssh/id_rsa.pub
  IdentitiesOnly yes

Host github-as-account1
  HostName github.com
  User git
  IdentityFile /Users/anderson/.ssh/dev_andersonvaz_rsa.pub
  IdentitiesOnly yes
```

And to clone repos of these servers I do something like:
`git clone git@github-as-account1:user/repo.git`

Everything works except for the `new-pull-request` and `repo` commands that raise the error bellow:

```
Error
  Unsupported hosting service.
  This command requires hosting on GitHub or Bitbucket.
```

It can't identify which server is being used.
I don't think that it's a bug so I will let this here as a suggestion.
If was possible to specify which hosting service the pull request should be created or automatically resolves from ssh `config` file which service the repo is using would be nice.

kevgo:
Thanks for letting us know! This is probably just because our current regex isn't flexible enough to ignore SSH identities in the Git remote URLs. Will fix in the next version!

PR #740, opened 2016-08-25
Kg ssh identities

kevgo:
@charlierudolph @allewun 

implements #739 

If the normal (and very precise) recognition fails, this now tries a more flexible recognition of the hosting service used. If your Git remote url contains "github", we treat it as Github compatible. I think this is a good compromise between being precise if possible and flexible if necessary.


allewun:
LGTM

src/helpers/string_helpers.sh
```diff
@@ -3,6 +3,18 @@
 # Helper functions for working with strings
 
 
+function contains_string {
+  local haystack=$1
+  # Note: cannot assign $2 to a variable here
+  #       because ShellCheck isn't smart enough to recognize the variable in the regex
```
charlierudolph: I'd rather ignore a shellcheck error instead of changing code since we think this is a missing feature / bug in shellcheck.

kevgo: I agree. Done.

src/drivers/code_hosting/activate.sh
```diff
@@ -14,6 +14,10 @@ function activate_driver_for_code_hosting {
     activate_driver 'code_hosting' 'github'
   elif [ "$origin_hostname" == 'bitbucket.org' ]; then
     activate_driver 'code_hosting' 'bitbucket'
+  elif [ "$(contains_string "$origin_hostname" github)" == true ]; then
```
charlierudolph: I don't like this as it. Its not a bad assumption but feels wrong to me to do. Ideally I'd like us to just read the ssh identities file to get the actual hostname that is associated with the custom host. 

Thats definitely harder and more work so I'm okay with this as is, if we document it somewhere that if you are using ssh identities and want it use github / bitbucket, ensure the custom hostname includes that.

kevgo:
All comments addressed

src/drivers/code_hosting/activate.sh
kevgo: These things are not mutually exclusive. We can try to read the SSH identities, and if that works great. If not, we can still do the heuristic approach that I'm adding here. That's another ticket in my opinion, which would need more research and development. I think for the time being here this is not very elegant, but pragmatic. And it doesn't make things worse, since one cannot use Git Town with custom Git remotes at all right now.

kevgo:
@charlierudolph ping

src/drivers/code_hosting/activate.sh
charlierudolph: I don't think the second half of my comment was addressed. I'd like this documented somewhere

kevgo: I have created #743 to track this.

kevgo:
All comments addressed.

src/drivers/code_hosting/activate.sh
charlierudolph: Thanks for the issue. What I was referring to though was documenting on "git new-pull-request" / "git-repo" that if you are using ssh identity that points to github/bithucket, the identity name needs to include github/bitbucket

kevgo: @charlierudolph good point. Added documentation for this. What do you think?

charlierudolph: Docs look good.

src/drivers/code_hosting/activate.sh
```diff
@@ -14,6 +14,10 @@ function activate_driver_for_code_hosting {
...
+    activate_driver 'code_hosting' 'github'
+  elif [ "$(contains_string "$origin_hostname" bitbucket)" == true ]; then
+    activate_driver 'code_hosting' 'bitbucket'
```
charlierudolph: This isn't tested or documented. Can we please add both or remove this?

kevgo: It is, in features/git-town-new-pull-request/bitbucket.feature in this PR:

kg_ssh_identities_by_kevgo_ _pull_request__740_ _originate_git-town

charlierudolph: Auhh, sorry my mistake. Thanks for updating the documentation

PR #759, opened 2016-11-12
Ruby 2.3.1

kevgo:
Updating to Ruby 2.3.1

@charlierudolph @allewun


kevgo:
Tests are passing now

PR #791, opened 2017-02-21
Rubopcop 2.3

dgjnpr:
update Rubocop linter to use Ruby 2.3 (the version used in rbenv file) and update all files to pass newer lint tests


kevgo:
Looks good to me. Let's sync this one after #790 is shipped, so that this branch only contains the Ruby changes.

kevgo:
This PR seems stalled. I have created #905 to replace it, it performs the same changes.

PR #1004, opened 2017-08-07
Tertestrial Go configuration

kevgo:
I find myself running the builder and a lot for Go files, since with compiled languages one leans on the type system to help write code. This change is an initial set of steps towards making that easier using Tertestrial.


kevgo:
I'd love to see your setup. My editor runs `golint` and `govet` on save, but not the full `gometalinter` shebang, since that takes too long. I prefer that to be run via Tertestrial.

PR #1045, opened 2017-10-14
Command class

### LINKED ISSUE #881: Extract a struct for running commands
Similar to https://github.com/github/hub/blob/master/cmd/cmd.go

```go
cmd = cmd.New("git", "rev-parse")
cmd.AddParam("--foo")
cmd.FullOutput()  // executes it if it didn't so far, then returns the full output
cmd.Success()     // executes if it didn't so far, then returns the response code
```

kevgo:
resolves #881 

I had to extract a `browsers` file as well due to cyclic dependencies.


src/script/script.go
```diff
@@ -70,7 +70,7 @@ func RunCommand(cmd ...string) error {
 	if runtime.GOOS == "windows" {
 		cmd = append([]string{"cmd", "/C"}, cmd...)
 	}
-	subProcess := exec.Command(cmd[0], cmd[1:]...)
+	subProcess := exec.Command(cmd[0], cmd[1:]...) // #nosec
```
kevgo: This was necessary to get the linter to pass. Not sure why this wasn't an issue before.

kevgo:
@charlierudolph ping

PR #1050, opened 2017-11-20
test coverage

### LINKED ISSUE #1048: test coverage
from https://github.com/avelino/awesome-go/pull/1681

- compile the tool in coverage mode
- run the ruby tests
- extract coverage information

kevgo:
https://github.com/avelino/awesome-go/pull/1681 has suggested measuring test coverage, and I agree.
So, while waiting at the airport and on the plane to Germany, unable to do real work because of sleep deprivation, I figured out how to do this with an external test runner. The way Go measures test coverage internally is seriously cool and well thought though.

resolves #1048


kevgo:
I have disabled the comment feature of coveralls. This is more annoying than helpful.

kevgo:
@charlierudolph this is the big change that the previous PRs were preparing. Please take a look!

kevgo:
The test coverage number is lower than it actually is because when the executable exits with an error, it does not create the coverage file.

kevgo:
This isn't ready for review yet, closing

PR #1054, opened 2017-11-20
Add comments

kevgo:
Some of the existing scripts in the `bin` folder have been missing comments. I think they are an important part of making the code base self-describing.


kevgo:
@charlierudolph ping

PR #1143, opened 2018-05-18
Minor documentation updates

ChrisMM:
* update cucumber website URL
* add Windows install

### DIFF
--- a/website/install.md
+++ b/website/install.md
@@ -34,6 +34,14 @@ please use the manual install and ensure your Git version is 2.7.0 or higher._
 
 ---
 
+# Windows
+
+* Install / Update:
+  * Download a 32-bit or 64-bit executable for Windows from the latest release [here](https://github.com/Originate/git-town/releases).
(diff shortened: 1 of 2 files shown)


website/install.md
```diff
@@ -34,6 +34,14 @@ please use the manual install and ensure your Git version is 2.7.0 or higher._
 
 ---
 
+# Windows
+
+* Install / Update:
+  * Download a 32-bit or 64-bit executable for Windows from the latest release [here](https://github.com/Originate/git-town/releases).
```
kevgo: Download **the** ~~a~~ 32-bit....

website/install.md
```diff
@@ -34,6 +34,14 @@ please use the manual install and ensure your Git version is 2.7.0 or higher._
...
+  * Run .exe file from a command line
```
kevgo: I think a better next step here is to put the file into the path. Then Git will run it. The user actually doesn't run git-town manually.

PR #1146, opened 2018-05-24
Document offline mode

kevgo:


kevgo:
Hey @Bowbaq, not sure if you are aware of this feature which Git Town supports since 6 months. I wrote it on airplanes without Wifi  :)

PR #1198, opened 2019-04-22
[new-pull-request] Proper URL encoding

### LINKED ISSUE #1136: New Pull Request didn't escape '#' in PR URL
Hi!

I just run into an issue while trying to open a new pull request with git-town where the branch name contained '#' and it was not escaped in the URL

I'm using Github for my repo.

Regards!
    --Ariel

allewun:
Fixes #1136

### DIFF
--- a/src/drivers/github.go
+++ b/src/drivers/github.go
@@ -43,11 +43,11 @@ func (d *githubCodeHostingDriver) CanMergePullRequest(branch, parentBranch strin
 }
 
 func (d *githubCodeHostingDriver) GetNewPullRequestURL(branch string, parentBranch string) string {
-	toCompare := url.PathEscape(branch)
+	toCompare := branch
 	if parentBranch != git.GetMainBranch() {
-		toCompare = url.PathEscape(parentBranch) + "..." + url.PathEscape(branch)
+		toCompare = parentBranch + "..." + branch
 	}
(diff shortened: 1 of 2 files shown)


src/drivers/github.go
```diff
@@ -42,9 +43,9 @@ func (d *githubCodeHostingDriver) CanMergePullRequest(branch, parentBranch strin
 }
 
 func (d *githubCodeHostingDriver) GetNewPullRequestURL(branch string, parentBranch string) string {
-	toCompare := branch
+	toCompare := url.PathEscape(branch)
 	if parentBranch != git.GetMainBranch() {
-		toCompare = parentBranch + "..." + branch
+		toCompare = url.PathEscape(parentBranch) + "..." + url.PathEscape(branch)
 	}
 	return fmt.Sprintf("%s/compare/%s?expand=1", d.GetRepositoryURL(), toCompare)
```
kevgo: Rather than calling `url.PathEscape` three times above, you could just call it here on `toCompare`.
```suggestion
	return fmt.Sprintf("%s/compare/%s?expand=1", d.GetRepositoryURL(), url.PathEscape(toCompare))
```

allewun: 👍 I originally thought that the ellipses would get encoded too.

PR #1200, opened 2019-04-24
Update docs

allewun:
- Added a description for `git offline`
- Various minor doc tweaks
- Use HTTPS links where possible

### DIFF
--- a/documentation/commands/repo.md
+++ b/documentation/commands/repo.md
@@ -5,6 +5,7 @@ Opens the repository homepage
 </blockquote>
 
 <a textrun="command-description">
+
 Supported only for repositories hosted on [GitHub](http://github.com/),
 [GitLab](http://gitlab.com/), and [Bitbucket](https://bitbucket.org/).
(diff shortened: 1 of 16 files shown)


documentation/commands/repo.md
```diff
@@ -5,6 +5,7 @@ Opens the repository homepage
 </blockquote>
 
 <a textrun="command-description">
+
```
allewun: @kevgo Seems like the GH markdown parser needs this newline here, otherwise the links below will not be parsed.

Semi-related: I tried to use a `<p>` tag here instead of `<a>` as that seems more semantically appropriate, but TextRunner doesn't like that. I took a brief look at TR to try to find out why but didn't reach a conclusion yet. Is that by design?

kevgo: Good find! Yes there is a need for empty lines here or there to make the text look good on GitHub and work in Text-Runner. 

The Markdown parser that Text-Runner uses treats `<p>` tags in weird ways that don't allow to use them as active tags as of now. If it becomes painful enough to not have support for them, I can look into it more.

PR #1211, opened 2019-08-06
Prettify more file types

kevgo:
Prettier can standardize a lot more files than Markdown. This adds prettification of `yml`, `json`, `js`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,14 @@ fix-cucumber:  # auto-fixes all Cucumber lint issues
 	bundle exec cucumber_lint --fix
 
 fix-markdown:  # auto-fixes all Markdown lint issues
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -54,7 +61,14 @@ lint-go:  # lints the Go files
...
 lint-markdown: build  # lints the Markdown files
-	node_modules/.bin/prettier -l "{,!(vendor)/**/}*.md"
+	@find . -type f \( \
+		-path '**/*.md' -o \
+		-path '**/*.yml' -o \
+		-path '**/*.json' -o \
+		-path '**/*.js' \) | \
+		grep -v node_modules | \
+		grep -v vendor | \
+		xargs node_modules/.bin/prettier -l
```
kevgo: This find command probably doesn't work on vanilla Windows, but we can deal with that later.

PR #1238, opened 2019-10-20
Better debug output

kevgo:


kevgo:
To be fair, the way Godog currently works is pretty idiomatic. Go intentionally doesn't provide assertions in the standard library. Test code is supposed to add meaningful descriptions of the error and then return it to signal failure. More background:
- https://golang.org/doc/faq#testing_framework
- https://golang.org/doc/faq#assertions

What I could see making sense here is Godog exposing an instance of the [built-in testing package](https://golang.org/pkg/testing).

PR #1241, opened 2019-10-26
Update golangci-lint

kevgo:


kevgo:
`wsl` is a pretty immature linter for whitespace that would enforce a bunch of newlines which I don't agree with. Hence I disable it.

PR #1280, opened 2019-12-10
Config option to delegate remote branch deletion to the hosting service

### LINKED ISSUE #1210: Ship command fails if "Automatically delete head branches" is enabled

kevgo:
resolves #1210


documentation/commands/ship.md
```diff
@@ -31,6 +31,12 @@ GitHub API. Setup:
    squash merge via the GitHub API. It will also update the base branch for any
    pull requests against that branch.
 
+If you use
+[GitHub's feature to automatically delete head branches](https://help.github.com/en/github/administering-a-repository/managing-the-automatic-deletion-of-branches),
+set `git-town.delete-remote-branches` to `false` in your local or global Git
```
charlierudolph: Thoughts on naming this `git-town.ship-delete-remote-branch`. Like the ship prefix as it only effects this command and don't think this needs to be plural as a single invocation will only effect one branch

kevgo: I like it! Renamed.

src/git/config.go
```diff
@@ -160,6 +160,15 @@ func (c *Configuration) GetCodeHostingOriginHostname() string {
 	return c.getLocalConfigValue("git-town.code-hosting-origin-hostname")
 }
 
+// GetDeleteRemoteBranches indicates whether to delete the remote branch after shipping.
+func (c *Configuration) GetDeleteRemoteBranches() bool {
```
charlierudolph: Thoughts on renaming this to `ShouldShipDeleteRemoteBranch`? (similar to `ShouldNewBranchPush` / `ShouldSyncUpstream`?

kevgo: Sorry I forgot to see this comment until now. Great catch to match the naming to the existing steps. Changed.

kevgo:
Please hold off from further reviews of this PR. I can make the end-to-end test for this feature more accurate once this is converted to Go.

kevgo:
@charlierudolph converting the `git ship` specs to Go yesterday has been unblocked this PR. I have finalized it this morning and need another 👀 from you. Thanks!

features/git-town-ship/current_branch/on_feature_branch/without_open_changes/delete_remote_branch_disabled.feature
```diff
@@ -0,0 +1,58 @@
+Feature: Skip deleting the remote branch when shipping the current branch
```
charlierudolph, line 1: These tests don't really make sense to me as this option only really works when shipping through the github API right? Otherwise the PR wouldn't be merged and thus the branch not auto deleted?

charlierudolph: That said, I'm not certain how we would test this otherwise.

kevgo: Good point! After thinking about this for a while, I realized that the only way to resolve this is to build this the right way 😄  It is not up to Git Town to have knowledge or opinions on whether the remote server deletes the branch or not. GitHub isn't the only remote that does such things. There might be other types of remotes who do something with shipped branches. One could build this into any Git server via a post-merge hook or bot. All that Git Town can do here is provide a setting to not delete remote branches, and apply it consistently. 

This results in correct behavior in all currently known circumstances:
- when GitHub config is set up, and we are ONLINE, it should use this setting
- when GitHub config is set up, and we are OFFLINE, it never deletes the branch
- when GitHub config is NOT set up, and we are ONLINE, it should use this setting
- when GitHub config is NOT set up, and we are OFFLINE, it never deletes the branch

If the user doesn't like any of this behavior, then they can change the configuration.

I have updated the documentation to make it clear that Git Town always applies this setting.

PR #1301, opened 2020-03-23
Update dependencies

kevgo:


kevgo:
Hard to update since Travis-CI is stuck on an old Node version. Probably easier to migrate to CircleCI at this point.

PR #1318, opened 2020-04-24
Make script.OpenInBrowser print the URL it's trying to open

tallpants:
Closes https://github.com/Originate/git-town/issues/1317

### DIFF
--- a/src/script/script.go
+++ b/src/script/script.go
@@ -32,6 +32,7 @@ func ActivateDryRun() {
 
 // OpenBrowser opens the default browser with the given URL.
 func OpenBrowser(url string) {
+  fmt.Printf("Opening URL: %s\n", url)
 	command := browsers.GetOpenBrowserCommand()
 	err := RunCommand(command, url)
 	if err != nil {
--- a/src/script/script.go
+++ b/src/script/script.go
@@ -32,7 +32,7 @@ func ActivateDryRun() {
 
 // OpenBrowser opens the default browser with the given URL.
 func OpenBrowser(url string) {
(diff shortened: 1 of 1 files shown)


tallpants:
I'm not able to get goimports running for some reason -- would you mind checking this out and running it @kevgo? Really sorry about this 😕

src/script/script.go
```diff
@@ -32,6 +32,7 @@ func ActivateDryRun() {
 
 // OpenBrowser opens the default browser with the given URL.
 func OpenBrowser(url string) {
+  fmt.Printf("Opening URL: %s\n", url)
```
kevgo, line 35: Sorry for the not-helpful-at-all error message from `goimports`, @tallpants! Goimports gets hung up that you use two spaces for indentation here, and insists that you use a single tab character instead. You can run `gofmt -s -w src` to fix this in your branch. You should have `gofmt` since it is a part of the standard Go toolchain. This suggested fix here on GitHub should take care of this issue as well:  
```suggestion
	fmt.Printf("Opening URL: %s\n", url)
```

kevgo: Btw as a reference, here is the documentation for `goimports` including instructions for installing it, in case you need it later again: https://godoc.org/golang.org/x/tools/cmd/goimports

tallpants:
Looks like travis decided to not work anymore?

kevgo:
We are in the middle of some infrastructure changes. I have merged your changes in. Thanks for your contribution!

tallpants:
Thanks @kevgo!

PR #1391, opened 2020-05-08
Convert new-pull-request and repo specs to Go

kevgo:
Opening for visibility that I'm working on this.

### DIFF
--- a/test/mocking_shell.go
+++ b/test/mocking_shell.go
@@ -57,7 +57,7 @@ func (ms *MockingShell) MockGit(version string) error {
 	// create "bin" dir
 	err := os.Mkdir(ms.binDir, 0744)
 	if err != nil {
-		return fmt.Errorf("cannot create mock bin dir: %w", err)
+		return fmt.Errorf("cannot create mock bin dir %q: %w", ms.binDir, err)
 	}
 	// write custom Git command
 	content := fmt.Sprintf(`#!/usr/bin/env bash\n\nif [ "$1" = %q ]; then\n  echo "git version %s"\nfi`, version, version)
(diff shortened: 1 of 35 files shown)


test/mocking_shell.go
```diff
@@ -17,39 +17,73 @@ import (
 // - Temporarily override certain shell commands with mock implementations.
 //   Temporary mocks are only valid for the next command being run.
 type MockingShell struct {
-	workingDir            string // the directory in which this runner runs
-	homeDir               string // the directory that contains the global Git configuration
-	tempShellOverridesDir string // the directory that stores the mock shell command implementations, ignored if empty
+	workingDir     string // the directory in which this runner runs
+	homeDir        string // the directory that contains the global Git configuration
+	binDir         string // the directory that stores the mock shell command implementations, ignored if empty
```
kevgo, line 22: This simplifies the infrastructure we use for mocking shell commands. We actually never mock more than a single command. I store mocked commands are in `GitEnvironment.Dir/bin` so that they are unique per test.

This also saves us from having to remove this directory at the end of the test.

test/mocking_shell.go
```diff
@@ -17,39 +17,73 @@ import (
...
+// MockCommand adds a mock for the command with the given name.
+func (ms *MockingShell) MockCommand(name string) error {
+	// create "bin" dir
+	err := os.Mkdir(ms.binDir, 0744)
+	if err != nil {
+		return fmt.Errorf("cannot create mock bin dir: %w", err)
 	}
-	return ioutil.WriteFile(ms.tempShellOverrideFilePath(name), []byte(content), 0744)
+	// write custom "which" command
+	content := fmt.Sprintf("#!/usr/bin/env bash\n\nif [ \"$1\" == %q ]; then\n  echo %q\nelse\n  exit 1\nfi", name, filepath.Join(ms.binDir, name))
```
kevgo, line 40: Simplified which command that returns the answer right away without filesystem access.

PR #1414, opened 2020-05-14
Create child feature branches without Git Town

kevgo:
Calling Git Town in test setup pollutes the undo stack of the test repo. Which breaks some `git ship` tests that expect an empty undo stack. 

In the future, this would ideally be implemented using the Git Town business logic in the tests.

### DIFF
--- a/features/git-town-set-parent-branch/set_parent_branch.feature
+++ b/features/git-town-set-parent-branch/set_parent_branch.feature
@@ -8,6 +8,7 @@ Feature: update the parent of a nested feature branch
   Background:
     Given my repository has a feature branch named "parent-feature"
     And my repository has a feature branch named "child-feature" as a child of "parent-feature"
+    And I am on the "child-feature" branch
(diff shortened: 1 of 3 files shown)


features/git-town-set-parent-branch/set_parent_branch.feature
```diff
@@ -8,6 +8,7 @@ Feature: update the parent of a nested feature branch
   Background:
     Given my repository has a feature branch named "parent-feature"
     And my repository has a feature branch named "child-feature" as a child of "parent-feature"
+    And I am on the "child-feature" branch
```
kevgo, line 11: The step to create child branches implicitly changed the currently checked out branch. This test was relying on this unintended side effect. This side effect no longer exists, so we make the branch we expect on explicit.

kevgo:
This also makes the tests faster since we do less checking out of branches (which changes files in the workspace).

PR #1418, opened 2020-05-14
Cross-platform Cucumber test runner

kevgo:
I have evaluated a number of options and ended up on this one as the most convenient and least hacky one. With this setup you run:

- `go test` for all cukes in parallel
- `godog` for individual cukes (with nice auto-completion of filenames in Fish)
- `go test ./src/... ./test/...` for the unit tests

This still allows debugging individual cukes as well.


kevgo:
Another benefit of this is that it doesn't leave a `godog.test` binary in the current directory when the Cucumber tests get interrupted.

PR #1447, opened 2020-05-18
Disable garbage collection for tests

kevgo:
This change disables garbage collection for all tests. Tests don't need garbage collection since they finish quickly, at which point the OS reclaims all memory used by the process anyways. There seem no noticeable changes in memory consumption, indicating that the GC mostly burns CPU checking non-reclaimable memory. On developer machines, this change reduces the CPU load by about 20%. It doesn't affect test runtimes since the Cucumber tests don't fully utilize the CPU and the GC runs mostly in the background, but it helps keep machines cooler and makes the fans spin less/battery last longer. In CPU-constrained environments like on CircleCI, where we only get slices of virtualized CPUs, this results in shorter runtimes. The savings seem congruent to the reduction of work shown in the flamegraphs below.

With GC:

Screenshot 2020-05-18 at 4 19 54 PM

Without GC:

Screenshot 2020-05-18 at 4 09 44 PM


kevgo:
I need this to make further optimizations more precise, so merging in now.

PR #1461, opened 2020-05-24
Make unit tests run on Windows

kevgo:
Now that we don't have Ruby anymore, development on Windows is now easy enough that we can support it.


Makefile
```diff
@@ -76,7 +76,7 @@ u:  # runs only the unit tests for changed code
 	env GOGC=off go test -timeout 5s ./src/... ./test/...
 
 unit:  # runs all the unit tests with race detector
-	env GOGC=off go test -count=1 -timeout 20s -race ./src/... ./test/...
+	env GOGC=off go test -count=1 -timeout 60s -race ./src/... ./test/...
```
kevgo, line 79: The unit tests run abysmally slow on Windows, but at least they run now.

test/tools_unix.go
```diff
@@ -0,0 +1,31 @@
+// +build !windows
```
kevgo, line 1: This compiler directive compiles this file only on non-Windows platforms.

test/tools_windows.go
```diff
@@ -0,0 +1,31 @@
+// +build windows
```
kevgo, line 1: This compiler directive compiles this file only on Windows platforms.

PR #1472, opened 2020-05-27
Basic git.ProdRepo setup

kevgo:
This PR implements the core part of https://github.com/git-town/git-town/pull/1459#discussion_r429648072: a `git.ProdRepo` consisting of a `Silent` runner (that runs Git commands without CLI output using a `command.SilentShell`) and a `Logging` runner (that runs Git commands with CLI output using a `git.LoggingShell`). These two runners share a number of objects that manage shared state.

Each Git Town command creates a `git.ProdRepo` instance and injects that into the step execution engine, which injects it into the `.Run` method of each step. This works pretty well and causes little overhead! 

Opening a PR now while this is still small enough to be reviewed. More conversions to the new world to follow in subsequent PRs.


src/git/logging_shell.go
```diff
@@ -0,0 +1,115 @@
...
+func (shell LoggingShell) MustRun(cmd string, args ...string) *command.Result {
+	return command.MustRun(cmd, args...)
+}
+
+// Run runs the given command in this ShellRunner's directory.
+func (shell LoggingShell) Run(cmd string, args ...string) (*command.Result, error) {
+	shell.PrintCommand(cmd, args...)
+	if dryrun.IsActive() {
+		if len(args) == 2 && cmd == "git" && args[0] == "checkout" {
+			dryrun.SetCurrentBranchName(args[1])
```
kevgo, line 44: In a future PR, I want to move the functionality inside the `dryrun` package into this class and get rid of that package altogether.

src/git/logging_shell.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot split command %q: %w", fullCmd, err)
+	}
+	cmd, args := parts[0], parts[1:]
+	return shell.Run(cmd, args...)
+}
+
+// RunStringWith runs the given command (including possible arguments) in this ShellInDir's directory.
+func (shell LoggingShell) RunStringWith(fullCmd string, options command.Options) (*command.Result, error) {
+	panic("this isn't used")
```
kevgo, line 86: I want to get rid of some of these methods. We should return proper errors almost everywhere.

src/git/logging_shell.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// PrintCommand prints the given command-line operation on the console.
+func (shell LoggingShell) PrintCommand(cmd string, args ...string) {
```
kevgo, line 90: The plan is to that other packages like this one will absorb what's in the current `script` package.

src/steps/checkout_branch_step.go
```diff
@@ -19,7 +19,7 @@ func (step *CheckoutBranchStep) CreateUndoStep() Step {
 }
 
 // Run executes this step.
-func (step *CheckoutBranchStep) Run() error {
+func (step *CheckoutBranchStep) Run(repo *git.ProdRepo) error {
```
kevgo, line 22: I will convert these steps in a subsequent PR, wanted to keep this one short.

PR #1477, opened 2020-05-28
Open links in native browser on ChromeOS

kevgo:
When installing Firefox into the Linux subsystem of ChromeOS, it automatically becomes the default browser for links opened from within Linux. This change makes Git Town always open the native Chrome browser.


kevgo:
Sorry I need this, merging now.

PR #1485, opened 2020-06-01
Improve error logging

blaggacao:
In case of error, it renders the url which fails, see:
```
Error Trace:github_test.go:146
Error:      Received unexpected error:
            GET https://api.github.com/repos/git-town/git-town/pulls?base=feature&state=open: 401 Bad credentials []
Test:       TestGitHubDriver_MergePullRequest
```

### DIFF
--- a/src/command/run_test.go
+++ b/src/command/run_test.go
@@ -14,7 +14,7 @@ import (
 
 func TestCommand_Run(t *testing.T) {
 	res, err := command.Run("echo", "foo")
-	assert.Nil(t, err)
+	assert.NoError(t, err)
 	assert.Equal(t, "foo\n", res.Output())
 }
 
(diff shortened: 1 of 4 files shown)


src/command/run_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func TestCommand_Run(t *testing.T) {
 	res, err := command.Run("echo", "foo")
-	assert.Nil(t, err)
+	assert.NoError(t, err)
```
kevgo, line 17: Nice idea to use `NoError` here! I wasn't aware of this. I'm going to change this everywhere in the codebase.

src/drivers/github_test.go
```diff
@@ -35,8 +35,7 @@ func TestGitHubDriver_CanMergePullRequest(t *testing.T) {
 	defer teardown()
 	httpmock.RegisterResponder("GET", currentPullRequestURL, httpmock.NewStringResponder(200, `[{"number": 1, "title": "my title" }]`))
 	canMerge, defaultCommintMessage, pullRequestNumber, err := driver.CanMergePullRequest("feature", "main")
-
-	assert.Nil(t, err)
+	assert.NoError(t, err)
```
kevgo, line 38: Thanks for removing the empty lines inside function bodies. They should all go away.

blaggacao: I could do a quick fixup on all places wothin this very pr.

src/command/run_test.go
blaggacao: I already did a global replace in the second commit.

src/drivers/github_test.go
kevgo: Better to do it as a separate PR. Each PR should "do" only one thing. Otherwise your feature branches live too long and make too many changes, which leads to merge conflicts and long review times. Thanks!

blaggacao:
Retargeted to master, converted in pure single topic change.

kevgo:
Excellent change, thanks so much @blaggacao!  👍

kevgo:
I'm not sure if you use this branch as a parent branch of other feature branches, so I leave it around here. Feel free to delete it if you don't need it anymore. Thanks!

PR #1488, opened 2020-06-02
Free from godot lint errors

blaggacao:


blaggacao:
@kevgo It is not suppressed, this is why they popped up all over the place. Maybe `golangci-lint` is not tagged, which would be a cool feature. :wink:

kevgo:
@blaggacao we have "tagged" golangci-lint to `1.23.8` [here](https://github.com/git-town/git-town/blob/master/Makefile#L63). This is also used on [CI](https://github.com/git-town/git-town/blob/master/.circleci/config.yml#L11). But I admit it's not a very elegant solution. Do you know a way to do this better?

blaggacao:
Yesterday, I tried to use the golang-ci docker image, work perfect locally, but that didn't just work with circleci, so I refrained.

kevgo:
CircleCI is already running everything inside Docker, so running Docker inside Docker probably doesn't work. The way to do this on CircleCI is to create another job that uses the Docker image you want. We currently run 4 different jobs in 4 different Docker containers:

Screenshot 2020-06-03 at 8 37 29 PM

[.circleci/config.yml](https://github.com/git-town/git-town/blob/master/.circleci/config.yml) defines the jobs, which Docker images they use, which commands run in each image, and in which order they run. One would add the golangci-lint Docker image as another job that runs in parallel with `markdown`, `go`, and `go-s`.

For local development, I prefer having as few dependencies as possible, so I don't want to make Docker a requirements for developers at this point. Git Town development only requires Go and optionally Node for Markdown linting. That should be available on every developer machine.

PR #1491, opened 2020-06-02
Bump golangci-lint verion

blaggacao:
> Can we add it to our Makefile?

#1488 Indeed I was probably using a newer version (globally) ...


kevgo:
I have created and shipped #1496 to do this, which also fixes additional lint errors. So I'm going to close this particular PR. Thanks for letting me know about the new version! It's amazing how many new linters for Go have been created in the last few weeks!

Btw in case it comes up in the future, if you log into CircleCI with your GitHub account, you should be able to see the builds and the build errors there.

PR #1492, opened 2020-06-02
Add autocompletion for bash, zsh, fish, powershell

### LINKED ISSUE #875: Improve fish autocompletion
https://github.com/github/hub/blob/master/etc/hub.fish_completion

### LINKED ISSUE #1208: Suggestion: add git-town alias library to oh-my-zsh
Hi, git-town user for 3+ years.  Also, use oh-my-zsh and thought to open a library as one doesn't exist. WIP PR here - https://github.com/robbyrussell/oh-my-zsh/pull/7989

Reaching out for suggestions/thoughts.

### LINKED ISSUE #1302: Fish completion overrides existing completion
The installed fish completion eats all `git` expansions, but instead it should delegate anything that isn't `git town <command>` to the preinstalled git completion.

I'm not an expert on fish completions so I'm not sure how to resolve this myself, but currently you can't have completions for both `git` and `git town` concurrently.

### LINKED ISSUE #1482: Add bash / zsh completion
https://github.com/spf13/cobra/blob/master/bash_completions.md
and
https://github.com/spf13/cobra/blob/master/zsh_completions.md

blaggacao:
closes #1482 
closes #1208 (somewhat)
closes  #875
closes #1302


blaggacao:
image

blaggacao:
Maybe it should be fine to leave instructions instead a magi one-in all command, or would this go against the philosophy of git town?

src/cmd/completion.go
```diff
@@ -0,0 +1,69 @@
...
+
+Bash:
+
+$ source <(git-town completion bash)
+
+# To load completions for each session, execute once:
+Linux:
+  $ git-town completion bash > /etc/bash_completion.d/git-town
+MacOS:
+  $ git-town completion bash > /usr/local/etc/bash_completion.d/git-town
```
blaggacao, line 24: Global?

src/cmd/completion.go
```diff
@@ -0,0 +1,69 @@
...
+
+Zsh:
+
+$ source <(git-town completion zsh)
+
+# To load completions for each session, execute once:
+$ git-town completion zsh > /usr/share/zsh/vendor-completions/_git-town
```
blaggacao, line 31: Global?

src/cmd/completion.go
```diff
@@ -0,0 +1,69 @@
...
+
+Fish:
+
+$ git-town completion fish | source
+
+# To load completions for each session, execute once:
+$ git-town completion fish > ~/.config/fish/completions/git-town.fish
```
blaggacao, line 38: Home?

features/git-town-completion/bash-completion.feature
```diff
@@ -0,0 +1,35 @@
...
+      """
+    And my computer now has a Git autocompletion file
+
+
+  Scenario: with an existing Git autocompletion file
+    Given my computer has an existing Git autocompletion file
+    When I run "git-town completion bash"
+    Then it prints the error:
+      """
+      Git autocompletion for Bash shell already exists
```
kevgo, line 33:
```suggestion
      Bash autocompletion already exists
```

features/git-town-completion/bash-completion.feature
```diff
@@ -0,0 +1,35 @@
...
+    When I run "git-town completion bash"
+    Then it prints:
+      """
+      Git autocompletion for Bash shell installed
+      """
+    And my computer now has a Git autocompletion file
+
+
+  Scenario: with an existing Git autocompletion file
+    Given my computer has an existing Git autocompletion file
```
kevgo, line 29: What do you think about this:

```suggestion
    Given my computer has an existing Git autocompletion file for Bash
```

features/git-town-completion/powershell-completion.feature
```diff
@@ -0,0 +1,35 @@
+Feature: Installing Powershell Shell autocomplete definitions
```
kevgo, line 1:
```suggestion
Feature: Installing Powershell autocomplete definitions
```

features/git-town-completion/powershell-completion.feature
```diff
@@ -0,0 +1,35 @@
...
+
+  As a Powershell shell user
```
kevgo, line 3:
```suggestion
  As a Powershell user
```

features/git-town-completion/powershell-completion.feature
```diff
@@ -0,0 +1,35 @@
...
+  I want to be able to install the autocomplete definitions for Git Town with an easy command
```
kevgo, line 4:
```suggestion
  I want to use Git Town with autocompletion
```

features/git-town-completion/bash-completion.feature
```diff
@@ -0,0 +1,35 @@
+Feature: Installing Bash Shell autocomplete definitions
+
+  As a Bash shell user
```
kevgo, line 3:
```suggestion
  As a Bash user
```

features/git-town-completion/bash-completion.feature
```diff
@@ -0,0 +1,35 @@
...
+  I want to be able to install the autocomplete definitions for Git Town with an easy command
```
kevgo, line 4:
```suggestion
  I want to use Git Town with autocompletion
```

src/cmd/completion.go
kevgo: Do you mean install globally for all users? I think most people would prefer to do this locally for each user.

src/cmd/completion.go
kevgo: I don't understand your comment. The path looks okay to me.

src/cmd/completion.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var completionNoDesc bool
+
+var completionCmd = &cobra.Command{
+	Use:   "completion [bash|zsh|fish|powershell]",
```
kevgo, line 12: The convention we have been using is `<` for required arguments and `[` for optional arguments.

```suggestion
	Use:   "completion <bash|zsh|fish|powershell>",
```

website/install.md
```diff
@@ -64,7 +64,7 @@ ensure your Git version is 2.7.0 or higher._
 
 #### Install autocompletion
 
-- for [Fish shell](http://fishshell.com): `git town install-fish-autocompletion`
+- check instructions from: `git town completion --help`
```
kevgo, line 67:
```suggestion
- follow instructions given by `git town completion --help`
```

kevgo:
The screenshot looks really cool! Is this how the autocompletion looks on your machine? Which shell do you use?

blaggacao:
I'm using zsh with `oh-my-zsh`

src/cmd/completion.go
blaggacao: Right, for dotfile repos.

src/cmd/completion.go
blaggacao: for dotfile repos..

blaggacao:
I need to change the feature description still, as it now requires the user to follow the instructions and place the files in his file system.

I researched and couldn't find a way to put those auto completions reliably into the home folder, there is no fixed convention, as it seems, so I'd prefer to put all into the  host folders, but with a hint that users can put them into the homer folder, e.g. when they are power users that use git for dot-files.

To be unit consistent, therefore, I'd give instructions for fish in a global location, too.

blaggacao:
Btw, it would be nice to have an extension for https://kakoune.org/, too.

kevgo:
Are there more reliable locations for global folders? In my experience they differ by distro as well, especially if you move between different "families" of Unix or Red Hat/Debian/SUSE based distros. I know that fish shell config is always in `~/.config/fish`, but that it can be different for the others (`.bashrc` vs `.bash_profile`). I think in the end it's not Git Town's job to have opinions where the "right" configuration of their users shell is. We could just say that people should run `git-town completions <shell name>` inside the startup script for their shell. Then give hints like "If you don't know where your shell startup script is, try .bashrc".

kevgo:
Kakoune is a text editor, no? What would a Git Town extension for it do?

blaggacao:
It has already a pretty solid git integration, so git town would fit right in... I just learned today about it and tried it a little, will check if I find my way around an eventually provide a `git-town.kak` somehow.

kevgo:
Do you have access to CI at this point?

Error:

```
golangci-lint run --enable-all -D dupl -D lll -D gochecknoglobals -D gochecknoinits -D goconst -D wsl -D gomnd -D dogsled src/... test/...
src/cmd/completions.go:46:36: `thier` is a misspelling of `their` (misspell)
_You might be a power user who has thier dotfiles under version control. Or you
                                   ^
src/cmd/completions.go:49:5: `offical` is a misspelling of `official` (misspell)
the offical docs for your shell._
 ```

**Comment by blaggacao** [comment]
Yes I have access now, for some reason however the features tests all fail locally.... To be debugged another day... For now it's a little back and forth with circleci...

**Comment by blaggacao** [comment]
Since it's hard to check the output, I think we should ask the community for how those shell output and mock that in a step. The idea of mocking powershell, ish, bash and zsh seems just too crazy of a work for me....

**Comment by blaggacao** [comment]
I think that one is a problem with text-runner not parsing inlied code blocks: https://app.circleci.com/pipelines/github/git-town/git-town/875/workflows/7ff0c742-4bdd-477c-b7a0-5268cdb77d0b/jobs/5882/parallel-runs/0/steps/0-107

However the formatting without code blocks is just too aweful. @kevgo What can we do?

**Comment by kevgo** [comment]
Text-Runner ensures that what the content of the Markdown files in `documentation/commands/` is exactly the same as what the CLI tool prints, for example when running `git town help completions`. The CLI shouldn't output Markdown but simple text. I suggest keeping detailed installation instructions out of the CLI output. We can add them to the [official installation instructions](https://www.git-town.com/install.html) on the web. Their source is at https://github.com/git-town/git-town/blob/master/website/install.md. Maybe you add a link to that URL? 

Btw you can run Text-Runner locally via `make lint-md`. The overall setup to match Markdown to CLI output is an experiment and at this point pretty finicky and flimsy. Don't spend too much time on that, it's going to be frustrating. I can help format the documentation.

**Comment by blaggacao** [comment]
Ok, understand the text-runner aspects. However, I'd like to keep instructions on the command line.
I personally hate to have to jump over to some external documentation.

**Comment by kevgo** [comment]
Good point. Let's leave everything in there then. What do you think about keeping the formatting simple for now so that this doesn't block the release, and then fine-tune the formatting in a subsequent PR?

**Comment by blaggacao** [comment]
The features inevitably fail, but I don't see a viable option to mock out all these shells. Maybe we can "trust" cobra to do the autocompletion-generation it right?

File: documentation/commands/completions.md [hunk 13]
```diff
@@ -0,0 +1,50 @@
...
+
+You might be a power user who has their dotfiles under version control. Or you
+might have another motivation to keep those scripts in your home folder. Since
+it sometimes depends on your particular setup, you probably should consult the
+official docs for your shell.</a>
+
+#### Usage
+
+<pre textrun="command-usage">
+git town completions <bash|zsh|fish|powershell>
```
blaggacao, line 49: @kevgo `<bash|zsh|fish|powershell>` seems to be a hard problem
`\<bash|zsh|fish|powershell\>` is not interpreted...

features/git-town-completions/shell-completions.feature
```diff
@@ -0,0 +1,41 @@
+Feature: Rendering Shell autocomplete definitions
+
+  As a shell user
+  I want to use Git Town with autocompletion
+  So that I can use this tool productively.
+
+  Scenario: loading autocompletion in fish
+    Given a fish shell
+    And I run "git-town completions fish | source"
```
kevgo, line 9: I think you are right. We cannot really test this. I would add that it is not the responsibility of our test suite to verify whether Cobra's completion stuff works correctly. They have their own tests for that. For example, [here](https://github.com/spf13/cobra/blob/master/zsh_completions_test.go) is their test for the Zsh completion. If there was a bug in Cobra's completion code, it should break their test suite, not ours. The fix should happen in their codebase. The responsibility of our test suite is to test our business logic. This includes that we call the Cobra completion generator properly. I would just check that it prints a piece of the expected output like `sync:Updates the current branch with all relevant changes` or `function _git-town_sync`.

documentation/commands/completions.md
kevgo: Ah, too bad. Let's change it back from `<` to `[` then. Most people won't know the difference anyways.

blaggacao: ok

src/cmd/completions.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+var completionsNoDesc bool
+
+var completionsCmd = &cobra.Command{
+	Use:   "completions <bash|zsh|fish|powershell>",
+	Short: "Generates completion scripts for Bash, zsh, fish, and PowerShell",
+	Long: `Generates completion scripts for Bash, zsh, fish, and PowerShell
+
+A Git Town user values productivity, so this is for you.
```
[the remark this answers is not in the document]
blaggacao: Just seeking engagement, not trying to treat programmers like machines... But maybe they are true hybrids :D

blaggacao:
Merge ready form my point of view...

src/cmd/completions.go
```diff
@@ -0,0 +1,81 @@
...
+
+$ git-town completions zsh > /usr/share/zsh/vendor-completions/_git-town
+
+Fish:
+
+$ git-town completions fish | source
+
+Persist and autoload on each session:
+
+$ git-town completions fish > /etc/fish/completions/git-town.fish
```
kevgo, line 45: Your friendly QA assistant @kevgo did some testing! The Bash autocompletion works! Fish autocompletion errors out with this: 

```
set: Variable name '__git-town_comp_commandLine' is not valid. See `help identifiers`.

~/.config/fish/completions/git-town.fish (line 60): 
        set __git-town_comp_commandLine (commandline)
        ^
in function '__git-town_prepare_completions'
in command substitution
```

Variable names in Fish cannot contain `-`. If I change the `-` to `_` in the variables (and only here), it works. This seems an upstream issue. We could replace all occurrences of `/__git-town_.*\b/` with that string that has `-` replaced with `_` as a workaround for now.

src/cmd/completions.go
```diff
@@ -0,0 +1,81 @@
...
+
+To enable completions:
+
+Bash:
+
+$ source <(git-town completions bash)
+
+Persist and autoload on each session:
+
+Linux: $ git-town completions bash > /etc/bash_completion.d/git-town
```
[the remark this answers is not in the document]
blaggacao: I'm totally against this, this absolutely pollutes `.bashrc` (if someone was to keep it clean).
Furthermore, on updating completions, since it's appending, it probably would result in unpredictable behavior. Unfortunately, there seems not to be a standard _enough_ place to put those in the home folder.

src/cmd/completions.go
```diff
@@ -0,0 +1,81 @@
...
+
+MacOS: $ git-town completions bash > /usr/local/etc/bash_completion.d/git-town
+
+Zsh:
+
+$ source <(git-town completions zsh)
+
+Persist and autoload on each session:
+
+$ git-town completions zsh > /usr/share/zsh/vendor-completions/_git-town
```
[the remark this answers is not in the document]
blaggacao: That would be suicide... :D
zsh has a bunch of places where to put those.
For oh-my-zsh users it could be: `~/.oh-my-zsh/completions` - but then, if you're not a oh-my-zsh user...

`vendor` is just the place for vendors (if we where to package git-town with the auto-packager tool, this is where it should go)

src/cmd/completions.go
blaggacao: oh... :D

blaggacao: fortunately, this might be a short lived problem: https://github.com/spf13/cobra/pull/1122

blaggacao: since judging by average this should be fixed in a few days, I'd rather say we refrain from a workaround now and trade some work for patience... Or we risk exposing the fish users and do a comment in the instructions meanwhile...

blaggacao: @kevgo If you are ok with the cuationary hint, the let{s ship this and open an issue to track the upstream fix.

blaggacao: Or can we pull in the patches to cobra directly into vendored cobra?

src/cmd/completions.go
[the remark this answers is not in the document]
kevgo: Just to be clear what I mean, this simply adds the string `source <(git-town completions bash)` to the end of the `.bashrc` file. It doesn't add the hundred lines of stuff that `git-town completions bash` outputs. Just this one line. With this in place, each time the user starts a new Bash shell, it runs `git-town completions bash` and evaluates them inside this shell, so that the autocompletions are active.

src/cmd/completions.go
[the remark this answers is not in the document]
kevgo: Okay, I just installed ZSH using HomeBrew for Linux, and it still has `/usr/share/zsh/vendor-completions`. I'm fine going with that.  🙂

src/cmd/completions.go
kevgo: You are a master of navigating open-source projects! How did you find that bug at Cobra? Given that this PR at Cobra is hanging around for 13 days already, let's do what you said and ship this. We will pick up this bug fix when Cobra releases a new version and we update dependencies. Not sure we can mess with the `vendor` folder, Go might checksum it. We can also use Cobra's `master` branch for a while. The Git Town test suite is comprehensive enough to create confidence that everything will work.

kevgo:
Shipped!  🎉

src/cmd/completions.go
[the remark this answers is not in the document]
blaggacao: Nice, thanks you double checked!

src/cmd/completions.go
blaggacao: master branch sounds good, I once heard of doing some alternative overlays in the old go packages world, but now with modules there, I'm still a bit confused in general.

blaggacao:
Thank you that's awesome! - Now my team is ready to rock!

PR #1493, opened 2020-06-02
Enable cobra completions without descriptions

blaggacao:
see: https://github.com/spf13/cobra/pull/1070
only merge once upstream commit merged + released + cobra updated.


kevgo (approved):
This is finally working! Thanks for the contribution! :+1:

PR #1497, opened 2020-06-02
Convert the remaining error assertions

kevgo:
Continues #1485 and replaces all new/other occurrences.


blaggacao:
oops I missed them, sorry. Didn't notice.

PR #1496, opened 2020-06-02
Update to golangci-lint 1.27.0

kevgo:
@blaggacao FYI, this fixes all issues with the current version so you should no encounter any more lint errors.

PR #1500, opened 2020-06-02
Adopt more transparent, declarative and scalable golang ci config

blaggacao:
As recomended on: https://golangci-lint.run/usage/configuration/

Follow up on  #1496

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,16 +1,16 @@
 linters:
   enable-all: true
   disable:
-    - dogsled             # The rationale for disabling dogsled is that all the cases it flags have been false positives in test code that intentionally ignores many return values because they test only one thing.
-    - dupl                # I think that one creates too many false positives as well. Our tests are kind of repetitive, which is intentional. They are supposed to be simple, straightforward scripts.
+    - dogsled # The rationale for disabling dogsled is that all the cases it flags have been false positives in test code that intentionally ignores many return values because they test only one thing.
+    - dupl # I think that one creates too many false positives as well. Our tests are kind of repetitive, which is intentional. They are supposed to be simple, straightforward scripts.
     # - exhaustive        TODO: enable after next release; current release at time of writing is v1.27
-    - gochecknoglobals    # We are currently using some globals. It's a lot more forgivable in a CLI tool than in a library.
-    - gochecknoinits      # I think the CLI framework we use requires them.
-    - goconst             # Our tests contain a ton of hard-coded test strings, for example branch names.
-    - goerr113            # Expose errors properly, once/if approaching library use, as opposed to current cli-only use
-    - gomnd               # Our tests contain hard-coded test data that wouldn't make sense to extract into constants
-    - lll                 # Following gofmt principles, we aren't enforcing a line length at this point.
-    - wsl                 # This linter creates too many false positives. Our policy is to not have any empty lines in code blocks anyways.
+    - gochecknoglobals # We are currently using some globals. It's a lot more forgivable in a CLI tool than in a library.
+    - gochecknoinits # I think the CLI framework we use requires them.
+    - goconst # Our tests contain a ton of hard-coded test strings, for example branch names.
+    - goerr113 # Expose errors properly, once/if approaching library use, as opposed to current cli-only use
+    - gomnd # Our tests contain hard-coded test data that wouldn't make sense to extract into constants
+    - lll # Following gofmt principles, we aren't enforcing a line length at this point.
+    - wsl # This linter creates too many false positives. Our policy is to not have any empty lines in code blocks anyways.
 
 issues:
   # List of regexps of issue texts to exclude, empty list by default.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,16 +1,16 @@
 linters:
   enable-all: true
   disable:
-    - dogsled             # The rationale for disabling dogsled is that all the cases it flags have been false positives in test code that intentionally ignores many return values because they test only one thing.
-    - dupl                # TODO: write rational for disabling here, brought over from Makefile
-    - exhaustive          # TODO: enable after next release; current release at time of writing is v1.27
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -1,11 +1,61 @@
...
+    - asciicheck
+    - bodyclose
+    - deadcode
+    - depguard
+    # - dogsled (TODO: write rational for disabling here, brought over from Makefile)
+    # - dupl (TODO: write rational for disabling here, brought over from Makefile)
+    - errcheck
+    - exhaustive
+    - funlen
+    # - gochecknoglobals (TODO: write rational for disabling here, brought over from Makefile)
```
kevgo, line 15: We are currently using some globals. It's a lot more forgivable in a CLI tool than in a library. I hope most of them become obsolete with the current refactoring that I'm doing. Then we can enable this check again.

.golangci.yml
```diff
@@ -1,11 +1,61 @@
...
+    # - gochecknoinits (TODO: write rational for disabling here, brought over from Makefile)
```
kevgo, line 16: I think the [CLI framework](https://github.com/spf13/cobra) we use requires them.

.golangci.yml
```diff
@@ -1,11 +1,61 @@
...
+    - gocognit
+    # - goconst (TODO: write rational for disabling here, brought over from Makefile)
```
kevgo, line 18: Our tests contain a ton of hard-coded test strings, for example branch names.

.golangci.yml
```diff
@@ -1,11 +1,61 @@
...
+    - gocritic
+    - gocyclo
+    - godot
+    - godox
+    # - goerr113 (TODO: expose errors proerply, once/if approaching library use, as oposed to current cli-only use)
+    - gofmt
+    - goimports
+    - golint
+    # - gomnd (TODO: write rational for disabling here, brought over from Makefile)
```
kevgo, line 27: Our tests contain hard-coded test data that wouldn't make sense to extract into constants.

.golangci.yml
```diff
@@ -1,11 +1,61 @@
...
+    - goprintffuncname
+    - gosec
+    - gosimple
+    - govet
+    - ineffassign
+    - interfacer
+    # - lll (TODO: write rational for disabling here, brought over from Makefile)
```
kevgo, line 34: Following `gofmt` principles, we aren't enforcing a line length at this point. Most code is well formatted, long lines happen only in exceptional cases like when code contains long URLs. The long line is acceptable in those situations.

.golangci.yml
```diff
@@ -1,11 +1,61 @@
...
+    - structcheck
+    - stylecheck
+    - testpackage
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+    - whitespace
+    # - wsl  (TODO: write rational for disabling here, brought over from Makefile)
```
kevgo, line 53: This linter creates too many false positives. Our policy is to not have any empty lines in code blocks anyways.

.golangci.yml
```diff
@@ -1,11 +1,61 @@
...
+  # please, do not use `enable-all`: it's deprecated and will be removed soon.
+  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
+  disable-all: true
+  enable:
+    - asciicheck
+    - bodyclose
+    - deadcode
+    - depguard
+    # - dogsled (The rationale for disabling dogsled is that all the cases it flags have been false positives in test code that intentionally ignores many return values because they test only one thing.)
+    # - dupl (TODO: write rational for disabling here, brought over from Makefile)
```
blaggacao, line 11: @kevgo This one slipped, as it seems.

blaggacao:
I looked in the codebase of golangci-lint and it looks like the enable-all + disable variant is not going away too quick, apart from the docs from where I took the yaml, I couldn't find it being flagged as deprecated.

This is an effective solution for opting in during updates.

.golangci.yml
kevgo: I think that one creates too many false positives as well. Our tests are kind of repetitive, which is intentional. They are supposed to be simple, straightforward scripts.

Makefile
```diff
@@ -48,7 +48,7 @@ help:  # prints all make targets
 lint: lint-go lint-md   # lints all the source code
 
 lint-go:  # lints the Go files
-	golangci-lint run --enable-all -D dupl -D lll -D gochecknoglobals -D gochecknoinits -D goconst -D wsl -D gomnd -D dogsled src/... test/...
+	golangci-lint run src/... test/...
```
kevgo, line 51: So nice that we can run golangci-lint from the command line now!

blaggacao:
@kevgo Cool, this should go green now, so we can ship it.

.golangci.yml
```diff
@@ -1,11 +1,21 @@
+linters:
+  enable-all: true
+  disable:
+    - dogsled # The rationale for disabling dogsled is that all the cases it flags have been false positives in test code that intentionally ignores many return values because they test only one thing.
+    - dupl # I think that one creates too many false positives as well. Our tests are kind of repetitive, which is intentional. They are supposed to be simple, straightforward scripts.
+    # - exhaustive        TODO: enable after next release; current release at time of writing is v1.27
+    - gochecknoglobals # We are currently using some globals. It's a lot more forgivable in a CLI tool than in a library.
+    - gochecknoinits # I think the CLI framework we use requires them.
+    - goconst # Our tests contain a ton of hard-coded test strings, for example branch names.
+    - goerr113 # Expose errors properly, once/if approaching library use, as opposed to current cli-only use
```
kevgo, line 10: We don't want to disable `goerr113` completely. It has many useful error checks, for example making sure we compare errors with `errors.Is`. We only want to disable the `do not define dynamic errors` check because that one applies to libraries only. That's what I did in the `exclude-rules` section. Can you please restore that section and remove `goerr113` from here?

blaggacao:
Would you agree to ship this?

kevgo:
A beautiful change, thanks for improving the linter configuration!

PR #1504, opened 2020-06-03
Use int64 for pr number (#1478)

blaggacao:
Carved out from #1499


blaggacao:
I rebased `dar-gitea-support` atop of this which now is clean.

PR #1508, opened 2020-06-03
WIP: new code hosting driver architecture

kevgo:
Hey @charlierudolph and @blaggacao, here comes a preview of how I plan on simplifying the driver architecture. This is unfinished right now but I wanted to share to get early feedback.

PR #1514, opened 2020-06-03
Improve code hosting driver architecture

kevgo:
Here is the production implementation of the spike in #1508. I chose to throw that spike away and redo the refactor in order to do this in many small steps that keep the tests green along the way. They say if you aren't ashamed of the code you wrote 6 weeks ago, you aren't growing as a developer anymore. This code here is years old, and I am very ashamed of how over-engineered it was. All I can say is that it sounded like a good idea at the time, and it's good to see I'm apparently still growing as a developer after 30 years of coding.  😆 

@blaggacao sorry for exposing you to the poor driver architecture before as part of your Gitea work. I was hoping you just create a copy of the GitHub driver, but you (rightfully) started to hack around the poor testability of that part of the codebase. This new architecture has much less boilerplate and properly injects all dependencies, allowing simple and straightforward unit testing. I'm also excited that we finally start using caller-defined interfaces in this codebase. It took me a while to wrap my head around how they work. I'm finally getting the hang of it.


src/steps/abort_merge_branch_step.go
```diff
@@ -1,13 +1,16 @@
...
+)
 
 // AbortMergeBranchStep aborts the current merge conflict.
 type AbortMergeBranchStep struct {
 	NoOpStep
 }
 
 // Run executes this step.
-func (step *AbortMergeBranchStep) Run(repo *git.ProdRepo) error {
+func (step *AbortMergeBranchStep) Run(repo *git.ProdRepo, driver drivers.CodeHostingDriver) error {
```
kevgo, line 14: I'm not too big of a fan of providing the drivers to every step like this when they are used only in some, but it's the most straightforward way that is easiest to understand. If we inject more dependencies in the future, we can refactor this into a `context` struct.

kevgo:
Dammit, I'm shipping this. @blaggacao you have waited long enough, so let's go. Just make your Gitea branch against `master` and we'll review and ship it! 

@charlierudolph review comments are as always appreciated!

kevgo:
@blaggacao your review is of course also welcome, even though I shipped this PR to unblock you. You have contributed a lot of good ideas already, for example those that led to this PR!

src/drivers/bitbucket_test.go
```diff
@@ -0,0 +1,44 @@
...
+}
+func (mbc mockBitbucketConfig) GetCodeHostingOriginHostname() string {
+	return mbc.configuredHostName
+}
+
+func TestLoadBitbucket(t *testing.T) {
+	driver := drivers.LoadBitbucket(mockBitbucketConfig{
+		codeHostingDriverName: "bitbucket",
+		remoteOriginURL:       "git@self-hosted-bitbucket.com:git-town/git-town.git",
+	})
```
blaggacao, line 30: Yep that's definitely way cleaner. Thanks!

src/drivers/core.go
```diff
@@ -4,35 +4,62 @@ import "github.com/git-town/git-town/src/git"
...
-			activeDriver.SetAPIToken(activeDriver.GetAPIToken())
-		}
-	}
-	return activeDriver
+// CodeHostingDriver defines the structure of drivers
+// for the different code hosting services.
+type CodeHostingDriver interface {
+
+	// CanMergePullRequest returns whether or not MergePullRequest should be
+	// called when shipping. If true, also returns the default commit message
```
blaggacao, line 12: "... and the pull request number"

src/drivers/github.go
```diff
@@ -22,8 +30,32 @@ type githubCodeHostingDriver struct {
...
+	configuredHostName := config.GetCodeHostingOriginHostname()
+	if configuredHostName != "" {
+		hostname = configuredHostName
+	}
+	if driverType != "github" && hostname != "github.com" {
+		return nil
+	}
+	repositoryParts := strings.SplitN(helpers.GetURLRepositoryName(originURL), "/", 2)
+	if len(repositoryParts) != 2 {
+		return nil
```
blaggacao, line 48: Have you found a case where it get's that far?

src/drivers/gitlab.go
```diff
@@ -8,14 +8,36 @@ import (
...
+
 type gitlabCodeHostingDriver struct {
 	originURL  string
 	hostname   string
 	repository string
 }
 
-func (d *gitlabCodeHostingDriver) CanBeUsed(driverType string) bool {
-	return driverType == "gitlab" || d.hostname == "gitlab.com"
+// LoadGitlab provides a GitLab driver instance if the given repo configuration is for a Github repo,
```
blaggacao, line 23: Gitlab repo

src/steps/abort_merge_branch_step.go
blaggacao: Conceptually I like the idea to introduce the driver at the very start of every command. That way it's always around with you for present and _future_ use cases.

src/drivers/github.go
kevgo: We should never hit this **in theory**. Not checking this feels too much like recklessly ignoring a possible error condition. Better safe than sorry.

src/drivers/gitlab.go
kevgo: addressed in #1517

src/drivers/core.go
kevgo: addressed in #1517

PR #1509, opened 2020-06-03
Add Gitea

blaggacao:
based off of ongoing hosting architecture refactoring #1508

Issue #1451, opened 2020-05-21
Add gitea support


blaggacao:
gitea is a highly popular git server written in go, which looks very similar to github.

kevgo:
Agree! Since the maintainers don't have access to Gitea, this is something that should come from the community.

Guidance for implementation:
- follow the structure of the other code hosting drivers in https://github.com/git-town/git-town/tree/master/src/drivers

blaggacao:
note to future self (or other): https://gitea.com/gitea/go-sdk

blaggacao:
must fix #1164 on the go

blaggacao:
@kevgo What's the (guaranteed) expected shape of the string returned by `MergePullRequest(MergePullRequestOptions) (string, error)` ? It wouldn't be too bad to replace those string interfaces with well-defined structs...

kevgo:
Wow, cool that Gitea provides a Go SDK to talk to their API! This makes creating support for Gitea super easy. Just follow the structure of the [GitHub driver](https://github.com/git-town/git-town/blob/master/src/drivers/github.go), it uses a Go SDK for the Github API.

The `MergePullRequest` method returns the SHA of the commit created by the merge. See [the implementation](https://github.com/git-town/git-town/blob/master/src/drivers/github.go#L161) for details. It picks the `sha` field from the [response](https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button) provided by GitHub. [Here](https://github.com/git-town/git-town/blob/master/src/steps/driver_merge_pull_request_step.go#L71) is where the result is used.

You are right that returning naked strings feels a bit "stringly typed". Git Town uses structs to structure data where possible. An example is the input ([MergePullRequestOptions](https://github.com/git-town/git-town/blob/master/src/drivers/github.go#L57)) to the method mentioned above. The return value of this method, however, is a single string. Creating a struct to contain just a single string feels like overkill since all it does is add a name for the string variable. I'll address your concern by adding names for the return variable. Thanks for helping improve Git Town!

kevgo:
I have created #1456 to add names to return values.

blaggacao:
Yeah, indeed it's relatively straight forward, however, there is [one "hard" problem](https://github.com/git-town/git-town/blob/63a65fa148d4cf6ab783bb68e9533bd5936fca7a/src/drivers/gitea.go#L196), I hope this is bearable.

kevgo:
[...]

Issue #1383, opened 2020-05-07
Support rebase strategy for `sync` and `new-pull-request`


aeneasr:
GitHub now allows us to require a linear history which prevents merge commits from being pushed to matching branches:

<img width="784" alt="Bildschirmfoto 2020-05-07 um 11 51 15" src="https://user-images.githubusercontent.com/3372410/81280745-35f59f00-9059-11ea-8dbc-acc050205f03.png">

Because of the way `git sync` works, we end up with merge commits in between our commits if the main dev branch had updates. This in turn can cause the "rebase" merge option on GitHub to not be available. I believe this happens when merge conflicts have been resolved. 

This then conflicts with the linear commit history dictated by the branch protection rule, leaving only "Squash & Rebase" as a viable merge strategy which destroys our beautiful commit messages!

Therefore, I would like to be able to tell git-town to use `git pull --rebase` (or whatever it is doing under the hood) instead of `git pull` when doing `git-town sync`

More context:

- https://help.github.com/en/github/administering-a-repository/requiring-a-linear-commit-history
- https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges

cirego:
I too would also like to see an option for using `rebase` instead of `merge` to keep my branch history linear.

kevgo:
I'm not against doing this. Just out of curiosity, why don't you do do `Squash and merge`? That's the most idiot-proof option in my experience (for people like me) and what Git Town recommends in order to maintain a linear history on long-lived branches. It allows you to commit frequently on feature branches, take detours, move fast and break things, swear in commit messages, merge as often as you want, and when shipping the feature branch, all this gets collapsed into one beautiful linear commit that looks perfect and has a thoughtfully crafted commit message.

If I look at some of your PRs ([this](https://github.com/ory/hydra/pull/1446/commits) or [this](https://github.com/ory/hydra/pull/1570/commits) or [this](https://github.com/ory/hydra/pull/1846/commits)), do you really want all these commits in your `master` branch?

aeneasr:
For all examples there are counter examples (https://github.com/ory/kratos/pull/388, https://github.com/ory/kratos/pull/392) and rebase pull is a core (and loved) git strategy. Sure, there are other tools to help with the flow but why not support it natively and follow best practices documented on GitHub? I guess writing my own git alias for this is easy as pie but it would be much nicer if this came out of the box!

[...]

PR #1518, opened 2020-06-04
Add Gitea

blaggacao:
Replaces #1509 #1451 #1499 #1454 #1487


documentation/development/drivers.md
```diff
@@ -1,9 +1,20 @@
-# Code Hosting Drivers
+# Drivers
```
kevgo, line 1: I just shipped some updates to this file. Can you please insert Gitea into the state that is currently on master?

src/drivers/gitea.go
```diff
@@ -0,0 +1,195 @@
...
+	openPullRequests, err := d.client.ListRepoPullRequests(d.owner, d.repository, gitea.ListPullRequestsOptions{
+		ListOptions: gitea.ListOptions{
+			PageSize: 50,
+		},
+		State: gitea.StateOpen,
+	})
+	if err != nil {
+		return false, "", 0, err
+	}
+	fmt.Printf("1111 %#v\n", openPullRequests[0])
```
kevgo, line 74: Please remove this debug statement

src/drivers/gitea.go
```diff
@@ -0,0 +1,195 @@
...
+//   children2 -> ancestor  --> master (retargeted to master after merge)
+//nolint:unparam
+func (d *giteaCodeHostingDriver) apiRetargetPullRequests(pullRequests []*gitea.PullRequest, newBaseName string) error {
+	for _, pullRequest := range pullRequests {
+		// if options.LogRequests {
+		// 	helpers.PrintLog(fmt.Sprintf("Gitea API: Updating base branch for PR #%d to #%s", *pullRequest.Index, newBaseName))
+		// }
+		helpers.PrintLog(fmt.Sprintf("Gitea API: Updating base branch for PR #%d to #%s", 1, newBaseName))
+		helpers.PrintLog(fmt.Sprintf("The Gitea API currently does not support retargeting, please restarget #%d manually, see https://github.com/go-gitea/gitea/issues/11552", pullRequest.Index))
+		// _, err = d.client.EditPullRequest(d.owner, d.repository, *pullRequest.Index, &gitea.EditPullRequestOption{
```
kevgo, line 187: Can you remove this commented out code? If you want to keep it, please add a comment that explains why it is commented out and when it can be uncommented.

src/drivers/gitea_test.go
```diff
@@ -0,0 +1,226 @@
...
+	mge.repo = "git-town"
+	mge.version = mge.root + "/version"
+	mge.prBase = mge.root + "/repos/" + mge.orga + "/" + mge.repo + "/pulls"
+	mge.prOpen = mge.prBase + "?limit=50&page=0&state=open"
+	mge.pr1 = mge.prBase + "/1"
+	mge.pr1Merge = mge.pr1 + "/merge"
+	return mge
+}
+
+func (mgc mockGiteaConfig) GetCodeHostingDriverName() string {
```
kevgo, line 43: Move these method definitions to line 19 to keep them together with the definition of the struct?

src/drivers/gitea.go
```diff
@@ -0,0 +1,195 @@
...
+	baseName := parentBranch
+	headName := d.owner + "/" + branch
+	pullRequests := filterPullRequests(openPullRequests, baseName, headName)
+	if len(pullRequests) != 1 {
+		return false, "", 0, nil
+	}
+	pullRequest := pullRequests[0]
+	if !pullRequest.Mergable {
+		return false, "", 0, nil
+	}
```
blaggacao, line 84: @kevgo I'd like to tell the user that the server reports it indeed has found a commit, but this commit is not mergeable, "please open the following url to check: ...". However it looks like this would be silently ignored in the current implementation.

I'd like to suggest a `LoadPullRequest()` which should return a user exposed error, and otherwise load all state (and keep it on the driver) for doing the pull request. 

https://github.com/git-town/git-town/blob/8cd928ed8952bbfd897646006a98f851d72c1eda/src/cmd/ship.go#L68-L69

Could become:
```golang
		driver := drivers.Load(repo.Configuration)
		config, err := gitShipConfig(args, driver)
		ok, err := driver.InitMergePullRequest(config) 
		if err != nil {
                         // tell the user why not and eventually go on
                }
                config.canShipWithDriver = ok
		if err != nil {
```

src/drivers/gitea_test.go
```diff
@@ -0,0 +1,226 @@
...
+	prBase   string
+	prOpen   string
+	pr1      string
+	pr1Merge string
+}
+
+func newMockGiteaEndpoints() (mge mockGiteaEndpoints) {
+	mge = mockGiteaEndpoints{}
+	mge.root = "https://gitea.com/api/v1"
+	mge.orga = "git-town"
```
kevgo, line 33: A lot of these fields are used only once, and are constants. An example is `mge.orga`. Given that this is just simple test data, can you inline as many fields as possible here? The only fields used in tests seem `prOpen`, `pr1Merge`, and `version`.

src/drivers/github_test.go
```diff
@@ -25,6 +25,34 @@ type mockGithubConfig struct {
...
+	pr1         string
+	pr2         string
+	pr3         string
+	pr1Merge    string
+}
+
+func newMockGitHubEndpoints() (mge mockGitHubEndpoints) {
+	mge = mockGitHubEndpoints{}
+	mge.root = "https://api.github.com"
+	mge.orga = "git-town"
```
kevgo, line 44: Same here, most of these fields aren't used in tests.

src/drivers/gitea.go
blaggacao: :) Tests are ffialing, again...

src/drivers/gitea.go
kevgo: Helpful error messages to the user are always welcome! Let's build that as a separate feature. Maybe as a `canShip(config)` method that returns a string. If the string is empty, it means no problem. If the string contains something, it is considered an error message, printed to the user, and the program exits.

src/drivers/github_test.go
```diff
@@ -38,7 +31,29 @@ func (mgc mockGithubConfig) GetCodeHostingOriginHostname() string {
 	return mgc.configuredHostName
 }
 
-func setupDriver(t *testing.T, token string) (drivers.CodeHostingDriver, func()) {
+type mockGitHubEndpoints struct {
```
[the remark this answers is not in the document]
blaggacao: Ah, that sounds good. I needed to change because of risk of naming conflict, cause the weren't prefixed. Bit prefixed constants seem juts fine.

blaggacao:
@kevgo Check Kevin! Let's call this a fun and successful raid! :+1: Enjoyed it very much and learned a ton. Next on the fun project list: kakoune plugin, I guess. (and of course getting completion shipped) :rocket:

kevgo:
Same here! Pick whatever feels the tastiest to you right now! Btw I really like your idea of supporting draft feature branches (#1519). I could see a lot of people - including you! - using this regularly.

kevgo:
Or #1383. You keep talking about this feature a lot!  😃

blaggacao:
It sill has to mature in mi mind, though :wink:

PR #1526, opened 2020-06-04
Use current gopath bin for local testing

blaggacao:
This is why my tests where always c=failing locally.
Look's like the godog also suffers from this problem, bu couldn't find yet where to direct it to the gopath binary.

### DIFF
--- a/text-run/command-description.js
+++ b/text-run/command-description.js
@@ -54,7 +54,9 @@ function getMd(activity) {
 
 function getCliDesc(activity) {
   const command = getCommand(activity.file)
-  const output = child_process.execSync(`git-town help ${command}`).toString()
+  const output = child_process
+    .execSync(`$GOPATH/bin/git-town help ${command}`)
+    .toString()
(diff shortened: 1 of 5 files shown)


kevgo:
$GOPATH is no longer required since a couple of years. Go uses `~/go` if this environment variable isn't set, or `c:\go` on Windows. For example, I have a standard install of Go, and I don't have this environment variables set at all.

It sounds like you don't have `$GOPATH/bin` in your $PATH. If that's the case, many other commands like `godog`, `golangci-lint`, and `gox` won't work either. The better seems to add `$GOPATH/bin` to your path. Maybe we add documentation to do this to https://github.com/git-town/git-town/blob/master/documentation/development/setup.md?

blaggacao:
The easiest fix is `rm $(which git-town)`... I think it's ok as a developer to eat your own dogfood, so documenting this would be ok... Best though in the `make setup`, with a check.

kevgo:
The solution is to make sure `$GOPATH/bin` comes before `/usr/bin` in your $PATH. Then it uses the `git-town` binary that you locally compile if one exists.

PR #1533, opened 2020-06-06
Move script.Fetch to git.Runner

kevgo:
- eliminates all usages of `script.RunCommand` are eliminated in `src/steps`
- eliminates all usages of `scriptRunCommandSafe`


kevgo:
Going to change more of these files, so merging in. Please review!

PR #1539, opened 2020-06-06
Extract helpers from "util"

kevgo:
- moves all CLI input and output from the `util` package to the new `cli` package. This package is responsible for reading and writing to the CLI interface of Git Town
- inlines shallow helper methods from the `util` package. The code now calls the respective functions from the standard library directly. This allows for more precise error handling and more descriptive error messages. I have documented them with additional feature specs that cover edge cases like invalid values in the Git Town configuration.
- removes `validateBooleanArgument` as a pre-run check. We are now parsing boolean input values directly. This supports more ways to enter boolean inputs, like `t` or `1` for `true`. This is documented in additional feature specs as well.
- getting rid of `validateBooleanArgument` allowed to simplified the Cobra argument specifications to how it is documented
- there are only string helpers left in `util`
- updates documentation to describe the new package setup


kevgo:
Also, merges the `cfmt` package with the `cli` package.

kevgo:
Also, merges the `cfmt` package with the `cli` package.

PR #1561, opened 2020-06-14
Convert util package to stringslice

kevgo:
`util` is a discouraged name for packages since it doesn't say much and leads to "trash can" packages that can contain all sorts of things. Packages better have names like `stringslice` that indicate that they provide features for working with string slices. This also helps with naming of functions: rather than `util.DoesStringArrayContain`, we now have `stringslice.Contains`, and instead of `util.RemoveStringFromSlice` we now have `stringslice.Remove`.

### DIFF
--- a/src/cli/print.go
+++ b/src/cli/print.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/fatih/color"
-	"github.com/git-town/git-town/src/util"
 )
 
 // Printf prints the given text using fmt.Printf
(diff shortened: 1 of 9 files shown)


src/cli/print.go
```diff
@@ -52,6 +51,6 @@ func PrintError(messages ...interface{}) {
 func PrintLabelAndValue(label, value string) {
 	labelFmt := color.New(color.Bold).Add(color.Underline)
 	PrintlnColor(labelFmt, label+":")
-	Println(util.Indent(value))
+	Println(Indent(value))
```
kevgo, line 54: This command is so specific to printing output indented that it belongs to the `cli` package instead of a shared `util` folder.

PR #1624, opened 2020-07-28
Yarn workspaces

kevgo:
Installs the dependencies of all 3 Node codebase in one swoop and shares them, thereby preventing having three separate `node_modules` folders. See https://classic.yarnpkg.com/en/docs/workspaces for details.


kevgo:
On second thought, this is a stupid solution. Closing..

PR #1691, opened 2022-01-16
Golangci-lint 1.43

kevgo:
Big update with lots of new linters. I'll enable a few more in subsequent pull requests to keep this one manageable.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,6 +9,7 @@ linters:
     - gomnd # tests contain hard-coded test data that wouldn't make sense to extract into constants
     - ifshort
     - interfacer
+    - ireturn
     - lll # we aren't enforcing a line length at this point
     - maligned  # deprecated
     - nlreturn
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ linters:
     - goconst # tests contain a ton of hard-coded test strings, for example branch names
     - golint
     - gomnd # tests contain hard-coded test data that wouldn't make sense to extract into constants
+    - ifshort
     - interfacer
     - lll # we aren't enforcing a line length at this point
     - maligned  # deprecated
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,9 +2,14 @@ linters:
   enable-all: true
   disable:
     - cyclop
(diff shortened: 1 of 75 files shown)


.golangci.yml
```diff
@@ -1,11 +1,25 @@
...
+    - gofmt   # TODO: activate
+    - gofumpt   # TODO: activate
+    - goimports  # TODO: activate
+    - golint
     - gomnd # tests contain hard-coded test data that wouldn't make sense to extract into constants
+    - ifshort
+    - interfacer
+    - ireturn
     - lll # we aren't enforcing a line length at this point
+    - maligned  # deprecated
```
kevgo, line 19: golangci-lint prints warnings that these linters are deprecated, hence I disable them here. The goal is a clean linter output without any errors or warnings.

src/cmd/kill.go
```diff
@@ -53,7 +53,7 @@ Does not delete perennial branches nor the main branch.`,
 	},
 }
 
-// nolint: funlen
+//nolint:funlen
```
kevgo, line 56: Idiomatic Go is that machine-readable comments have no spaces.

src/cmd/root_test.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 func TestIsAcceptableGitVersion(t *testing.T) {
+	t.Parallel()
```
kevgo, line 12: Consistent parallelization of unit tests reduces unit test time from 3s to 2.3 sec on my machine, i.e. 24% reduction.

src/drivers/bitbucket.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/git-town/git-town/v7/src/drivers/helpers"
 )
 
-// bitbucketCodeHostingDriver provides access to the API of Bitbucket installations.
-type bitbucketCodeHostingDriver struct {
+// BitbucketCodeHostingDriver provides access to the API of Bitbucket installations.
+type BitbucketCodeHostingDriver struct {
```
kevgo, line 13: I am avoiding a few unnecessary interface returns, which isn't good Go style. This requires exporting this struct.

PR #1710, opened 2022-01-19
Make all asset filenames lowercase

kevgo:
Filenames should ideally not contain uppercase characters because that's just another way to cause confusion.


kevgo:
Verified that the installer still works on macOS.

PR #1775, opened 2022-01-26
Modernize shared features

kevgo:


features/alias/alias.feature
```diff
@@ -40,27 +40,3 @@ Feature: shorter Git Town commands
     Given my workspace is currently not a Git repo
     When I run "git-town alias true"
     Then it does not print "Not a git repository"
-
-  Scenario: no argument
```
kevgo, line 44: this is checked in shared already

features/git-town/edge_cases/invalid_git_version.feature
```diff
@@ -1,9 +0,0 @@
-Feature: minimum Git version
```
kevgo, line 1: checked in shared already

PR #1857, opened 2022-02-09
Simplify branch steps

kevgo:


features/append/append.feature
```diff
@@ -1,11 +1,11 @@
 Feature: append a new feature branch to an existing feature branch
 
   Background:
-    Given my repo has a feature branch "existing"
+    Given a feature branch "existing"
```
kevgo, line 4: The old version of this step was misleading. It doesn't create a branch in the user's local repo but also in origin.

features/append/append.feature
```diff
@@ -1,11 +1,11 @@
...
     And the commits
       | BRANCH   | LOCATION      | MESSAGE         |
       | existing | local, origin | existing commit |
-    And I am on the "existing" branch
+    And the current branch is "existing"
```
kevgo, line 8: Humans aren't on Git branches. Humans can run Git commands to check out Git branches, or make them the currently checked out branch.

features/diff-parent/current_branch/edge_cases/unknown_parent.feature
```diff
@@ -2,8 +2,7 @@
 Feature: ask for missing parent
 
   Scenario: on feature branch without parent
-    Given my repo has a branch "feature"
-    And I am on the "feature" branch
+    Given the current branch is "feature"
```
kevgo, line 5: Setting the current branch now also creates the branch if needed.

PR #1872, opened 2022-03-15
Fix docs for pull-branch-strategy parameters

brandonaut:
Seems the original docs were copy-pasted from the perennial branches command


kevgo (approved):
A clear example of falling victim to copy-paste... Thanks for fixing this! :+1:

PR #1875, opened 2022-05-14
Tests fail on non-English locales

manuelprinz:
Tests fail in non-English locales. A note was added to the development docs.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -17,6 +17,7 @@ This page helps you get started hacking on the Git Town codebase.
    [Node.js](https://nodejs.org), [Yarn](https://yarnpkg.com/),
    [scc](https://github.com/boyter/scc)
 5. run the tests: <code type="make/command">make test</code>
+   * Note: When running on a non-English locale, run with `LANG=C make test` to prevent tests from failing.
 6. compile the tool: <code type="make/command">make build</code>
 
 ## dependencies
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -17,6 +17,7 @@ This page helps you get started hacking on the Git Town codebase.
    [Node.js](https://nodejs.org), [Yarn](https://yarnpkg.com/),
    [scc](https://github.com/boyter/scc)
 5. run the tests: <code type="make/command">make test</code>
+   * Note: When running on a non-English locale, run with `LANG=C make test` to prevent tests from failing.
```
kevgo, line 20: This just adds an environment variable `LANG` and then runs `make test` normally. I assume that's needed for one of the commands that `make test` calls. Can pinpoint which command needs this, and then add the `LANG=C` prefix to the correct line inside the Makefile? This would allow the international developers to simply run `make test`, and Make will then set `LANG=C` for the test program that needs it.

kevgo (commented):
Thanks for this fix, it will help our international developers a lot!

kevgo:
Please sync this branch to fix the CI build.

DEVELOPMENT.md
manuelprinz: Seemed to be related to cuke. I reverted this commit and pushed a fix.

kevgo (approved):

PR #1879, opened 2022-05-15
Update golangci-lint to 1.46.1

kevgo:


kevgo:
Newer versions seem to contain changes that require Go 1.18. We'll stay on the current version until we update Go.

PR #1883, opened 2022-07-05
Update Regex for hostname extraction to support more ssh usernames

git-commit:
Add support for using other usernames than `git` for ssh urls.
For example, Bitbucket added support for using
`your-username@bitbucket.org` as a repository url quite a while back.

See for an explanation when this might be needed:
https://bitbucket.org/blog/better-support-multiple-ssh-keys

The regex change itself is rather basic, we could also use a more elaborate
regex to match the user name, such as: https://github.com/shinnn/github-username-regex#githubusernameregex


kevgo (approved):
Thanks for this improvement!

PR #1889, opened 2022-08-07
Fix go vet error reported using Go 1.18.

tjni:
Error: Println arg list ends with redundant newline

If the extra newline is preferred, let me know and I'll update this patch to keep it.


kevgo (approved):
Great catch, thanks for the submission!

PR #1890, opened 2022-08-26
Update Powershell autocompletion command

brandonaut:
`source` seems to have been copy-pasted from the bash examples. Outputting the completions to a string and running them with `Invoke-Expression` has worked for me in PowerShell.


kevgo (approved):
Thanks for fixing this. Yes, this is a copy-and-paste error.

PR #2145, opened 2023-03-09
docs: update `git town set-parent` -> `git town set-parent-branch`

arxanas:
Also, the `git town install` command link is broken, but I don't know what to update it to.


kevgo:
Thanks for fixing this and apologies for the broken documentation. The docs were describing the new behavior in v8.0. Now that v8.0 is out, the docs are (finally) correct. Hence, I'm closing this PR since it is no longer needed. However, thanks again for your contribution. :bow:

PR #2169, opened 2023-03-18
Improve the core architecture

kevgo:
A larger core architecture cleanup after a few years of changes and lots of learning about the domain and Go.

### reduces shared mutable state

The old [ProdRepo struct](https://github.com/git-town/git-town/blob/c10acf6f910848a70d1868b95a09c588feaa4415/src/git/prod_repo.go#L31-L51) contains two almost identical `Repo` instances. Both instances share mutable configuration data and in-memory caches that they both read and write to. The new architecture has a single [Config instance](https://github.com/git-town/git-town/blob/24e5c3cd68cdcac13b4b7eae9455c42993b374e2/src/git/prod_runner.go#L5) and leaner components that match the structure of the domain more closely (see below). 

### closer match of the code architecture to the domain

Git Town runs Git commands in three different ways: 
1. in the backend to (quietly) determine the state of the repository and which frontend commands to run
2. in the frontend to modify the repository while the end user watches
3. in tests while stubbing out various parts of the environment and simulating user input

(1) and (3) commands provide the output and error code of the command run while (2) commands only provide the error and stream the output to the CLI. In the old architecture, all three runners implement the same API and thereby pretend that they all return the output and the error code. Runner (2) "cheats" by always returning `nil` for the output. Callers have to be aware of this oddity and manually avoid nil pointer dereferences. 

Given that there is very little overlap between backend, frontend, and test Git commands in practice, the new architecture splits the Git commands up into three separate sets: 

1. [BackendCommands](https://github.com/git-town/git-town/blob/kg-runner-refactor/src/git/backend_commands.go) that a [BackendRunner](https://github.com/git-town/git-town/blob/24e5c3cd68cdcac13b4b7eae9455c42993b374e2/src/subshell/backend_runner.go) executes while capturing their output. If `dry-run` is set, they are run using a [BackendLoggingRunner](https://github.com/git-town/git-town/blob/24e5c3cd68cdcac13b4b7eae9455c42993b374e2/src/subshell/backend_logging_runner.go) instead, which also prints the commands it runs to the CLI.
2. [FrontendCommands](https://github.com/git-town/git-town/blob/kg-runner-refactor/src/git/frontend_commands.go) that stream their output to the CLI using a [FrontendRunner](https://github.com/git-town/git-town/blob/24e5c3cd68cdcac13b4b7eae9455c42993b374e2/src/subshell/frontend_runner.go). If `dry-run` is set, they run using a [FrontendDryRunner](https://github.com/git-town/git-town/blob/24e5c3cd68cdcac13b4b7eae9455c42993b374e2/src/subshell/frontend_dry_runner.go), which prints the commands but does not execute them. 
3. [TestCommands](https://github.com/git-town/git-town/blob/24e5c3cd68cdcac13b4b7eae9455c42993b374e2/test/test_commands.go) that execute in a mocked environment in a test directory and send fake user input into the subshell using a [MockingRunner](https://github.com/git-town/git-town/blob/24e5c3cd68cdcac13b4b7eae9455c42993b374e2/test/mocking_runner.go).

The new names reflect their responsibilities better. For example, the old `SilentRunner` wasn't silent if the `debug` option was enabled.

### reduce coupling

The old codebase contains a few Go anti-patterns like an [interface](https://github.com/git-town/git-town/blob/c10acf6f910848a70d1868b95a09c588feaa4415/src/subshell/runner.go)/impl pair ([impl 1](https://github.com/git-town/git-town/blob/c10acf6f910848a70d1868b95a09c588feaa4415/src/subshell/silent_runner.go), [impl 2](https://github.com/git-town/git-town/blob/c10acf6f910848a70d1868b95a09c588feaa4415/src/subshell/logging_runner.go)) in the `subshell` package. Many other packages (`browser`, `config`, `git`) depend on them, even though it's none of their business which package runs shell commands and how. In the new architecture, the `browser`, `config`, and `git` packages define their subshell needs via their own client-side interfaces. This reduces unnecessary dependencies between these packages and makes using stubs in tests easier.

### more clear and direct expression of intent

The old architecture initializes the runtime with the address of a variable containing the value of the `debug` CLI flag. This variable is initially set to `false` and code is not supposed to look at it until later when at some point it is "magically" set to the correct value by the Cobra runtime. 

The new architecture waits until it has the correct value for all CLI flags and then initializes the runtime based on the actual configuration. This enables the use of different implementations of components to provide different behavior based on the given configuration. 

This new architecture lists all the CLI flags that a Git Town command supports  in the command's source code file. It allows to define flags and read the Git repository specific for each command and only for commands that actually need this.

### misuse-proof APIs

The new architecture uses a pattern where the build fails if a command forgets to define a Cobra flag it uses, or forgets to use a Cobra flag it defines.


kevgo:
@charlierudolph good questions :slightly_smiling_face: I have added more details to the PR description. PTAL and please let me know if you have any questions or feedback! :pray: 

Sorry for the massive PR. I have extracted as many changes as I could from it already. What remains is somewhat coupled together.

PR #2187, opened 2023-03-29
Basic Azure DevOps support

### LINKED ISSUE #1657: Feature Request for Supporting Pull Requests on Azure DevOps
Would like to see Azure Repos supported for opening a new pull request. 

If you have contribution guidelines for adding a new provider or any pointers for someone new to the project I'd appreciate it. Perhaps I could contribute this in the near future then.

kevgo:
resolves #1657


features/repo/azure_devops.feature
```diff
@@ -0,0 +1,26 @@
+@skipWindows
+Feature: Azure DevOps
+
+  Scenario Outline:
+    Given the origin is "<ORIGIN>"
+    And tool "open" is installed
+    When I run "git-town repo"
+    Then "open" launches a new pull request with this url in my browser:
+      """
+      https://dev.azure.com/organization/repository
```
kevgo, line 10: This is supposed to be the web page for the repo on Azure DevOps. Similar to https://github.com/git-town/git-town for the Git Town codebase.

kevgo (commented):
Here is initial basic support for Azure DevOps. It supports creating new pull requests. I found different variations of URLs for Azure DevOps repos online. Would be good to get feedback from somebody using Azure DevOps how their respective URLs look like.

kevgo:
@sheldonhull @whitebear-gh  ^

features/repo/azure_devops.feature
sheldonhull: Sorry it's a bit more convoluted than this.

- Here's a good reference [lazygit](https://github.com/jesseduffield/lazygit/blob/b17c38befdd8ea6d4f71353391deb33fd6cc7176/pkg/commands/hosting_service/definitions.go#L45)

1. Source from and to can be pre-populated
2. SSH url is different from the https based url
3. Many still use the visualstudio.com domain even though the dev.azure.com is the more modern and probably the best one to use as it's future proof, if the url isn't be parsed via the underlying provider. If it is it will need to recognize both azure devops + visualstudio.com urls even if the end url is hard coded to the dev.azure.com.

sheldonhull:
taking a peek now.

sheldonhull (commented):
so quick feedback as I started on a few tweaks to push in a fork to show.... 

CleanShot 2023-04-25 at 19 54 32@2x


You'll see project needs to be provided as unlike GitHub, their is a sub grouping of Project level. 


In addition, the url scheme for repos if following dev.azure.com is `https://dev.azure.com/${ORG}/${PROJECT}/_git/${REPO}/` (using https). This can't be parsed the same way though if the visualstudio project format is used I think as the structure is different. SSH also changes the path format  unlike github's approach. 

That's a quick bit of feedback and hopefully I can add more in a fork to help such as updating the `.feature` files to reflect azure-devops reality.

Because of other restrictions such as linking work-items and more I'll probably stick with using my cli tool az-pr, but I do think this will make pr's faster for many I work with that don't want to do it fully via CLI.

PR #2217, opened 2023-04-12
read env variables for GitHub token

### LINKED ISSUE #2216: Read GitHub token from env variable
Instead of having the token in the git config it should first test whether `GITHUB_TOKEN` or `GITHUB_AUTH_TOKEN` env vars are set

Reasoning: many applications use one or the other env variable and using them eliminates the requirement to duplicate the configuration in multiple places

sascha-andres:
priority:
  GITHUB_TOKEN > GITHUB_AUTH_TOKEN > git config 

Closes #2216

### DIFF
--- a/src/hosting/github.go
+++ b/src/hosting/github.go
@@ -119,7 +119,7 @@ func NewGithubConnector(gitConfig gitTownConfig, log logFn) (*GitHubConnector, e
 // It first checks the GITHUB_TOKEN environment variable.
 // If that is not set, it checks the GITHUB_AUTH_TOKEN environment variable.
 // If that is not set, it checks the git config.
-func getGitHubApiToken(gitConfig gitTownConfig) string {
+func getGitHubAPIToken(gitConfig gitTownConfig) string {
 	apiToken := os.ExpandEnv("$GITHUB_TOKEN")
 	if apiToken == "" {
 		apiToken = os.ExpandEnv("$GITHUB_AUTH_TOKEN")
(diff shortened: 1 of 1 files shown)


kevgo:
@sascha-andres  Thanks for adding this super useful feature! :pray:  Can you please run `make fix` or (if that doesn't work on your platform) `make test-go` and make the linter happy? One issue that it complains about is:

> Error: src/hosting/github.go:122:6: var-naming: func getGitHubApiToken should be getGitHubAPIToken

sascha-andres:
@kevgo should be done by now

make fix actually went into some infinite loop - had to turn off my laptop by long pressing shutdown button. Just FYI

kevgo:
Okay, thanks! There are new build errors now:

```
Error: src/hosting/github.go:102:14: undeclared name: `getGitHubApiToken` (typecheck)
	apiToken := getGitHubApiToken(gitConfig)
	            ^
Error: src/runstate/jsonstep.go:37:15: invalid operation: cannot compare j.Step == nil (operator == not defined on untyped nil) (typecheck)
	if j.Step == nil {
	             ^
```

> make fix actually went into some infinite loop - had to turn off my laptop by long pressing shutdown button. Just FYI

Let's fix that. Can you please share your OS and configuration? The Make tasks work for me on Linux, macOS, and Windows. Did your machine lock up so hard that you couldn't kill the runaway process?

sascha-andres:
Error: src/runstate/jsonstep.go:37:15 is a bit weird as the change is quite local and should not touch the runstate

Pushed the changed call

System: Ubuntu 22.04, AMD64

It was a total lock, could not kill

kevgo:
> Error: src/runstate/jsonstep.go:37:15 is a bit weird as the change is quite local and should not touch the runstate

Agree, this seems unrelated, please ignore. Apparently golangci-lint reports such findings inconsistently.

> System: Ubuntu 22.04, AMD64
> 
> It was a total lock, could not kill

The Makefile downloads and executes a number of executables (linters, formatters). Does your system have protection against executing downloaded binaries? Some of these systems ask the user whether to allow running such untrusted binaries. If that dialog is modal, it would block your entire UI until it is dismissed. If you don't see the dialog, maybe it displays on another display?

PR #2249, opened 2023-05-20
Update GitLab merge request URL

manuelprinz:
GitLab versions 16.x and later removed the redirect that was in place for the URL that Git Town currently uses, which will break the functionality to open new pull/merge requests. The new URL has been in place since (at least) GitLab version 15.x, so this change is backwards-compatible. I was not able to access GitLab installations older than that.

GitLab 16.x is rolled out partially already on GitLab.com, and will be fully rolled out in the next days, which will break Git Town for all users of GitLab.com starting Tuesday.


kevgo:
Thanks for helping keep Git Town up to date! :pray:

manuelprinz:
Sure! Thanks for writing and maintaining Git Town! :pray:

PR #2458, opened 2023-09-15
Check for additional commits before resetting branches

kevgo:
Resetting branches as part of undoing activities throws away a particular commit. This PR introduces a check that the branch didn't receive any additional commits before throwing away the outdated commit.


kevgo:
This breaks a lot of undo features. Probably better to not merge it as-is.

PR #2478, opened 2023-09-22
Always run local git commands in English

### LINKED ISSUE #2469: Panic while parsing git output when git language is not English
### Steps to reproduce

```shell
$ LANG=fr_FR git status
Sur la branche improve-setup
Votre branche est en avance sur 'origin/improve-setup' de 4 commits.
  (utilisez "git push" pour publier vos commits locaux)

rien à valider, la copie de travail est propre

$ LANG=fr_FR git town sync
panic: cannot determine the sync status for Git remote "[origin/improve-setup: en avance de 4] Gemfile: declare usage on Darwin 22" and branch name "improve-setup"
```

### Cause

From what I understand of the code, `git-town` attempts to parse the branch status, but doesn't find any of the (English) keywords in the git output (because git runs in French).

### Notes

If this is a legitimate issue (and not a misconfiguration on my side), I'm willing to open a PR that ensures git-town invokes git using the English locale.

kemenaran:
Otherwise the code parsing the output of git commands fails, because git replies with localized versions of the messages.

Fix #2469

I tested locally on my French machine, and it works (git town no longer panics). I'm not sure how to write a relevant test for this though…

### DIFF
--- a/src/subshell/backend_runner.go
+++ b/src/subshell/backend_runner.go
@@ -45,6 +45,7 @@ func (r BackendRunner) execute(executable string, args ...string) ([]byte, error
 	if r.Dir != nil {
 		subProcess.Dir = *r.Dir
 	}
+	subProcess.Env = append(subProcess.Environ(), "LC_ALL=C")
 	outputBytes, err := subProcess.CombinedOutput()
 	if err != nil {
 		err = ErrorDetails(executable, args, err, outputBytes)


kevgo:
Thanks for this contribution! I'll add an end-to-end test for this separately.

kevgo:
FYI this has been released as [v10.0.0](https://github.com/git-town/git-town/releases/tag/v10.0.0). Apologies for the delay, this PR came in the middle of a large-scale product cleanup that involved a number of breaking changes that need to be shipped together but took a while to finalize and test.

PR #2655, opened 2023-11-10
Fix config command to view/set main-branch

vbekiaris:
A minor omission in `git town config` command


kevgo:
Thank you @vbekiaris 🙏

PR #2663, opened 2023-11-18
Update installation notes for Scoop

aliesbelik:
Since actual version of `git-town` is available from Scoop's `main` bucket, no additional steps with adding specific buckets are required.


kevgo:
Thanks for this update! Very cool that Scoop supports Git Town out of the box now 🚀

Issue #2079, opened 2023-02-22
new config option: "git-town.sync-before-ship"


kevgo:
### Motivation

In the typical usage scenario, it often doesn't make sense to sync right before shipping.

- I sync my branches regularly using `git sync` or `git sync --all`
- CI tests the branch to ship
- once a branch is green, it should be shipped
- syncing it again would require another CI run

In this scenario, it doesn't make sense to sync all branches right before shipping. Let's therefore introduce a new config option `sync-before-ship` and default it to `false`.

kevgo:
This issue tracker tracks all the possible ideas for Git Town, for both the near and far future. You can see the tickets that will go into the next release in the upcoming [milestone](https://github.com/git-town/git-town/milestone/6).

PR #2714, opened 2023-12-01
Add New Config Option `sync-before-ship`

harrismcc:
Adds a new configuration option `sync-before-ship`, which will enable/disable syncing when the ship command is run #2079

### DIFF
--- a/src/cmd/config.go
+++ b/src/cmd/config.go
@@ -76,6 +76,7 @@ func determineConfigConfig(run *git.ProdRunner) (ConfigConfig, error) {
 	pushNewBranches := fc.Bool(run.Config.ShouldNewBranchPush())
 	shouldSyncUpstream := fc.Bool(run.Config.ShouldSyncUpstream())
 	syncFeatureStrategy := fc.SyncFeatureStrategy(run.Config.SyncFeatureStrategy())
+	syncBeforeShip := fc.Bool(run.Config.SyncBeforeShip())
 	return ConfigConfig{
 		branchTypes:           branchTypes,
 		deleteOrigin:          deleteOrigin,
@@ -90,6 +91,7 @@ func determineConfigConfig(run *git.ProdRunner) (ConfigConfig, error) {
 		pushNewBranches:       pushNewBranches,
 		shouldSyncUpstream:    shouldSyncUpstream,
 		syncFeatureStrategy:   syncFeatureStrategy,
+		syncBeforeShip:        syncBeforeShip,
 	}, fc.Err
 }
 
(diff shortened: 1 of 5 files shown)


harrismcc:
This is a first-attempt at this so if it's not the right way I can fix it up.

Also, I wasn't able to run the test suite on my M1 mac (`ERROR: This application does not seem to support your platform.`) so I'm not sure if this will mess with any tests

kevgo:
Thanks for implementing this, the code is very well written! 🙏 Sorry about the broken tests, I was tinkering on some automation around that today. It should work after you sync your branch with upstream.

One last action item are the end-to-end tests. Your (good) changes the output of the `git town config` command break a few end-to-end tests that document/verify the output of that command. Please run `make cuke` and fix the broken tests. The test files are in the `features` folder, the test runner gives you their specific path. 

While you are at it, you could also add an end-to-end test for your new feature. Add a file `sync-before-ship.feature` to https://github.com/git-town/git-town/tree/main/features/ship/current_branch/features. You can take the content of https://github.com/git-town/git-town/blob/main/features/ship/current_branch/features/ship_delete_remote_branch.feature as a template for your E2E test. Just change the name of the `ship-delete-remote-branch` config variable to `sync-before-ship`. To run just your E2E test, add `@this` above the `Scenario:` or `Feature:` line and run `make cuke-this`.

Please don't hesitate to hit me up with any questions around this, no matter how big or small 😃

harrismcc:
@kevgo I've fixed the broken tests and added a new one. Let me know if that looks good or if there are other changes.

I had a bit of trouble with `make cuke` locally though. When I ran it, a seemingly random test would fail. When I did `make cukethis` it would just hang and never pass or fail. I tried on my mac and an ubuntu machine with the same results. Luckily I was able to test this out though by using Github CI on my fork.

kevgo:
Excellent additions, I have merged this in. Thanks again @harrismcc! 🙏 

The end-to-end test failures on your machine surprise me since they run reliably for me on Fedora, Ubuntu (on GitHub CI), Debian, Asahi, and macOS for both Intel and M1. I haven't experienced any flakiness for a while. Do you get the flaky end-to-end tests on the `main` branch as well?

Here is an example for how to place `@this` to be used together with `make cukethis`: https://github.com/git-town/git-town/blob/kg-demo-cuke/features/ship/current_branch/features/sync_before_ship.feature#L11

If you keep running into these flakiness issues, please send me the output of the failing tests and I'll investigate.

kevgo:
FYI this has been released as Git Town 11 🚀

kevgo:
This feature you built is so useful that in Git Town 11 it is active (set to false) by default. `git ship` no longer syncs the branch anymore unless the user explicitly activates the setting for it.

Issue #3233, opened 2024-03-26
The installer currently does not work :/


alexus37:
The current installer by running:

```bash
curl https://git-town.com/install.sh | sh
```

fails.

```
GIT TOWN INSTALLER SCRIPT

This installer is under development. Please report bugs at
https://github.com/git-town/git-town/issues/new.

No other Git Town installation found, proceeding to install Git Town 12.1.0 for linux_intel_64.

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     9  100     9    0     0    121      0 --:--:-- --:--:-- --:--:--   123

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
mv: cannot stat '.git-town-download/git-town': No such file or directory

I have successfully:
- downloaded https://github.com/git-town/git-town/releases/download/v12.1.0/git-town_12.1.0_linux_intel_64.tar.gz
- extracted the Git Town 12.1.0 binary to /home/vscode/.local/bin/git-town

Please perform these steps to finalize the installation:
- add /home/vscode/.local/bin to the PATH
```

PR #3234, opened 2024-03-26
fix wrong install path in install.sh

alexus37:
- fixes https://github.com/git-town/git-town/issues/3233
The download path seems wrong this pr fixes it.

However, after running the script. I am getting this error:

```bash
> git town config setup
git: 'town' is not a git command. See 'git --help'.

The most similar command is
        notes
```

### DIFF
--- a/website/src/install.sh
+++ b/website/src/install.sh
@@ -83,7 +83,7 @@ download_url() {
 		CPU=intel_64
 		EXT=zip
 	fi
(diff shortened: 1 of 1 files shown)


kevgo:
Yes, as of v11 Git Town binaries no longer contain the version. This change was made to make installing easier - but this simplification broke our own installer 🤣

Long story short, thanks for this fix! 🙏

PR #3285, opened 2024-04-02
Add alternative in wiki for UI based ships

nishchay-manwani-glean:
Adding a small line of context inside the https://www.git-town.com/stacked-changes#shipping-the-refactor wiki, as it caused  some confusion for some people in our repository.

I checked out the [development wiki](https://github.com/git-town/git-town/blob/main/DEVELOPMENT.md) and tested the changes on the local development server (see image attached below) and ran `make fix`. Please let me know if any other steps are required.
Screenshot 2024-04-02 at 7 56 17 PM

Thanks for maintaining such a great tool 🙏

### DIFF
--- a/website/src/stacked-changes.md
+++ b/website/src/stacked-changes.md
@@ -151,6 +151,10 @@ main
     3-rename-bar
 ```
 
+If you ship feature branches via the code hosting API or web UI, run
+`git sync --all` or `git sync` on the youngest child branch to update the
+lineage.
+
 ## Synchronizing our work with the rest of the world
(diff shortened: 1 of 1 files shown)


kevgo:
Thanks for this update! 🙏

kevgo:
Btw you should be able to run `git sync --all` on ANY branch and it should update all local branches. If it doesn't do that, and you have to go to the youngest branch, that would be a bug in Git Town.

kevgo:
The website is updated!

PR #3300, opened 2024-04-05
Fix zsh completion command

utkinn:
Zsh's `source` command does not work with pipes. Instead, one has to use process substitution to supply the autocompletion script. See [this StackOverflow question](https://unix.stackexchange.com/questions/176873/feed-source-command-with-a-pipe).


kevgo:
Thank you!! 🙏

PR #3340, opened 2024-04-15
Allow merge flag to be passed to git checkout

### LINKED ISSUE #3321: git town switch: -m flag
`git town switch` runs `git checkout <branch>`. This fails when there are uncommitted changes that conflict with the new branch. Running [git checkout -m <branch>](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---merge) performs a 3-way merge while switching, thereby allowing to switch branches in more situations.

Add a similar CLI option to `git town switch`. Running `git town switch -m` runs `git checkout -m <branch>`.

sascha-andres:
resolves #3321

### DIFF
--- a/src/cmd/switch.go
+++ b/src/cmd/switch.go
@@ -18,21 +18,23 @@ const switchDesc = "Displays the local branches visually and allows switching be
 
 func switchCmd() *cobra.Command {
 	addVerboseFlag, readVerboseFlag := flags.Verbose()
+	addMergeFlag, readMergeFlag := flags.SwitchMerge()
 	cmd := cobra.Command{
 		Use:     "switch",
 		GroupID: "basic",
-		Args:    cobra.NoArgs,
+		Args:    cobra.ArbitraryArgs,
 		Short:   switchDesc,
 		Long:    cmdhelpers.Long(switchDesc),
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			return executeSwitch(readVerboseFlag(cmd))
+			return executeSwitch(readVerboseFlag(cmd), readMergeFlag(cmd))
 		},
 	}
+	addMergeFlag(&cmd)
 	addVerboseFlag(&cmd)
 	return &cmd
 }
 
-func executeSwitch(verbose bool) error {
+func executeSwitch(verbose, merge bool) error {
 	repo, err := execute.OpenRepo(execute.OpenRepoArgs{
 		DryRun:           false,
 		OmitBranchNames:  false,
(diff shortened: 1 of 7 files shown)


src/cmd/switch.go
```diff
@@ -14,25 +15,42 @@ import (
...
+		if err != nil {
+			panic(fmt.Sprintf("command %q does not have a bool %q flag", cmd.Name(), mergeMessageLong))
+		}
+		return value
+	}
 	cmd := cobra.Command{
 		Use:     "switch",
 		GroupID: "basic",
-		Args:    cobra.NoArgs,
+		Args:    cobra.MaximumNArgs(2),
```
kevgo, line 41: I think this setting is for positional arguments only. Can you please try reverting to `NoArgs`? The flag you added should still work.

Background: [git town compress](https://github.com/git-town/git-town/blob/main/src/cmd/compress.go#L45) uses plenty of CLI flags as well and also has `NoArgs`.

src/git/frontend_commands.go
```diff
@@ -37,12 +37,14 @@ func (self *FrontendCommands) AbortRebase() error {
 
 // CheckoutBranch checks out the Git branch with the given name in this repo.
 func (self *FrontendCommands) CheckoutBranch(name gitdomain.LocalBranchName) error {
```
kevgo, line 39: In my (humble) opinion, you could just add the `merge bool` argument to the existing `CheckoutBranch` method and not create the two additional methods here. Other public methods in this file also have such optional arguments, for example https://github.com/git-town/git-town/blob/main/src/git/frontend_commands.go#L231.

kevgo (approved):
Thanks so much for this @sascha-andres! 🙏 

This PR implements https://github.com/git-town/git-town/issues/3321, so you could add `resolves #3321` in the body of the PR to automatically close that issue when we ship this PR.

The implementation is great! Below are a few suggestions to move some things to where we store related info in this codebase. If you don't have the bandwidth to implement the comments, just let me know and I'll merge this in and fix it up myself.

I'm excited to use the -m flag for `git town switch`! 🚀

src/cmd/switch.go
sascha-andres: NoArgs is defined like this:

`// NoArgs returns an error if any args are included.
func NoArgs(cmd *Command, args []string) error {
	if len(args) > 0 {
		return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
	}
	return nil
}
`

I believe ArbitraryArgs is what is required here

src/git/frontend_commands.go
sascha-andres: done

src/cmd/switch.go
```diff
@@ -14,25 +15,42 @@ import (
...
+	addMergeFlag := func(cmd *cobra.Command) {
+		cmd.PersistentFlags().BoolP(mergeMessageLong, mergeMessageShort, false, mergeDesc)
+	}
+	readMergeFlag := func(cmd *cobra.Command) bool {
+		value, err := cmd.Flags().GetBool("merge")
+		if err != nil {
+			panic(fmt.Sprintf("command %q does not have a bool %q flag", cmd.Name(), mergeMessageLong))
+		}
+		return value
+	}
```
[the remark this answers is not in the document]
sascha-andres: done

kevgo:
Thanks again, very well done! 🙏

PR #3354, opened 2024-04-16
Fix year in LICENSE

koppor:
The copyright year in the installer felt wrong. Thus, I updated it.

image


kevgo:
Thanks @koppor! 🙏

PR #3452, opened 2024-04-26
Remove unnecessary optionality from giturl.Parts

kevgo:


kevgo:
The Windows error is:

```
cannot parse Git URL "C:\\Users\\kevin\\AppData\\Local\\Temp\\1451745797\\undo_1\\origin"
```

PR #3636, opened 2024-06-12
Update command to source completions in Bash

stephenwade:
When I run the original command in my shell, I get the following error:

```console
$ git-town completions bash | source
bash: source: filename argument required
source: usage: source filename [arguments]
```


kevgo:
Thanks! 🙏

PR #3644, opened 2024-06-14
ship: remove sync-before-ship

kevgo:


src/cmd/ship.go
```diff
@@ -37,13 +37,7 @@ const shipDesc = "Deliver a completed feature branch"
...
-- syncs the main branch
-- pulls updates for <branch_name>
-- merges the main branch into <branch_name>
-- squash-merges <branch_name> into the main branch
-  with commit message specified by the user
-- pushes the main branch to the origin repository
-- deletes <branch_name> from the local and origin repositories
+Ships only direct children of the main branch. To ship a child branch, ship or kill all ancestor branches first.
 
 Ships direct children of the main branch. To ship a child branch, ship or kill all ancestor branches first.
```
stephenwade, line 42: **quibble:** These two lines are almost identical.

kevgo: Thanks, good catch! I have removed the duplicate line.

Issue #3511, opened 2024-05-05
[question/feature] Flag to disable feature branch push for `git town sync`


defunctzombie:
The documentation and help for `git town sync` say the following happens when running on feature branches:

```
When run on a feature branch:
- syncs all ancestor branches
- pulls updates for the current branch
- merges the parent branch into the current branch
- pushes the current branch
```

I would like a way to disable the last step. Here's my workflow:

I come back to a feature branch I had previously started, maybe a work-in-progress. I know I will be making changes to the code. I do a checkout to get to that feature branch and run `git town sync` so I can have the latest changes from upstream branches. The current behavior will correctly sync everything and then _push_ my feature branch. This is unfortunate for me since it can kick off CI on any PRs I have open with this branch when I am about to make more changes and don't have a need to run CI since I will push those changes soon.

This is a minor thing but having a flag for `git town sync` that could avoid pushing the current branch since I am about to work on it could be nice. I definitely put this in the niche nice-to-have category and don't consider it wrong that the default is to push.

kevgo:
Git Town has just the thing for you! Run [git town observe](https://www.git-town.com/commands/observe) on the branch. Now when you run `git sync`, it won't push your local changes to the tracking branch. See the [end-to-end test](https://github.com/git-town/git-town/blob/main/features/sync/current_branch/observed_branch/tracking_branch/with/happy_path.feature) for what exactly happens.

To make this observed branch a "normal" feature branch again (that gets pushed), run `git hack` on it.

defunctzombie:
At first glance I feel like I would accidentally `observe` a branch and then forget to un-observe it. In the case I describe, I have done a `git checkout` to start work on the existing branch and want to bring it up-to-date locally and then will push the changes once I'm done working. For this flow I am in the mindset of sync and work; observe takes me off that path.

If this doesn't align with how you imagine folks use sync or prefer to push the observe workflow feel free to close this. If you would entertain a flag for `sync` I am happy to try my hand at a PR proposing the change.

kevgo:
Thanks for the additional details. Yes, `git town observe` is intended as a long-term setting, and there is a risk of forgetting to un-observe a branch that isn't really an observed branch. Sounds like what you need is a `--no-push` option for `git sync` to skip pushing occasionally.

This is definitely a useful setting, and I would accept a PR that introduces this feature. Some guidance how to implement it:

[...]

PR #3648, opened 2024-06-16
Add --no-push flag to `sync` command

defunctzombie:
This change proposes a new boolean flag `--no-push` for the `sync` command.

When this flag is enabled, the final `git push` of the current feature branch is omitted. This allows a developer to run the `sync` command to fetch all upstream changes before starting to work on the feature branch without incurring the "cost" of pushing the feature branch about to be worked on.

See https://github.com/git-town/git-town/issues/3511 for more context


src/cmd/sync.go
```diff
@@ -50,23 +50,26 @@ func syncCmd() *cobra.Command {
 	addVerboseFlag, readVerboseFlag := flags.Verbose()
 	addDryRunFlag, readDryRunFlag := flags.DryRun()
 	addAllFlag, readAllFlag := flags.Bool("all", "a", "Sync all local branches", flags.FlagTypeNonPersistent)
+	addNoPushFlag, readNoPushFlag := flags.Bool("no-push", "", "Do not push branches", flags.FlagTypeNonPersistent)
```
defunctzombie, line 53: It wasn't obvious to me from your comment (https://github.com/git-town/git-town/issues/3511#issuecomment-2101035875) whether you wanted `no-push` to be in the `flags` directory for re-use like the `DryRun` flag or if I should add it only to this command. For a smaller diff review, I added it only to this comment.

src/cmd/sync.go
```diff
@@ -99,7 +102,7 @@ func executeSync(all, dryRun, verbose bool) error {
 			InitialBranch: data.initialBranch,
 			Remotes:       data.remotes,
 			Program:       &runProgram,
-			PushBranch:    true,
+			PushBranch:    !noPush,
```
defunctzombie, line 105: I believe this prevent _all_ branches in the sequence from being pushed? I'm thinking the behavior I want is that only the current feature branch is not pushed but could discuss the tradeoffs.

stephenwade: I would certainly be surprised if I wrote `--no-push` and it pushed anything. I think preventing all branches from being pushed is the right move.

kevgo: I second that. If the stated goal is reducing CI cost/pressure, then it should not push any branches.

defunctzombie: That's a reasonable stance. Certainly fair to start with this behavior and see if it feels right over time.

features/sync/current_branch/feature_branch/merge_sync_strategy/no_push/enabled.feature
```diff
@@ -0,0 +1,45 @@
...
+
+  Background:
+    Given the current branch is a feature branch "feature"
+    And the commits
+      | BRANCH  | LOCATION | MESSAGE               |
+      | main    | local    | local main commit     |
+      |         | origin   | origin main commit    |
+      | feature | local    | local feature commit  |
+      |         | origin   | origin feature commit |
+    And Git Town setting "no-push" is "true"
```
kevgo, line 11: No need for adding a separate Git Town config setting since there is already the command-line switch.

```suggestion
```

features/sync/current_branch/feature_branch/merge_sync_strategy/no_push/enabled.feature
```diff
@@ -0,0 +1,45 @@
...
+    Then it runs the commands
+      | BRANCH  | COMMAND                                 |
+      | feature | git fetch --prune --tags                |
+      |         | git checkout main                       |
+      | main    | git rebase origin/main                  |
+      |         | git push                                |
+      |         | git checkout feature                    |
+      | feature | git merge --no-edit --ff origin/feature |
+      |         | git merge --no-edit --ff main           |
+    And all branches are now synchronized
```
kevgo, line 24: I expect this line to fail since the `--no-push` flag will leave the tracking branches unsynchronized. I suggest replacing this line with the one suggested below. It contains an empty placeholder table. When you runn the end-to-end test (see above how to do that), the failing end-to-end test will print the missing content for the table.

```suggestion
    And these commits exist now
      | BRANCH   | LOCATION      | MESSAGE         |
```

features/sync/current_branch/feature_branch/merge_sync_strategy/no_push/enabled.feature
```diff
@@ -0,0 +1,45 @@
...
+    And feature branch is not pushed
```
kevgo, line 25: The fact that the feature branch is not pushed will be apparent in the list of commits in the table on the line right above: the tracking branch will not contain the local commits.

```suggestion
```

features/sync/current_branch/feature_branch/merge_sync_strategy/no_push/enabled.feature
```diff
@@ -0,0 +1,45 @@
...
+      |         |               | local main commit                                          |
+      |         |               | Merge branch 'main' into feature                           |
+
+  Scenario: undo
+    When I run "git-town undo"
+    Then it prints:
+      """
+      nothing to undo
+      """
+    And it runs no commands
```
kevgo, line 45: Starting point for the end-to-end test for undoing this sync. As usual, it's easiest to run it, copy-and-paste the missing table content, and verify that it is indeed what you want here.

```suggestion
    Then it runs the commands
      | BRANCH   | COMMAND               |
    And the current branch is still "feature"
    And the initial commits exist
    And the initial lineage exists
```

kevgo (commented):
Very cool, thanks for implementing this 🙏

If you get stuck on something, don't hesitate to tag me. I'm happy to help!

stephenwade:
@defunctzombie Have you made any progress on the end-to-end tests? I'm really looking forward to this feature.

stephenwade:
I'd like to help with the E2E tests, but should this wait until #3699 is done? @kevgo

kevgo:
I'm merging this in and polishing it up. Thanks for this contribution @defunctzombie! 🙏

kevgo:
@stephenwade thanks for helping with this. In general, just fire away. The end-to-end tests can easily be adjusted to the new format if needed.

defunctzombie:
Thanks for getting this merged! I've been busy with work and then out of vacation. Glad to return to see it merged :D

PR #3726, opened 2024-07-18
Make git.Commit use better default data

kevgo:


kevgo:
This ended up more complicated than necessary, closing in favor of #3727

PR #3887, opened 2024-08-22
Always retry failed opcodes unless the opcode says otherwise

kevgo:
Currently opcodes need to indicate whether they want to be retried when they fail. This only works assuming that the opcodes who don't indicate this cannot fail. In reality, all opcodes can fail, for example when a concurrent Git process suddenly starts running.

This PR changes the interpreter engine at the heart of Git Town to always retry failed opcodes.


kevgo:

PR #3889, opened 2024-08-22
Document loading compdef command on zsh

kevgo:


kevgo:

PR #3892, opened 2024-08-22
Print all API requests on the CLI

kevgo:
Git Town already logs some of the API requests in its output. This PR makes it log all API requests to hosting platforms.


kevgo:

PR #3930, opened 2024-09-04
Add ship-strategy.md to SUMMARY.md

enigma:
This will generate the page that currently 404s: https://www.git-town.com/preferences/ship-strategy.html
And fix the links are coming from here: https://www.git-town.com/commands/ship


kevgo:
Nice work on this fix! 🙌 This will be included in the v16.1 release.

PR #3953, opened 2024-09-07
Fix wrong branch name in stacked changes example

rbardini:
Replace `1-fix-typos` with `1-refactor` to match previous commands.


kevgo:
Thank you! 🙏  It's amazing how such mistakes can slip through even after proof-reading this so many times! 😆

Issue #3933, opened 2024-09-04
limit the branches that git town switch displays


kevgo:
`git town switch` is Git Town's command to switch between branches. By default, it switches between all branches. Options can trim down the list of branches to switch.

### display only branches of particular types

This is useful for example when cleaning up observed branches. While doing so I want to see only observed branches so that I can switch to one of them and sync/delete/change it.

For the syntax there are two options.

Option 1: separate flag for each branch type

```zsh
# switch to a new feature branch
git town switch --feature

# switch to a new feature, prototype, or parked branch
git town switch --feature --prototype --parked
```

This has the advantage that we can use shorter flags:

```zsh
git town switch -f -p -P

git town switch -fpP
```

But "prototype" and "parked" start with the same letter. 🙁 

Option 2: `--type` flag

```zsh
# switch to a new feature branch
git town switch --type=feature

# switch to a new feature, prototype, or parked branch
git town switch --type=feature,prototype,parked
```

This could also be shortened:

```zsh
git town switch --type=f,pr,pa
```

One can use several possible separator characters:

```zsh
git town switch --type=f,pr,pa
git town switch --type=f+pr+pa
git town switch --type="f&pr&pa"
git town switch --type="f|pr|pa"
```

### display only the given branches

give as arguments

```zsh
# switch only between branch-1, branch-2, and branch-3
git town switch branch-1 branch-2 branch-3

# switch to any of my branches, even if they are remote
git branch -a | grep 'remotes/origin' | sed 's/  remotes\/origin\///' | grep -v 'HEAD' | grep '^kg-' | xargs git town switch
```

Does it make sense to treat all given branch names as regexes?

```zsh
# display only branches whose name starts with "kg-" 
git town switch ^kg-

# display only branches whose name starts with "kg-" or with "foo/" 
git town switch ^kg- ^foo\/

# display the main branch and branches whose name starts with "kg-"
git town switch main ^kg-
```

A risk with this is displaying branches like `fix-remaining-issues` because that branch name contains "main". 

Branch names can also be read from STDIN. In this case the branch names are treated as literal branch names, i.e. never as regexes:

```zsh
# switch to any of my branches, even if they are remote
git branch -a | grep 'remotes/origin' | sed 's/  remotes\/origin\///' | grep -v 'HEAD' | grep '^kg-' | git town switch
```

kevgo:
Combining arguments: switch to a new prototype branch that contains "api":

```zsh
git town switch --prototype api

[...]

PR #3980, opened 2024-09-10
switch: filter branches by regex

kevgo:
part of #3933


kevgo:
This is now available as part of [Git Town 16.2](https://github.com/git-town/git-town/releases/tag/v16.2.0). Enjoy!

PR #3983, opened 2024-09-10
Fix spelling error "pererennial"

stephenwade:
This PR fixes a spelling error on the website where "perennial-regex" is spelled "pererennial-regex".


kevgo:
I made the same mistake in two places! 🤣

PR #4020, opened 2024-09-13
Print proposal numbers in API logs

kevgo:


kevgo:
This is now available in [v16.3](https://github.com/git-town/git-town/releases/tag/v16.3.0).

PR #4026, opened 2024-09-13
API log: Display whether a PR was found or not

kevgo:


kevgo:
This is now available in [v16.3](https://github.com/git-town/git-town/releases/tag/v16.3.0).

Issue #4021, opened 2024-09-13
Error: cannot make the main branch a feature branch


ruudk:
### Steps to reproduce

On a new project, I tried to run `git town hack` and it asked me:

![Screenshot 2024-09-13 at 13 29 31@2x](https://github.com/user-attachments/assets/779db19d-be45-4cc3-84f9-509a1ac6c33c)

I pressed enter, and was prompted with:
![Screenshot 2024-09-13 at 13 30 27@2x](https://github.com/user-attachments/assets/e2f47f5c-8852-42bd-a78b-76a14cdc9e44)

I feel this might be related to https://github.com/git-town/git-town/issues/3683 as I configured it as:
```
Branches:
  main branch: main
  perennial branches: (none)
  perennial regex: (not set)
  default branch type: observed
  feature regex: ^ruudk\/.*
  parked branches: (none)
  contribution branches: (no
  observed branches: (none)
```

### Command output

```shell
.
```

ruudk:
I removed my global feature regex:
```
git config unset --global git-town.feature-regex
```

It did not solve the problem.

kevgo:
Which branch did you have checked out when this happened? Given that it was a new repo, I assume it was `main`?

According to your screenshot you ran `git town hack` (without a branch name).
When called without arguments, `git town hack` converts the current branch into a feature branch ([docs](https://www.git-town.com/commands/hack#positional-arguments)).
Git Town therefore (correctly) tells you that you cannot make the main branch a feature branch.
So I think this is working as intended. Am I missing something?

ruudk:
Ah, normally I use git hack, an alias that prompts me for a branch name and shows a prefix. I think I was so used to that, that I didn't even notice it. So I think I just didn't use it correctly. But maybe the error can better explain that I was being silly 😁

kevgo:
> But maybe the error can better explain that I was being silly 😁

Do you have suggestions for what Git Town could have said in this situation that would have helped you understand better what is going on? I think the error message is pretty clear, but I also agree its a bit concise.

ruudk:
> When called without arguments, `git town hack` converts the current branch into a feature branch ([docs](https://www.git-town.com/commands/hack#positional-arguments)).

Maybe if the branch is the default branch, and you run `git town hack` without arguments, it could say: "You're trying to convert the main branch to a feature branch. That's not possible. If you want to create a feature branch, did you forget to add the branch name?"

kevgo:
Love it, thanks for the suggestion!

PR #4033, opened 2024-09-14
hack: improve error message

kevgo:
addresses https://github.com/git-town/git-town/issues/4021#issuecomment-2350881433


ruudk:
🥳

PR #4052, opened 2024-09-23
Consistently use un-aliased forms in documentation

kevgo:
This makes the documentation easier to understand for people new to Git Town, and separates the Git Town commands from Git commands.

PR #4055, opened 2024-09-24
Consistently use un-aliased forms in documentation

stephenwade:
This makes the documentation easier to understand for people new to Git Town, and separates the Git Town commands from Git commands.

This PR fixes one example missed in #4052.


kevgo:
Thanks!! 🙏

Issue #4038, opened 2024-09-16
rename "git town rename-branch" to "git town rename"


kevgo:

kevgo:
On the other hand, pretty much all Git Town commands are about branches. So why group this one only?

stephenwade:
It would be reasonable to rename "git town rename-branch" to "git town rename".

Like you said, most Git Town commands are about branches.

- `git town hack` (on a new branch)
- `git town sync` (the current branch)
- `git town switch` (to another branch)
- `git town ship` (the current branch)
- `git town kill` (the current branch)

Why not `git town rename` (the current branch)?

kevgo:
Great idea!

bengeois:
I was thinking the same today when using this command. I can propose a PR if you're up to ?

kevgo:
Yes, please do! All contributions are very welcome no matter how big or small! All the tickets listed here are actionable unless they have the `idea stage` tag.

PR #4133, opened 2024-10-15
Rename "rename-branch" command to "rename"

bengeois:
Fix https://github.com/git-town/git-town/issues/4038

Through this PR, I've tried to rename all occurrences of the `git town rename-branch` command to `git town rename` in the code, documentation and tests

Quite some changes :)
Let me know if you'd like any adjustments!


vendor/github.com/google/go-github/v58/github/repos.go
```diff
@@ -1440,12 +1440,12 @@ func (s *RepositoriesService) GetBranch(ctx context.Context, owner, repo, branch
 	return b, newResponse(resp), err
 }
 
-// renameBranchRequest represents a request to rename a branch.
```
kevgo, line 1443: Please revert the changes made to this file. This is not part of the Git Town source code, but an external dependency that was downloaded into this monorepo for faster and more reliable builds.

An easy way to revert this is running 

```
git checkout main vendor/github.com/google/go-github/v58/github/repos.go
```

kevgo (changes requested):
Thanks for this very clean and well done change! Just one review comment, then this is good to go!

internal/cmd/rename.go
```diff
@@ -47,17 +47,17 @@ When run on a perennial branch:
...
 - registers the new perennial branch name in the local Git Town configuration`
 
-func renameBranchCommand() *cobra.Command {
+func renameCommand() *cobra.Command {
 	addDryRunFlag, readDryRunFlag := flags.DryRun()
 	addForceFlag, readForceFlag := flags.Force("force rename of perennial branch")
 	addVerboseFlag, readVerboseFlag := flags.Verbose()
 	cmd := cobra.Command{
-		Use:   "rename-branch [<old_branch_name>] <new_branch_name>",
+		Use:   "rename [<old_branch_name>] <new_branch_name>",
```
stephenwade, line 55: @kevgo 2 questions:

1. Should we keep an undocumented `rename-branch` command so this isn't a breaking change?
2. Does Git Town have the ability to update the local Git aliases from `git rename-branch` to `git rename` (or detect old alises and suggest that the user update them)?

kevgo: Re 1: yes, I'll add that separately
Re 2: That's a great idea, will also implement separately

vendor/github.com/google/go-github/v58/github/repos.go
bengeois: Yes you're right, missing that wrong changes, I've reverted this change

internal/cmd/rename.go
bengeois: Agree with the second point. I was thinking that day how git-town would handle changes in aliases.

kevgo: Git Town already automatically updates the configuration settings stored in Git metadata when their names change. We can use this mechanism to update the aliases, which are also stored in Git metadata.

kevgo (approved):

PR #4142, opened 2024-10-16
Make shipping resilient against concurrent Git processes

kevgo:


kevgo:
This is now shipped in [Git Town 16.5](https://github.com/git-town/git-town/releases/tag/v16.5.0).

Issue #4145, opened 2024-10-17
Update existing Git aliases for "kill" to "delete"


kevgo:
To make it easier for Git Town users to adjust to https://github.com/git-town/git-town/issues/4039, upgrade existing Git aliases `git kill` to `git delete`.

- add the old and new keys to the [DeprecatedKeys map](https://github.com/git-town/git-town/blob/main/internal/config/configdomain/key.go#L156-L167) which will handle the upgrade automatically
- To verify that the upgrade works, add an end-to-end test to the [Git configuration upgrade tests](https://github.com/git-town/git-town/tree/main/features/config/upgrade).

PR #4160, opened 2024-10-20
WIP: Add old kill alias to DeprecatedKeys map

bengeois:
Fixes https://github.com/git-town/git-town/issues/4145

### DIFF
--- a/internal/config/configdomain/key.go
+++ b/internal/config/configdomain/key.go
@@ -51,6 +51,7 @@ const (
 	KeyDeprecatedCodeHostingDriver         = Key("git-town.code-hosting-driver")
 	KeyDeprecatedCodeHostingOriginHostname = Key("git-town.code-hosting-origin-hostname")
 	KeyDeprecatedCodeHostingPlatform       = Key("git-town.code-hosting-platform")
+	KeyDeprecatedAliasKill				   = Key("alias.kill")
 	KeyDeprecatedMainBranchName            = Key("git-town.main-branch-name")
 	KeyDeprecatedNewBranchPushFlag         = Key("git-town.new-branch-push-flag")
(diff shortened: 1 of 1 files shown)


bengeois:
Hi @kevgo, just one question about the expected behavior.
Adding the key to `DeprecatedKeys` map does nothing. If I also add the key to `keys` maps it updates the alias but not its value, is this the expected behavior?

Alias in my git configuration after the upgrade:

alias.delete = town kill

You can see the associated changes

I'll write the test after my question

kevgo:
You are correct, right now the implementation only renames the config setting key without updating its value. While that's expected behavior for a simple key rename, in this case, we also need to adjust the config's value, so the current behavior falls short.

A potential approach to handle this would be introducing a new data structure, `ConfigUpdates`, similar to `DeprecatedKeys`. Here's an example:

```go
type ConfigSetting struct {
	key   Key
	value string
}

type SettingUpdate struct {
	before ConfigSetting
	after  ConfigSetting
}

var SettingUpdates = []SettingUpdate{
	{
		before: ConfigSetting{
			key:   KeyDeprecatedAliasKill,
			value: "town kill",
		},
		after: ConfigSetting{
			key:   KeyAliasDelete,
			value: "town delete",
		},
	},
}
```

We could then iterate over `SettingUpdates` in the same logic that currently processes `DeprecatedKeys` and apply the necessary updates to both the keys and values. 

What do you think about that approach? Feel free to implement this any way you like.

Also, just a heads-up: you can run `make fix` to resolve any formatting issues automatically.

bengeois:
@kevgo I've tried an implementation with a unit test. Feel free to give me feedback on it

kevgo:
@bengeois excellent work! :bow:

kevgo:
Thanks for implementing this!

bengeois:
Thanks ! I'll implement as well for the `rename-branch` command asap

PR #4208, opened 2024-11-11
Update developer instructions

stephenwade:
- Updates https://github.com/git-town/git-town/blob/main/docs/DEVELOPMENT.md to reflect the current Go version used in the codebase
- Updates https://github.com/git-town/git-town/blob/main/.devcontainer/devcontainer.json to add the Node feature so that `make lint` works inside the container


kevgo:
Thanks @stephenwade, this is super helpful, especially the Docker update! :pray:

PR #4236, opened 2024-11-20
Make resetting the runstate robust

kevgo:
Resetting the runstate can fail if the runstate contains invalid information. This PR makes the `runstate reset` delete the runstate without loading it first.


kevgo:
This is released in [v16.7.0](https://github.com/git-town/git-town/releases/tag/v16.7.0).

Issue #2914, opened 2024-01-03
Incorrect Self hosted BitBucket  PR URL


smaldored:
Hey there,

I am pretty new to using git-town, but I believe I have it set up correctly. `git hack`, `git sync` works. `git propose` works up until the URL it creates for the PR.

We have a self hosted Bitbucket server where our repos stay.

Here is the URL `git propose` creates:
`open https://stash.company.local/project/repo/pull-requests/new?source=git-town-test&dest=project%2Frepo%3Adev`

But, this is an invalid URL. 

It needs to look like this:
`open http://stash.company.local/projects/project/repos/repo/pull-requests?create&targetBranch=dev&sourceBranch=git-town-test`
Notice `http` instead of `https`, `/projects`, `/repos`, `?create` instead of `/new?`, `sourceBranch` instead of `source`, `targetBranch` instead of `dest`

Here is what I get when running `git config --get-regexp git-town` :
```
git-town.code-hosting-platform bitbucket
```

Do I need to use `git-town.code-hosting-origin-hostname=<hostname>` for something? Is there some sort of configuration I am missing?

I've come across #1164 which seems like it is a similar issue to what I am seeing. 

Thanks for the help in advance

kevgo:
Thanks for reaching out. Your configuration looks okay to me. The issue seems that different BitBucket versions use different URLs for creating pull requests. Support for distinguishing different BitBucket versions would have to be added to Git Town.

To create pull requests on BitBucket, Git Town uses the URL reported by one of our users: https://github.com/git-town/git-town/issues/2019. If this URL doesn't work for you, I suspect you work with a different version of BitBucket and your version uses a different URL for creating pull requests. Which BitBucket version do you work with?

`git-town.code-hosting-origin-hostname=<hostname>` is for when you use different SSH identities. See https://www.git-town.com/preferences/code-hosting-origin-hostname for details. In your case, Git Town already correctly uses `stash.company.local` as the hostname, so your configuration seems okay.

To make this work for you, it seems like we would need to add a `bitbucket_version` configuration setting that makes Git Town use the right URL for the bitbucket version. This feature doesn't exist yet.

smaldored:
Thanks for getting back to me.

It looks like we are using `Atlassian Bitbucket v7.4.1` and Bitbucket Server. From the looks of #2019 that is for the Bitbucket Cloud.

[...]

PR #4239, opened 2024-11-20
feat(bitbucket): implement datacenter API

Crocmagnon:
Here's where I'm at for now.

Lots to do still, and I'm really not convinced by the library I found. I'll probably switch to raw HTTP calls, maybe using `carlmjohnson/requests` if you don't mind. Or maybe I'll explore oapi-codegen or something in the same vein if I find an official OpenAPI spec.

At first glance, it looks like the API doesn't support changing the source branch of a PR (and to be fair the UI doesn't support it either). Would you be open to a "not implemented" error in this case? Or should we go the extra mile and close the PR and re-open a new one? I don't think that would be very desirable though, it would mean a loss of context for discussions.

Related to #2914


kevgo:
Very well done so far!

I'm sure you looked at it, but want to double-check: the Bitbucket-Cloud driver is using https://github.com/ktrysmt/go-bitbucket for talking to the Bitbucket API. Is the Bitbucket Datacenter API different enough that this library doesn't work?

I don’t have a strong preference on the library choice for the DataCenter driver—whether you use an existing library or build something custom using a lower-level HTTP library is totally fine. Go with whatever feels most maintainable and appropriate here.

It's okay to not implement changing the source branch if the API doesn't support it. Most other hosting APIs (GitHub, Gitlab, Gitea) also don't support it. In this case the `UpdateProposalSourceFn` method should return `None` instead of a function. See the [GitHub driver](https://github.com/git-town/git-town/blob/affb0de4268464d358dfec794cc8bc0b06ba547b/internal/hosting/github/connector.go#L80-L82) as an example. It's a very optional feature, the business logic degrades gracefully without it.

I think closing the old PR and opening a new one would create as many new problems as it solves, so we are better off not doing that here.

Looking forward to seeing the next iteration!

kevgo:
Found the answer to my question: https://github.com/ktrysmt/go-bitbucket/issues/287

kevgo:
This one is more helpful: https://github.com/ktrysmt/go-bitbucket/issues/10

Crocmagnon:
I just pushed an update to cleanup the code and start fresh. I implemented a single function, and successfully tested it locally:
```console
$ /Users/gaugendr/go/bin/git-town set-parent

Looking for proposal online ... #312
Selected parent branch for "fix/relax-test": feat/app_prefix


branch "fix/relax-test" is now a child of "feat/app_prefix"
```

It should now be easier to work the way through implementing most available features.

kevgo:
Nice progress! A few tools that you might find helpful later:

- run the linters: `make lint`
- run the unit tests: `make unit`
- fix the formatting that some checks might complain about: `make fix`

Crocmagnon:
Thanks :)
I've pushed some more cleanup and also added the search proposal feature. I'll stop here because updating pull requests source/target branches via API looks like a chore and it's required for shipping.

I think this already provides a nice bundle of features, others can implement more if needed.

Crocmagnon:
I have a failing test but it also fails on main:
image

go.mod
```diff
@@ -1,11 +1,14 @@
 module github.com/git-town/git-town/v16
 
-go 1.22
+go 1.22.0
+
+toolchain go1.22.3
```
kevgo, line 5: [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) reports a number of vulnerabilities in Go 1.22.3. Can you either update this to 1.22.5 or higher, or alternatively not add this line in this PR?

internal/hosting/hostingdomain/unsupported_service_error.go
```diff
@@ -8,6 +8,7 @@ func UnsupportedServiceError() error {
 
 This command requires hosting on one of these services:
 * Bitbucket
+* Bitbucket datacenter
```
kevgo, line 11: Thanks for updating this error message. Please also update the end-to-end test that documents/verifies it. It's in features/repo/edge_cases/unsupported_hosting_service.feature.

kevgo:
Based on the logs you shared I assume you run on Windows? If yes, the failing end-to-end test (features/propose/features/branch_deleted_at_remote.feature) does not work on Windows.

go.mod
Crocmagnon: done

internal/hosting/hostingdomain/unsupported_service_error.go
Crocmagnon: Hm strange, I had another one failing but not this one.

`features/propose/edge_cases/unsupported_hosting_service.feature`

I updated the test you indicated and ran `make cuke` but nothing changed.

Crocmagnon:
> Based on the logs you shared I assume you run on Windows? If yes, the failing end-to-end test (features/propose/features/branch_deleted_at_remote.feature) does not work on Windows.

I run on macOS.

go.mod
```diff
@@ -1,11 +1,14 @@
 module github.com/git-town/git-town/v16
 
-go 1.22
+go 1.22.0
```
kevgo, line 3: Apologies for bothering you about this again. Your change on this line sets the Go version on CI to 1.22.0, which triggers the govulncheck warning again. 

```suggestion
go 1.22.9
```

kevgo (approved):
This looks great, thanks for this contribution! :pray: If this is all the functionality you need, and it works for you, I'm happy to merge it in.

go.mod
Crocmagnon: done

Crocmagnon:
> This looks great, thanks for this contribution! 🙏

Thanks 😊 

> If this is all the functionality you need, and it works for you, I'm happy to merge it in.

Yeah pretty much! Proposing changes works, which is the main thing I intend to use. I mostly ship using the web interface anyway and rarely need to update source/target branches. If the need arise I know where to look now 😀 

Thank you very much for your support, it's been very helpful 🙏🏻

Crocmagnon:
Would you like me to squash commits or do you prefer doing it with the web UI?

Crocmagnon:
fixing docs issues

kevgo:
Thanks again for implementing this! I'll release this in the next couple of days.

kevgo:
This is released in [v16.7.0](https://github.com/git-town/git-town/releases/tag/v16.7.0).

PR #4265, opened 2024-11-30
Use correct output format for optional config strings

kevgo:


kevgo:
This is now available in [Git Town 17](https://github.com/git-town/git-town/releases/tag/v17.0.0). Enjoy!

PR #4299, opened 2024-12-11
Simplify error handling in test code

kevgo:
Makes the test code even more concise and explosive

### DIFF
--- a/test/asserts/check.go
+++ b/test/asserts/check.go
@@ -1,7 +1,0 @@
-package asserts
-
-// transparently checks the error value returned by a fallible operation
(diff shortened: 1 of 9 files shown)


test/asserts/check.go
```diff
@@ -0,0 +1,7 @@
+package asserts
+
+// transparently checks the error value returned by a fallible operation
+func Check1[T any](value T, err error) T { //nolint:ireturn
```
ccoVeille, line 4: This maybe?

```suggestion
func NoError1[T any](value T, err error) T { //nolint:ireturn
```

kevgo: I like it, thanks for the suggestion! :pray:

PR #4304, opened 2024-12-12
Auto-update outdated config values only for Git aliases

kevgo:
This update mechanism is intended only for Git Town aliases and shouldn't update anything else.


kevgo:
This is now available in [Git Town 17](https://github.com/git-town/git-town/releases/tag/v17.0.0). Enjoy!

Issue #4165, opened 2024-10-24
git propose: unsupported hosting platform


ruudk:
### Steps to reproduce

When doing OSS contributions I often setup 2 remotes:

```command
$ git remote
fork
upstream
```

But as soon as I do `git town propose` I get

```command
$ git town propose

Error: unsupported hosting platform

This command requires hosting on one of these services:
* Bitbucket
* GitHub
* GitLab
* Gitea
```

Why is this?

### Command output

```shell
It works fine on projects with just 1 remote.
```

kevgo:
I think that's because Git Town uses the `origin` remote to determine the hosting platform. If no such remote exists, Git Town currently stops and gives up. 

One way to make it work better in situations like the one you describe would be to use the `upstream` remote if there is no origin remote.

ruudk:
Why can't it use the first one it finds?

kevgo:
The thinking was to use the platform on which the pull requests gets made. In case different remotes are on different platforms.

Admittedly that's a highly unlikely scenario. 

Technically we could use the remote embedded in the name of the tracking branch of the current branch or the first ancestor branch that has a tracking branch.  

If none of that yields any result, 
you make a good point that Git Town could just use the first available remote, and that would be correct in almost all cases.

kevgo:
As part of syncing, Git Town might need to create tracking branches at the remote. When it does that, it currently runs this command: `git push -u origin <branch>`. With the remote setup above, `origin` would need to be replaced with another remote name. Which remote should Git Town push the branch to in this case? My best guess is "the remote at which other tracking branches exist". This information might not exist if there are no branches. Also, there might be remote branches at different remotes.

stephenwade:
Suggestion: Ask the user and/or create a config setting for which remote to use (that defaults to "origin").

kevgo:
Yeah that might be the best way forward here. I was thinking we could use the remote that the main branch uses, but a config option would be better because it also helps in use cases like #1181.

kevgo:
After thinking about this a few more days, this definitely needs to be a config option. There are too many edge cases that prevent detecting this automatically in a reliable way. 

While we are at it, let's make the `upstream` remote configurable as well.

Idea for how to represent this in the config file:

```toml
[remotes]
dev = "origin"
upstream = "upstream"
```

kevgo:
Yes, configuration can be per project. If you can't add a config file, for example because you work on a fork of an open-source project, you would (in the future) configure this in Git metadata:

[...]

PR #4331, opened 2024-12-14
Config option for origin remote

kevgo:
part of #4165


kevgo:
Missing: document the new config option on the website

PR #4339, opened 2024-12-17
Use fullsize SHA values

kevgo:
Previously Git Town uses 7-digit abbreviations of SHA values. This PR changes them to the full 40 characters to avoid collisions.


kevgo:
This is now available in [Git Town 17.1](https://github.com/git-town/git-town/releases/tag/v17.1.0). Enjoy!

PR #4344, opened 2024-12-18
fix(commands.go/Version/Regex): handle patch being alphanumeric

pratikpc:
Fixes the error message

```
Error: 'git version' returned unexpected output: "git version 2.47.GIT".
Please open an issue and supply the output of running 'git version'
```

If one uses Git's development branches, the version number is controlled by https://github.com/git/git/commits/master/GIT-VERSION-GEN file.

For development versions, the Git versioning can often be

MAJOR.MINOR.GIT

example 2.47.GIT

This however breaks with the current regex which assumes that PATCH starts with a numeric value at least.

### DIFF
--- a/internal/git/commands.go
+++ b/internal/git/commands.go
@@ -772,7 +772,7 @@ func (self *Commands) UndoLastCommit(runner gitdomain.Runner) error {
 
 // Version indicates whether the needed Git version is installed.
 func (self *Commands) Version(querier gitdomain.Querier) (Version, error) {
-	versionRegexp := regexp.MustCompile(`git version (\d+).(\d+).(\d+)`)
+	versionRegexp := regexp.MustCompile(`git version (\d+).(\d+).(\w+)`)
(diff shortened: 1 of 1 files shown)


pratikpc:
As of this moment, the GIT-VERSION-GEN is an RC release, which works with Git Town as the version is 2.48.0-rc0 which starts with a numeric value.   
So you might have to switch to an older commit 29e5596eb8f82015ddb8666079039ae851b8d182 for testing it out.  
https://github.com/git/git/blob/29e5596eb8f82015ddb8666079039ae851b8d182/GIT-VERSION-GEN

kevgo:
#4347 adds an end-to-end test to document this requirement and make sure this change will not get reverted accidentally.

kevgo:
This is now available in [Git Town 17.1](https://github.com/git-town/git-town/releases/tag/v17.1.0). Enjoy!

PR #4355, opened 2024-12-18
Add test for error message with alternative config filename

stephenwade:
```
Feature: print nice error message for invalid alternative config file

  Scenario: Config file with alternative filename and invalid TOML content
```


kevgo:
Thanks for adding this test! :rocket:

Issue #3338, opened 2024-04-13
stack refactoring


kevgo:
### motivation

Not sure if this is something people do a lot or even should should do a lot. But if they do it, they should do it using Git Town. Because Git Town

- provides a workflow engine to do these complex operations step-by-step, with redoing steps after problems
- updates associated proposals
- provides safe undo in case of unexpected issues (which are likely with this type of operation)

### how it works

- all branches to refactor must be in sync
  - the changes made here are complex enough, let's not make them more complex by also syncing
- all branches to refactor must have simple commits
  - no merge commits
  - people who use the `merge` sync-strategy need to `compress` their branch first
- the `--onto` option for rebase will do the heavy lifting here
- update associated proposals

### Branch commands

Some of these commands could be top-level commands, but they aren't used often enough to justify being a top-level command.

- [x] `git town set-parent` - moves the current branch to a different position in the lineage
  - does what the `up`, `down`, `first`, and `last` commands do, and more (moving to a different stack)
  - That would be useful because it's debatable what "up" and "down" mean in a stack. Arguments can be made either way depending whether the main branch is at the top or bottom.
  - Using "up" and "down" repeatedly becomes slow because it always has to update the proposals as well.

  <details>
    <summary>UI</summary>

    The UI is similar to `git town switch`, but moves the highlighted branch up and down.

    ```
    main
      branch-1
    *   branch-2
          branch-3 
    ```

- [x] `git town merge` - merge two consecutive branches in a stack

  <details>
  <summary>example</summary>
  Given this stack:
  
  ```
  main
   \
    branch-1
     \
      branch-2
       \
        branch-3
  ```
  
  And I am on `branch-2`
  When I run `git town merge`
  Then I have this stack:
  
  ```
  main
   \
    branch-1
     \
      branch-3
  ```
  
  And `branch-1` contains the changes of `branch-1` and `branch-2`.
  </details>
- [x] `git town hoist` - detach the current branch from its stack (make it an independent top-level branch)

  <details>
  <summary>example</summary>
  
  Given this stack:
  
  ```
  main
   \
    branch-1
     \
      branch-2
       \
        branch-3
  ```
  
  And I am on `branch-2`
  When I run `git town detach`
  Then I have this stack:
  
  ```
  main
   \
    branch-1
     \
      branch-3
   \
    branch-2
  ```
  
  </details>
- [x] `git town swap` - a branch inside a stack swaps position with its predecessor

  <details>
  <summary>example</summary>
  
[...]

PR #4356, opened 2024-12-18
prepend: --beam option

kevgo:
part of #3338


kevgo:

PR #4358, opened 2024-12-19
Remove duplicate initial commit in test fixtures

stephenwade:
`testruntime.Initialize` already runs `git commit --allow-empty -m "initial commit"`.


kevgo:
Good catch, and this also makes the test suite run faster. Really nice work, and thank you! 🙌

PR #4368, opened 2024-12-19
Fix flaky E2E test by ignoring stderr when running test commands

stephenwade:
I noticed that the tests occasionally fail with the following error:

```
Error: "warning: refname 'dead' is ambiguous." is not a valid Git SHA
```

(example run: https://github.com/git-town/git-town/actions/runs/12420440196/job/34677946315)

This happens if there is a branch with a name that could also refer to a commit hash. In the example, there was a branch named `dead` and a commit with a SHA starting with `dead...`. When the test code lists the commits in the problem branch, it tries to parse the warning and crashes.

I considered writing a special case for `strings.Contains(line, "is ambiguous")`, but I realized it would be simpler to just ignore stderr when running test commands. Then there is no chance for this or any other warning to cause an issue.

Check the commit history to see how I reproduced the issue. I reverted my reproduction test case since I don't see any other "test the tests" type feature files in the repo.


kevgo:
Great catch, really good fix, and thanks for the background on why this happens! :pray:  I’ve run into this problem intermittently as well, but wasn’t able to reproduce it reliably enough to track down the root cause or fix it. The stderr output isn't even used, the business logic only looks at the exit code and stdout.

PR #4371, opened 2024-12-20
fix: bitbucket datacenter open existing PR

Crocmagnon:


Crocmagnon:
Awesome, thanks! Can't wait for it to be available in homebrew :)

PR #4382, opened 2024-12-22
Improve GitHub Actions job names

stephenwade:
Sometimes, GitHub Actions displays the jobs that were run against a commit without displaying their workflow names.

For example, here's what I see after clicking "View details" in a merged PR:

The message at the top of the screenshot says that a commit was merged. Below that is a list of the checks that were run on the commit. All 4 checks are just labeled 'run'.

Updating the job names in the workflows will make it possible to tell which check belongs to which workflow.

### DIFF
--- a/.github/workflows/cuke.yml
+++ b/.github/workflows/cuke.yml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 jobs:
-  run:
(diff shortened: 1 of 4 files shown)


kevgo:
That's a very cool idea!

PR #4385, opened 2024-12-22
Format deadcode exclusions with tabs

stephenwade:
In the `Makefile`, exclusions to the deadcode linter were formatted inconsistently. Most of the lines were indented using spaces, but some of them used tabs, and VS Code uses tabs by default when adding new lines.

This PR updates exclusions to the deadcode linter to be always indented with tabs instead of with spaces.


kevgo:
This reads much better and removes the unnecessary level of indentation. Nicely done!

PR #4389, opened 2024-12-23
Configure core.checkStat in dev container to fix issue with rebase

stephenwade:
More info: https://stackoverflow.com/a/79217456/1102945


kevgo:
Great work on this fix, thank you! I don't personally use the Docker container, so there may still be other issues lurking. Addressing these makes the project more accessible and paves the way for even more contributions!

PR #4391, opened 2024-12-23
Consistently use sentence case for headings

stephenwade:


kevgo:
Thanks for cleaning that up. I experimented with the format of headings a while ago... But the way you format them is better.

PR #4392, opened 2024-12-23
Miscellaneous docs cleanup

stephenwade:
This PR deals with files other than preferences and commands. I plan to review those files in a separate PR.


kevgo:
Thanks, these documentation improvements are at least as important as code improvements!

PR #4403, opened 2024-12-23
Add lint-tests-sorted

legeana:
The number of paths are currently ignored as they do not conform to the
style.

### DIFF
--- a/tools/tests_sorted/tests_sorted.go
+++ b/tools/tests_sorted/tests_sorted.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// file paths to ignore
+var ignorePaths = []string{ //nolint:gochecknoglobals
+	// Need fixing.
+	"internal/cli/dialog/components/checklist_test.go",
+	"internal/cli/dialog/components/test_input_test.go",
+	"internal/cli/format/indent_test.go",
+	"internal/cmd/completions_test.go",
+	"internal/config/configdomain/dryrun_test.go",
+	"internal/config/configdomain/hosting_platform_test.go",
+	"internal/config/configdomain/lineage_test.go",
+	"internal/config/configdomain/ship_delete_tracking_branch_test.go",
+	"internal/config/configdomain/sync_strategy_test.go",
+	"internal/git/commands_test.go",
+	"internal/git/gitdomain/branch_infos_test.go",
+	"internal/git/gitdomain/branch_name_test.go",
+	"internal/git/gitdomain/local_branch_names_test.go",
+	"internal/git/gitdomain/shas_test.go",
+	"internal/gohacks/error_collector_test.go",
+	"internal/gohacks/slice/append_all_missing_test.go",
+	"internal/gohacks/slice/contains_any_test.go",
+	"internal/gohacks/slice/find_many_test.go",
+	"internal/gohacks/slice/get_all_test.go",
+	"internal/gohacks/slice/hoist_test.go",
+	"internal/gohacks/slice/remove_at_test.go",
+	"internal/gohacks/slice/remove_test.go",
+	"internal/gohacks/slice/truncate_last_test.go",
+	"internal/gohacks/stringslice/collector_test.go",
+	"internal/gohacks/stringslice/lines_with_prefix_test.go",
+	"internal/hosting/detect_test.go",
+	"internal/undo/undobranches/branch_changes_test.go",
+	"internal/undo/undobranches/branch_span_test.go",
+	"internal/undo/undoconfig/config_test.go",
+	"internal/vm/optimizer/remove_duplicate_checkout_test.go",
+	"internal/vm/program/program_test.go",
+	"pkg/prelude/option_test.go",
+	"test/datatable/branch_setup_table_test.go",
+	"test/datatable/data_table_test.go",
+	"test/filesystem/copy_directory_test.go",
+	"test/fixture/factory_test.go",
+	"test/fixture/fixture_test.go",
+	"test/output/git_town_output_parser_test.go",
+	"test/subshell/test_runner_test.go",
+	"test/testgit/locations_test.go",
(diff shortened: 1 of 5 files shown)


kevgo:
Love the idea, I was thinking the same! Many of the existing linters in this monorepo have been created in just a couple of hours, by having a LLM create a simple prototype, which I then polished by hand.

kevgo:
If it helps keep this linter simple, all we need here is listing the level-1 unit tests that aren't sorted alphabetically. Consider this example:

```go
func TestFoo(t *testing.T) {
  t.Parallel
  t.Run("level 1", func(t *testing.T) {        // the tests on this level should be sorted alphabetically
    t.Run("level 2", func(t *testing.T) {   // okay if this the tests on this level are in any order
    }
  }
}
```

A super naive implementation that just looks at all lines starting with `  t.Run(` (two spaces in front) and extracts the text in quotes following that prefix could be all we need here.

The [lint-steps](https://github.com/git-town/git-town/tree/main/tools/lint_steps) linter is an example of such a linter. It verifies that the step definitions in [test/cucumber/steps.go](https://github.com/git-town/git-town/blob/main/test/cucumber/steps.go) are sorted alphabetically.

kevgo:
If your tool finds too many unformatted tests, I can help format them after this is shipped.

kevgo:
@legeana, thank you for your thoughtful and valuable contributions to Git Town so far—they are greatly appreciated! To make your experience with Git Town even more fun and productive, I've invited you to join the Git Town organization on GitHub. This is entirely optional, of course, so feel free to accept if it aligns with your preferences.

As a member of the organization, you'll have "commit bits," which means you can work directly on the main Git Town repository instead of using a fork, and you'll receive a Git Town badge on your GitHub profile.

You can also merge your own pull requests (CL). At Git Town, we follow a non-blocking code review process to strike a balance between the benefits of code reviews and maintaining momentum and productivity. Here's how it works:

- All contributions are submitted via pull requests, and every pull request gets reviewed.
- It's okay to merge your pull requests immediately to keep moving forward. These will still be reviewed later, and any actionable feedback gets addressed in subsequent pull requests.
- If you want to get a second opinion on an idea, or it's a more impactful change that would alter user-visible behavior, there is always the option to wait for feedback before merging.
- We are still experimenting with this idea, so feel free to explore it together with us.

The intention is to eliminate disruptions to the momentum and productivity of contributors as much as possible, and reduce the cost of code maintenance so that more of it happens in real life, while keeping the benefits of code reviews.

Whether or not you choose to join the organization, welcome aboard—we value your contributions and hope you enjoy working on this developer productivity tool as much as we do!

legeana:
> thank you for your thoughtful and valuable contributions to Git Town so far

Thank you, I appreciate that.

> To make your experience with Git Town even more fun and productive, I've invited you to join the Git Town organization on GitHub.

I appreciate that. I accepted the invitation.

> At Git Town, we follow a non-blocking code review process to strike a balance between the benefits of code reviews and maintaining momentum and productivity.

That makes sense. Thank you for the detailed explanation. I will probably settle on waiting for review on everything except trivial stuff like typos until I am much more familiar with the project.

legeana:
@kevgo this is now ready for review. One more cleanup that I have in mind is disabling prealloc linter for the entire codebase since it's rarely useful, thus reducing the configuration differences between subdirectories. But this is outside of scope of this PR.

tools/tests_sorted/tests_sorted.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		issues = append(issues, fileIssues...)
+		return nil
+	})
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	if len(issues) > 0 {
+		fmt.Fprintln(os.Stderr, issues)
```
kevgo, line 80: Nitpick: Given that the point of a linter is to print lint warnings, one could argue that these lint warnings are the regular output of this tool and should therefore go to STDOUT. STDERR is for when the linter itself doesn't work properly. What do you think?

kevgo (approved):
Thanks for the hard work on this! I'm really excited to start using this linter. It’s going to make navigating and maintaining the tests so much smoother. Previously, I’ve had to open countless PRs just to re-sort things alphabetically, only to see it slip back into chaos over time. :laughing: Thanks for solving this once and for all! 🚀

kevgo:
I’d like to request that we don't disable the prealloc linter across the entire repo. The end-to-end test suite currently calls the Git Town executable over 1600 times. At this scale, even small inefficiencies can compound, making it worthwhile to tackle some of the low-hanging fruit around performance.

For context, there was a time when running `make cuke` took 18 minutes. Through targeted, data-driven optimizations, we've managed to bring that down to just 22 seconds today (on Apple Silicon running Asahi Linux—macOS remains slower due to filesystem bottlenecks). This makes it possible to lean on the end-to-end tests as part of development.

Here are a couple of examples of optimizations that made a meaningful impact on the test suite’s performance:

- https://github.com/git-town/git-town/pull/1441 found that creating a Git repo from scratch for each test was a major bottleneck. The solution involved setting up a fully configured Git environment once at the start of the test suite and cloning it for each test.
- https://github.com/git-town/git-town/pull/1447 found that disabling the garbage collector significantly reduces test suite runtime on CI (which is a severely performance constrained environment).

I'm curious about how you see this, and if I'm missing something. Let me know if you’d like more background or additional details! :slightly_smiling_face:

tools/tests_sorted/tests_sorted.go
legeana: Done.

legeana:
> Thanks for the hard work on this! I'm really excited to start using this linter. It’s going to make navigating and maintaining the tests so much smoother. Previously, I’ve had to open countless PRs just to re-sort things alphabetically, only to see it slip back into chaos over time. 😆 Thanks for solving this once and for all! 🚀

Still need to clean up the backlog, but yea, that was my thinking when you first asked me to do this. Such styleguides are not enforceable without a linter in my experience.

> I’d like to request that we don't disable the prealloc linter across the entire repo.

Well, it's going to be a separate PR. Since you made multiple comments about this particular topic in multiple places let's me combine these threads outside of this PR.

PR #4390, opened 2024-12-23
Add MergeNoFastForward

legeana:
Add unit tests for both MergeFastForward and MergeNoFastForward.

Add unit tests for both MergeFastForward and MergeNoFastForward.
Commands.CommitsInPerennialBranch is used because NewCommitMessages is
more convenient for testing than the test_commands's CommitsInBranch.

Commands.CommitsInBranch(branch, None) ignores the branch name as it
is just a wrapper around CommitsInPerennialBranch if the parent is
None, so it is more readable to just call the CommitsInPerennialBranch
directly.

See #4381.

Issue #4381, opened 2024-12-21
[FR] always-merge ship strategy


legeana:
### Your message to us

Hi,
I am wondering if it would be possible to support an equivalent of `git merge --no-ff` merge strategy, e.g. `git town ship -s no-fast-forward`.

Sometimes when developing a feature that consists of multiple small but related commits it can be nice to keep them grouped together via an explicit merge commit.

https://www.git-town.com/preferences/ship-strategy currently only supports
- [api](https://www.git-town.com/preferences/ship-strategy#api) which requires hosting configuration
- [fast-forward](https://www.git-town.com/preferences/ship-strategy#fast-forward) which doesn't create the merge commit
- [squash-merge](https://www.git-town.com/preferences/ship-strategy#squash-merge) that squashes the commits

kevgo:
Thanks for reaching out about this. Yes, it's possible to add additional ship strategies if the need arises. That said, most Git Town users don't actually use `git town ship`. Instead, they typically 
- create a pull request for their feature branch with `git town propose`, 
- merge the branch via the web UI of their code hosting service (or some sort of merge queue) 
- run `git sync --all` to clean up the local counterparts of shipped branches and sync the shipped changes into their other feature branches.

Are you actively using `git town ship` and need the `--no-ff` strategy?

legeana:
I started using git-town very recently and I find it incredibly convenient. I don't use git at work so my experience with git-town is mostly small hobby projects that I either own or co-own where the code review isn't always necessary. As such using `git town propose` can often be an overkill: there is no need to open a pull request if no one is going to review it.

Before adopting git-town most of the time I effectively used an equivalent of a fast-forward strategy, except when I wanted to highlight a major feature/rework as a single unit of work, e.g. ![introducing an sqlite into my hobby project](https://github.com/user-attachments/assets/b0f93bb7-43ea-46e8-8a7c-75c68cc5f7d1). Here I used `git merge --no-ff sqlite-registry`. This would translate to a hypothetical `git town ship -s no-fast-forward`.

This isn't a very common thing I do, so I could live without it, but I do like consistency in tooling, this is why I opened this feature request.

I would be happy to work on implementing this feature in git-town if you would be willing to accept it. Of course I understand adding every feature increases the maintenance burden on the project, so please let me know what you think.

kevgo:
Thanks for the background. I'd be happy to merge a PR that adds the `no-ff` ship strategy! Some pointers to get started implementing this:

[...]

PR #4404, opened 2024-12-23
Add "always-merge" strategy (inaccessible to users for now)

legeana:
"always-merge" clearly states that this is a merge commit. "always" also
implies that the merge commit is always going to be created.

Alternative names considered:
- "no-fast-forward" is longer than "forced-merge" and relies on users
  knowing the implementation detail of the merge command: --no-ff flag.
  It is reasonable to call the internal operation MergeNoFastForward
  (see #4390) as it clearly resembles `merge --no-ff`, but it shouldn't
  be exposed the same way to the end users. Plus having negative words
  in the name makes it harder to read.
- "no-ff" is inconsistent with "fast-forward" and "squash-merge". Both
  use full words instead of abbreviations.
- "forced-merge" clearly states that this is a merge commit. "forced"
  also implies that the merge commit is always going to be created, but
  it does have a connotation that this isn't a good idea to do so,
  similar to `git push --force`. See
  <https://github.com/git-town/git-town/pull/4404#issuecomment-2560555323>.

See #4381.

### DIFF
--- a/internal/cli/dialog/ship_strategy.go
+++ b/internal/cli/dialog/ship_strategy.go
@@ -34,6 +34,7 @@ func ShipStrategy(existing configdomain.ShipStrategy, inputs components.TestInpu
 			Data: configdomain.ShipStrategyFastForward,
 			Text: `fast-forward: in your local repo, fast-forward the parent branch to point to the commits on the feature branch`,
 		},
+		// TODO: #4381 - add ShipStrategyAlwaysMerge
 		{
 			Data: configdomain.ShipStrategySquashMerge,
 			Text: `squash-merge: in your local repo, squash-merge the feature branch into its parent branch`,
--- a/internal/cmd/ship/core.go
+++ b/internal/cmd/ship/core.go
@@ -121,6 +121,9 @@ func executeShip(args []string, message Option[gitdomain.CommitMessage], dryRun 
 			return err
 		}
 		shipProgramFastForward(prog, sharedData, mergeData)
+	case configdomain.ShipStrategyAlwaysMerge:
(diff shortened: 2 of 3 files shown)


kevgo (approved):
LGTM! Thanks for putting so much thought into evaluating the naming here—it's clear you've considered this carefully. In general, I think it's okay to lean on Git terminology, since Git Town users are likely familiar with it.

One additional candidate to consider: `always-merge`. This name emphasizes that the strategy consistently creates a merge, which might make the behavior a bit clearer. It avoids potential confusion with "force", which in Git often carries the connotation of bypassing safety or correctness checks (e.g., force-push to overwrite data). "Always" also feels a bit softer and more descriptive than "force."

That said, I don’t feel strongly about this and fully support your final choice. Thanks again for the thoughtful work here!

stephenwade:
> "forced-merge" clearly states that this is a merge commit.

I disagree. To me, "forced" means "even though it might not be a good idea", which as a user would make me worried about Git Town using `git push --force` or `git reset` on my branches.

I would prefer `merge-commit` or `always-merge`. I think it is not clear that "forced" means "always use a merge commit" instead of the typical meaning of `--force`.

legeana:
> > "forced-merge" clearly states that this is a merge commit.
> 
> I disagree. To me, "forced" means "even though it might not be a good idea", which as a user would make me worried about Git Town using `git push --force` or `git reset` on my branches.
> 
> I would prefer `merge-commit` or `always-merge`. I think it is not clear that "forced" means "always use a merge commit" instead of the typical meaning of `--force`.

I agree with you, `always-merge` is better. PTAL.

stephenwade (approved):

stephenwade:
Looks great, thanks!

PR #4424, opened 2024-12-24
Fix flags Read*Func comments

legeana:
- Fix "dry-run" copy&paste.
- Add consistent quotes around flag names to avoid confusion with normal words in the description.
- Add type name prefix as per <https://google.github.io/styleguide/go/decisions#doc-comments>.


kevgo:
Thanks for cleaning this up!

PR #4425, opened 2024-12-25
Add sync --strategy flag

legeana:
This allows overriding sync strategy without changing git-town config.

For example in git-town the sync-feature-strategy is `merge`, and this
flag would allow a user to override this default on per-sync basis.

Current implementation doesn't warn if a user tries to override a
strategy for a branch type with a fixed strategy such as observed
branch.

Alternative implementation considered:

Instead of overriding the config we could instead pass around an
explicit strategyOverride parameter. This has several advantages:

- It is clear that the strategy can be overridden. Modifying config is
  subtle, and while it follows an existing practice of changing the
  config (e.g. `validatedConfig.NormalConfig.ShipStrategy =
  shipStrategyOverride`) it doesn't make it more obvious.
- It is easier to determine that a strategy override is unused for
  certain branch types and return an error.
- Certain helper functions already accept an explicit `syncStrategy`,
  e.g. `FeatureBranchProgram`.

It also has a several disadvantages:
- More parameters to pass around. Longer function signatures make them
  harder to read.
- We could add a new optional `SyncStrategyOverride` member to
  `BranchProgramArgs`, but it already has a `Config` member, so this
  increases a possibility of confusion which sync strategy to use. And
  since existing code directly accesses `BranchProgramArgs.Config` to
  query an appropriate `SyncStrategy` it is very easy to introduce a
  subtle bug where we don't apply the override appropriately.
- The strategy override feature is effectively spread over the entire
  codebase. We need to type `args.SyncStrategyOverride.GetOrElse(
  args.Config.NormalConfig.Sync<Type>Strategy.SyncStrategy())` every
  time we want to access the value.


kevgo (commented):
Thanks, this is a useful feature that I was also contemplating to add. 

There is one issue that we need to think through. [Prototype branches](https://www.git-town.com/commands/prototype.html) use the [sync-prototype-strategy](https://www.git-town.com/preferences/sync-prototype-strategy.html) and perennial branches use the [sync-perennial-strategy](https://www.git-town.com/preferences/sync-prototype-strategy.html). If I'm on a prototype branch, and I run `git town sync --feature-strategy=rebase`, should it apply the overridden sync strategy or not? Technically no, because I was specifically overriding the `sync-feature-strategy` and the branch uses the `sync-prototype-strategy`. But it would also feel wrong that Git Town ignores a flag I provided. 

Possible solutions:

1. Rename this flag to `--strategy` and then always apply the given strategy no matter what the branch type is.
2. Add `--prototype-strategy` and `--perennial-strategy` flags. Display a warning when the flag is being ignored. Ideally also point the user to the correct flag to use.

A downside of (1) is when I run `git town sync --all --strategy=compress`, it would now compress all my branches, including the main branch and perennial branches. So (1) seems the better option when syncing a single branch, and (2) seems the better option when syncing all branches.

website/src/commands/sync.md
```diff
@@ -46,6 +46,16 @@ developers.
...
 
+### --feature-strategy &lt;name&gt;
+
+Overrides the configured
+[sync-feature-strategy](../preferences/sync-feature-strategy.md). Allowed
+values:
+
+- `compress`
+- `merge`
+- `rebase`
```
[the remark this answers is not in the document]
legeana: Yea that's fine. I added these values for consistency. See #4428.

legeana:
> Thanks, this is a useful feature that I was also contemplating to add.
> 
> There is one issue that we need to think through. [Prototype branches](https://www.git-town.com/commands/prototype.html) use the [sync-prototype-strategy](https://www.git-town.com/preferences/sync-prototype-strategy.html) and perennial branches use the [sync-perennial-strategy](https://www.git-town.com/preferences/sync-prototype-strategy.html). If I'm on a prototype branch, and I run `git town sync --feature-strategy=rebase`, should it apply the overridden sync strategy or not? Technically no, because I was specifically overriding the `sync-feature-strategy` and the branch uses the `sync-prototype-strategy`. But it would also feel wrong that Git Town ignores a flag I provided.
> 
> Possible solutions:
> 
> 1. Rename this flag to `--strategy` and then always apply the given strategy no matter what the branch type is.
> 2. Add `--prototype-strategy` and `--perennial-strategy` flags. Display a warning when the flag is being ignored. Ideally also point the user to the correct flag to use.
> 
> A downside of (1) is when I run `git town sync --all --strategy=compress`, it would now compress all my branches, including the main branch and perennial branches. So (1) seems the better option when syncing a single branch, and (2) seems the better option when syncing all branches.

According to [branch-types documentation](https://www.git-town.com/branch-types.html#prototype-branches): Syncing prototype branches follows the [sync-prototype-strategy](https://www.git-town.com/preferences/sync-prototype-strategy) or - if this setting isn't present - the [sync-feature-strategy](https://www.git-town.com/preferences/sync-feature-strategy). This is implemented in `NewSyncPrototypeStrategyFromSyncFeatureStrategy`. So using feature-strategy for prototype branches is an already established thing that git-town does. Making this behaviour sensible to the end user when implementing strategy overrides is a separate question however. We would have to consider how to handle the following situations:

| config feature strategy | config prototype strategy | feature strategy override | effective prototype strategy |
| - | - | - | - |
| unset | unset | unset | default prototype strategy |
| set | unset | unset | config feature strategy |
| unset | set | unset | config prototype strategy |
| set | set | unset | config prototype strategy |
| unset | unset | set | *feature strategy override* since this mimics the existing behaviour |
| set | unset | set | *feature strategy override* since this mimics the existing behaviour |
| unset | set | set | *feature strategy override* or *config prototype strategy*, this is confusing and the override priority is unclear |
| set | set | set |  *feature strategy override* or *config prototype strategy*, similarly confusing behaviour |

It is unclear to me whether we should override `prototype-sync-strategy` if it's explicitly set in the config, but only `--feature-strategy` is specified.

As far as I can see there are two major types of branches in git-town:
- Perennial branches.
- Short-lived branches:
  - Feature branches.
  - Prototype branches, effectively local feature branches. It's not unreasonable to treat them similarly.
  - Contribution branches. They are not synced, so let's ignore them here.
  - Observed branches. They are always synced with rebase strategy, so we can ignore them here.
  - Parked branches. Similarly, as they are not synced they can be ignored.

So I don't believe applying `--feature-strategy` flag to perennial branches makes sense. They belong to completely separate categories.

On a separate UX concern, if `--all` is set it never makes sense to print warnings if an override does not apply. However I see some benefit for warning a user that `--feature-strategy` doesn't apply to a single synced branch.

On the other hand one of the ways I see this flag used is the following. Sync strategies for feature/prototype branches are owned by the user, not the project. The branches themselves are owned by the user. It's totally fine to be able to provide sensible defaults for a project, but as a user I may want to have my own global overrides. As I already have `alias.sync = town sync`, I may as well make it  `alias.sync = town sync --feature-strategy=rebase [--prototype-strategy=...]` and have my preferred workflow regardless of the project I am contributing to. Unfortunately this will lead to constant warning messages in the solution proposed in (2).

What are your thoughts?

kevgo:
Thanks for this brilliant analysis! For context, here is some additional background on configuration settings and their precedence. This is currently not well [documented](https://www.git-town.com/preferences.html) - I'll fix that. There are three types of configuration data (soon to be four if this PR ships):

1. Team-wide repo-specific settings: These go into the [configuration file](https://www.git-town.com/configuration-file.html) and provide defaults for all team members for this repo.
2. User-specific Git configuration: Each team member can configure their preferred settings using Git metadata. For instance, if I always want to use the `rebase` sync-feature-strategy across all repositories, I would run: `git config --global git-town.sync-feature-strategy rebase`. This setting overrides (1).
3. User-specific repo-specific Git configuration: Local Git metadata takes precedence over (1) and (2). For example, if I want `rebase` as the default strategy for all repositories except in the `foo` repo I want to use merge, I'd first configure the global setting in (2), then run in the `foo` repo: `git config git-town.sync-feature-strategy merge` (no `--global` flag).
4. CLI flag introduced in this PR: This new flag would allow overriding all of the above on a per-command basis. It seems particularly useful for one-off sync operations or experimenting with different sync strategies.

Does this make sense? Happy to improve this if there are better ideas.

legeana:
> * User-specific Git configuration: Each team member can configure their preferred settings using Git metadata. For instance, if I always want to use the `rebase` sync-feature-strategy across all repositories, I would run: `git config --global git-town.sync-feature-strategy rebase`. This setting overrides (1).
> * User-specific repo-specific Git configuration: Local Git metadata takes precedence over (1) and (2). For example, if I want `rebase` as the default strategy for all repositories except in the `foo` repo I want to use merge, I'd first configure the global setting in (2), then run in the `foo` repo: `git config git-town.sync-feature-strategy merge` (no `--global` flag).

I actually tried fumbling with these settings but didn't figure out the correct git-config key myself. That helps a lot to me personally and it's absolutely very useful to document how to configure this. I would love to see something like this (disregarding the style) in the official documentation: "to override project settings use `git config [--global] git-town.<preference> value` where preference is the exact key from https://www.git-town.com/preferences.html, e.g. `git config git-town.sync-feature-strategy merge`. It might be also useful to allow `git town config` to configure global/local git-town settings.

Back to the current PR. Now it makes sense to rethink on how `sync --xxx-strategy` flags can be used. As I don't see any value in setting all of them unconditionally for all syncs, my concern about unnecessary warnings no longer applies. And having flag overrides match specific config options is already an established style and in my opinion is very easy to understand. Thus I propose the following: Have trivial per-branch-type override flags without any fallback logic.

- `--feature-strategy` applies only to feature branches and produces a warning if the current branch isn't a feature branch.
- `--perennial-strategy` applies only to perennial branches and produces a warning if the current branch isn't a prototype branch. Future PR.
- `--prototype-strategy` applies only to prototype branches and produces a warning if the current branch isn't a prototype branch. Future PR.
- `--all` suppresses all warnings above.

WDYT?

kevgo:
This is shaping up nicely—we’re getting close to a great solution! Here’s a suggestion to simplify the approach further based on your outline:

- Introduce a single `--strategy` flag: This flag applies universally when syncing a single branch, regardless of its type. `git town sync --strategy=compress` is pretty intuitive and self-explanatory.
- Prevent combining `--strategy` with `--all`: When syncing all branches, the strategy cannot be overridden via CLI flags. I don’t foresee a practical use case where someone would run something like `git town sync --all --feature-strategy=foo --perennial-strategy=bar --prototype-strategy=baz`, when those strategies differ from the configured Git metadata.

How does that sound to you?

features/sync/current_branch/feature_branch/override/sync_strategy_override.feature
```diff
@@ -0,0 +1,54 @@
+Feature: override the configured sync-feature-strategy
+
+  Background:
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And Git Town setting "sync-feature-strategy" is "merge"
```
kevgo, line 8: I just shipped a change that will break this. Here is the fix:

```suggestion
    And Git setting "git-town.sync-feature-strategy" is "merge"
```

legeana: Thanks!

legeana:
> * Prevent combining `--strategy` with `--all`: When syncing all branches, the strategy cannot be overridden via CLI flags. I don’t foresee a practical use case where someone would run something like `git town sync --all --feature-strategy=foo --perennial-strategy=bar --prototype-strategy=baz`, when those strategies differ from the configured Git metadata.

I think it makes sense. I updated the CL and its description, please take another look. There are 2 main approaches I see: updating the config, or passing an override explicitly. Both have advantages and disadvantages, and I'm not so sure which one is better.

features/sync/all_branches/strategy_override/error.feature
```diff
@@ -0,0 +1,19 @@
...
+      | main   | origin        | main commit  |
+      | alpha  | local, origin | alpha commit |
+    And the current branch is "alpha"
+    When I run "git-town sync --all --strategy=rebase"
+
+  Scenario: result
+    Then Git Town prints the error:
+      """
+      sync doesn't support --all and --strategy flags at the same time
+      """
```
kevgo, line 19: I would also specify that Git Town doesn't run any commands in this situation.

```suggestion
      """
    And Git Town runs no commands
```

features/sync/current_branch/feature_branch/strategy_override/sync_strategy_override.feature
```diff
@@ -0,0 +1,54 @@
+Feature: override the configured sync-feature-strategy
+
+  Background:
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And Git setting "git-town.sync-feature-strategy" is "merge"
```
kevgo, line 8: Maybe move this line down to above line 16 so that the currently set `sync-feature-strategy` and the override are close to each other? This would make it easier to see which strategy is set and what it's overridden to.

internal/cli/flags/sync_strategy.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+		strategyOpt, err := configdomain.ParseSyncStrategy(value)
+		if err != nil {
+			return None[configdomain.SyncStrategy](), err
+		}
+		strategy, hasStrategy := strategyOpt.Get()
+		if !hasStrategy {
+			return None[configdomain.SyncStrategy](), nil
+		}
+		return Some(strategy), nil
```
kevgo, line 32: I think this can be simplified.

```suggestion
		return strategy, nil
```

kevgo (approved):
Thanks a ton for implementing this and updating the documentation! Just a couple of minor nitpicks, and we’re good to merge. 🚀

kevgo:
> There are 2 main approaches I see: updating the config, or passing an override explicitly. Both have advantages and disadvantages, and I'm not so sure which one is better.

I agree. Passing the override is technically more precise, but causes more complex logic when evaluating the sync strategy to use later. I think your approach of overriding the config is semantically also correct, and simpler, so I'm okay with going that route.

website/src/commands/sync.md
```diff
@@ -57,6 +57,17 @@ developers.
 Use the `--dry-run` flag to test-drive this command. It prints the Git commands
 that would be run but doesn't execute them.
 
+### --strategy &lt;name&gt;
```
stephenwade, line 60: Please update this heading to match the new style (suggested change below), move this option between `--stack` and `--detached`, and add `[--strategy <name>]` to the command summary in the same position.

```suggestion
#### `--strategy <name>`
```

stephenwade (commented):
We just updated the headings style in #4437. See the review comment for details.

website/src/commands/sync.md
stephenwade: Looks great! 🎉

stephenwade (approved):

stephenwade:
@legeana I'm interested in using this feature. Do you have any plans to continue working on this PR?

legeana:
@stephenwade apologies, was very busy with work.

I have correctness concerns with the current implementation. I am unsure that the proposed implementation is valid, so I don't think this should be submitted in the current form.

I will get back to this this weekend. Sorry for stalling this.

stephenwade:
No worries, thank you for the update!

legeana:
I was trying to remember the issue with this proposal, and to the best of my memory it's along these lines:
- As implemented in this PR `--sync-strategy=X` will set sync strategy to X for all branch types.
- If the sync of the current branch triggers a sync of another branch then we inadvertently apply X strategy to another branch. This is a potentially surprising behaviour.

I was unable to reproduce this issue, but I do vaguely remember seeing this type of behaviour. Though there is a chance this isn't an issue now that `--stack` and `--all` are disabled. So this may only be a hypothetical. But I don't feel comfortable pushing this feature in the current state as it may break an existing repository. And unfortunately I don't have the time to try to fix it right now.

If you think you can make it work better feel free to grab my work in progress and either add appropriate tests, or redo it, I don't mind.

Apologies again for not making progress.

PR #4427, opened 2024-12-25
Add an optional message parameter to MergeNoFastForward

legeana:
- Unlike MergeFastForward where there are no additional commits created, MergeNoFastForward creates a merge commit and we can specify a custom commit message. Since `git town ship` supports `--message` flag it is a good UX to support custom commit messages.
- SquashMerge doesn't have a commit message parameter because it doesn't create a commit by itself, instead it adds all the files from other commits into the stage area, and a commit is created explicitly with `Commit(...)` later.

See #4381.


kevgo:
Nicely done!

PR #4428, opened 2024-12-25
Remove duplicated ship strategy values from ship.md

legeana:
See <https://github.com/git-town/git-town/pull/4425#discussion_r1897394679>.


kevgo:
Thanks!

Issue #4429, opened 2024-12-26
Use code formatting for all command-line input


stephenwade:
### Your message to us

We need to change all command-line input to use code formatting. This will allow us to turn on the `smart-punctuation` setting in mdBook (to get nicer quotation marks) without it converting "--" to "–" in command-line options.

Ref: #4386, #4408

PR #4430, opened 2024-12-26
Use code formatting for all command summaries

stephenwade:
Part of #4429.

This PR adds an mdBook preprocessor to add custom markup to code blocks with the "command-summary" language.

mdBook provides the preprocessor a JSON array with context and the book contents over standard input. The preprocessor may modify the book contents and must output the book contents JSON to standard output.

The preprocessor is written in JavaScript because I'm more familiar with the language, and because I had a painful time trying to deal with JSON in Go.

<details><summary>My struggle with JSON in Go</summary>

Each book item in the `sections` array is either an object with a `"Chapter"` key, an object with a `"PartTitle"` key, or the string `"Separator"`. Go doesn't have a way to represent that a struct item can be one of multiple types (e.g. struct or string), so I got a couple hours into figuring out how to write custom marshaling code to deal with the JSON conversion before I got too confused. To me, it makes more sense to write this in a language that can deal with JSON natively (and that I know very well).

</details>

https://github.com/user-attachments/assets/756f8d62-24c8-4818-9acd-bffc9dcceb9a


stephenwade:
@kevgo PTAL – I'll wait to convert the rest of the command summaries to this format until you okay the addition, since it is a new kind of addition to the project (an mdBook preproecssor).

tools/mdbook_preprocessor/mdbook_preprocessor.js
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * For example:
+ *
+ * ```command-summary
+ * git town append [--prototype] <branch-name>
+ * ```
+ *
+ * will become:
+ *
+ * <pre><code><div class="gt-command-summary" style="padding-left: 16ch; text-indent: -16ch"
```
kevgo, line 114: Given that there is an additional CSS file now, why are you defining styles inline here?

kevgo: Nevermind, I see these values come from variables.

stephenwade:
I think the preprocessor is easy enough to understand, so I'll leave it in. We can always hard-code the markup later if it is needed for some reason.

tools/mdbook_preprocessor/mdbook_preprocessor.js
```diff
@@ -0,0 +1,151 @@
...
+    return;
+  }
+  // bookItem is { "Chapter": Chapter } or { "PartTitle": string }
+  if ("PartTitle" in bookItem) {
+    return;
+  }
+  // bookItem is { "Chapter": Chapter }
+
+  const chapter = bookItem.Chapter;
+  processChapter(chapter);
```
kevgo, line 79: Nitpick: you could inline the variable `chapter`.

```suggestion
  processChapter(bookItem.Chapter);
```

tools/mdbook_preprocessor/mdbook_preprocessor.js
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @param {string} code
+ * @returns {string}
+ */
+function processCommandSummary(code) {
+  return `<pre><code>${
+    code
+      .split("\n")
+      .map(line => {
+        const command = line.match(/^.*?(?=$| [\[<])/)?.[0];
```
kevgo, line 130: Would the regex still work if you omit the `^.*` at the beginning? It seems like the main goal here is to capture something in the middle of the line, so matching the start of the line and then allowing arbitrary characters afterwards might be unnecessary.

website/src/SUMMARY.md
```diff
@@ -42,7 +42,6 @@
     - [config.get-parent](commands/config-get-parent.md)
     - [config.remove](commands/config-remove.md)
     - [config.setup](commands/config-setup.md)
-    - [setup](commands/config-setup.md)
```
kevgo, line 45: Whoops, thanks for removing this double-entry!

website/src/commands/append.md
```diff
@@ -36,6 +38,11 @@ main
 *   feature-2
 ```
 
+### --prototype / -p
```
kevgo, line 41: What's the motivation to move this block up? Right now they are sorted alphabetically. I'm fine with a different sorting order, as long as its clear which one it is so that I can follow it too :slightly_smiling_face:

website/src/commands/compress.md
```diff
@@ -1,6 +1,8 @@
 # git town compress
 
-> git town compress [--message &lt;text&gt;] [--stack]
+```command-summary
+git town compress [-m <text> | --message <text>] [-s | --stack] [--dry-run] [-v | --verbose]
+```
```
kevgo, line 5: This is a real improvement in readability. No more HTML-encoded characters!

kevgo (approved):
Thank you so much for this fantastic improvement to the website! The wording changes are also really nice. I’m excited to get this published!

tools/mdbook_preprocessor/mdbook_preprocessor.js
stephenwade: No. The goal is to capture from the beginning of the line until the end or until the first parameter. It uses a [lookahead assertion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Lookahead_assertion) (`(?=...)`) to match but not consume `$` (end of line) or `␣[\[<]`. So the part that gets returned is the text matched by `.*?`, with the lookahead assertion to force the `.*?` to consume as much as I want it to.

stephenwade: I'll add a code comment explaining this.

kevgo: Thanks for the explanation, looks good to me now!

website/src/commands/append.md
stephenwade: The motivation is to move arguments common to many commands to the end of the list. Specifically, I moved any instances of `[-d | --detached]`, `[--dry-run]`, and `[-v | --verbose]` to the end of the options list.

I think users visiting a page will be more interested in the options specific to that command.

I added a style guide to DEVELOPMENT.md so we can both remember the sorting decision: 728ecb379

kevgo: I like it, thanks for improving this!

PR #4431, opened 2024-12-26
Automatically update plugin versions in dprint-changelog.json

stephenwade:


kevgo (approved):
Wow, I totally forgot that separate dprint config. Good catch!

PR #4434, opened 2024-12-26
Remove Node from dev container

stephenwade:
This PR makes it possible to use Node-based tools without Node being installed on the system. It ensures that all Node tools are run with run-that-app. To make that possible, it also replaces Yarn with npm.

I tested this PR by removing the "node" feature from the dev container and verifying that I can still run all the Makefile targets.


stephenwade:
FYI @kevgo, there is a bug in run-that-app. I get a "node: command not found" error when I use `npm` or `npx`. I work around it in this PR because I couldn't figure out how to fix it in the original Rust code.

I filed https://github.com/kevgo/run-that-app/issues/298 for this issue.

kevgo:
Thanks for slimming down the dev-container by completely omitting Node. Nobody likes heavy Docker containers. This will improve the dev experience for everyone using it.

Also, major props for finding a clever workaround to the run-that-app bug—really impressive problem-solving! 🙇‍♂️

PR #4440, opened 2024-12-28
Remove all `//nolint` comments for disabled linters

stephenwade:


kevgo:
Good catch! It's interesting that golangci-lint flags unnecessary exclusions, but seems to miss cases where linters are completely disabled.

PR #4442, opened 2024-12-28
Update RTA to 0.9

kevgo:


kevgo:
@stephenwade Thanks again for identifying and reporting the RTA bug, as well as for implementing the workaround in the Makefile. This PR includes your proposed fix for the RTA issue. I've tested it on my machine (without Node.js installed) and it works as expected. Could you please verify on your end to ensure everything runs smoothly before I merge this in and undo your workaround? Thanks!

stephenwade:
The issue is fixed! I was able to run `make lint` in the dev container that doesn't have Node installed.

Please also update the RTA version in website/Makefile and remove the "work around "node: command not found"" comments.

kevgo:
Thanks for the review!

PR #4443, opened 2024-12-29
MergeNoFastForward: stop relying on default value of --edit in git merge

legeana:
`git merge` determines the default value of the `--edit` flag based on whether git runs on a tty. This can be convenient for bare git where it can be used both interactively and non-interactively. However git-town is always interactive but cucumber tests run subcommands without tty. This creates a problem where `git merge` behaves differently in cucumber tests compared to real interactive use, e.g. it doesn't run an editor even though we use `and close the editor` or `and enter "..." for the commit message`, which can be extremely confusing.

Assuming that git-town will always run interactively it is much easier to set an appropriate `--edit` value explicitly:
- If an explicit message is set, we don't need to worry about `--edit`, git will not run an editor regardless of tty.
- If a message is not present and we run the command interactively we should prompt a user to edit the default message, so we use `--edit`.
- If a message is not present and we run the command non-interactively we should accept the default message, so we use `--no-edit`.

The behaviour is controlled by the `useDefaultMessage` parameter named after a similar parameter in `Commands.Commit()`.

See also:
- [default_edit_option in builtin/merge.cc](https://github.com/git/git/blob/306ab352f4e98f6809ce52fc4e5d63fb947d0635/builtin/merge.c#L1082-L1106).
- #4381.

### DIFF
--- a/internal/git/commands.go
+++ b/internal/git/commands.go
@@ -532,10 +532,21 @@ func (self *Commands) MergeFastForward(runner gitdomain.Runner, branch gitdomain
 }
 
 // MergeNoFastForward merges branch into the current branch and always creates a merge commit.
-func (self *Commands) MergeNoFastForward(runner gitdomain.Runner, message Option[gitdomain.CommitMessage], branch gitdomain.LocalBranchName) error {
+func (self *Commands) MergeNoFastForward(runner gitdomain.Runner, useMessage configdomain.UseMessage, branch gitdomain.LocalBranchName) error {
 	gitArgs := []string{"merge", "--no-ff"}
-	if messageContent, has := message.Get(); has {
-		gitArgs = append(gitArgs, "-m", messageContent.String())
+	switch {
+	case useMessage.IsCustomMessage():
+		message := useMessage.GetCustomMessageOrPanic()
+		gitArgs = append(gitArgs, "-m", message.String())
+	case useMessage.IsUseDefault():
+		gitArgs = append(gitArgs, "--no-edit")
+	case useMessage.IsEditDefault():
+		// Unlike `git commit`, `git merge` only launches the editor in a tty.
+		// Until cucumber tests run git subcommands in a tty we have to explicitly set
+		// `--edit` mode to test commit message behaviour.
+		gitArgs = append(gitArgs, "--edit")
+	default:
+		return fmt.Errorf("unhandled %#v case", useMessage)
 	}
 	// Add branch name as the last argument.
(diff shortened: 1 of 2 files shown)


internal/git/commands.go
```diff
@@ -526,10 +526,17 @@ func (self *Commands) MergeFastForward(runner gitdomain.Runner, branch gitdomain
...
 // MergeNoFastForward merges branch into the current branch and always creates a merge commit.
-func (self *Commands) MergeNoFastForward(runner gitdomain.Runner, message Option[gitdomain.CommitMessage], branch gitdomain.LocalBranchName) error {
+func (self *Commands) MergeNoFastForward(runner gitdomain.Runner, message Option[gitdomain.CommitMessage], useDefaultMessage bool, branch gitdomain.LocalBranchName) error {
 	gitArgs := []string{"merge", "--no-ff"}
 	if messageContent, has := message.Get(); has {
 		gitArgs = append(gitArgs, "-m", messageContent.String())
+	} else if useDefaultMessage {
+		gitArgs = append(gitArgs, "--no-edit")
+	} else {
+		// Unlike `git commit` `git merge` only launches the editor in a tty.
```
stephenwade, line 536:
```suggestion
		// Unlike `git commit`, `git merge` only launches the editor in a tty.
```

legeana: Done, thanks!

kevgo (approved):
Thanks for implementing this! I completely agree that Git Town should invoke Git commands in a way that explicitly specifies the expected behavior. Besides the TTY issue highlighted in this ticket, this also ensures predictable behavior given the variety of configuration options that can influence Git's behavior. We had a number of bug reports where people had some exotic option set and then Git commands called by Git Town didn't have the documented behavior. Great idea!

internal/git/commands.go
```diff
@@ -526,10 +526,17 @@ func (self *Commands) MergeFastForward(runner gitdomain.Runner, branch gitdomain
 }
 
 // MergeNoFastForward merges branch into the current branch and always creates a merge commit.
-func (self *Commands) MergeNoFastForward(runner gitdomain.Runner, message Option[gitdomain.CommitMessage], branch gitdomain.LocalBranchName) error {
+func (self *Commands) MergeNoFastForward(runner gitdomain.Runner, message Option[gitdomain.CommitMessage], useDefaultMessage bool, branch gitdomain.LocalBranchName) error {
```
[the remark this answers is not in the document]
legeana: Done in https://github.com/git-town/git-town/pull/4451.

PR #4444, opened 2024-12-29
Add MockCommitMessage test

legeana:
While debugging confusing [`git merge
--edit`](https://github.com/git/git/blob/306ab352f4e98f6809ce52fc4e5d63fb947d0635/builtin/merge.c#L1082-L1106) behaviour I wasn't sure if `MockCommitMessage` works properly, so let's add a test.

### DIFF
--- a/test/subshell/test_runner_test.go
+++ b/test/subshell/test_runner_test.go
@@ -32,6 +32,20 @@ func TestMockingRunner(t *testing.T) {
 		must.EqOp(t, "foo called with: bar", res)
 	})
 
+	t.Run("MockCommitMessage", func(t *testing.T) {
+		t.Parallel()
+		runner := subshell.TestRunner{
+			WorkingDir: t.TempDir(),
+			HomeDir:    t.TempDir(),
+			BinDir:     filepath.Join(t.TempDir(), "bin"),
+		}
+
+		runner.MockCommitMessage("test commit message")
+
+		res := runner.MustQuery("bash", "-c", "$GIT_EDITOR output; cat output")
+		must.Eq(t, "test commit message", res)
+	})
+
 	t.Run("Run", func(t *testing.T) {
 		t.Parallel()
 		runner := subshell.TestRunner{


test/subshell/test_runner_test.go
```diff
@@ -32,6 +32,20 @@ func TestMockingRunner(t *testing.T) {
...
+		t.Parallel()
+		runner := subshell.TestRunner{
+			WorkingDir: t.TempDir(),
+			HomeDir:    t.TempDir(),
+			BinDir:     filepath.Join(t.TempDir(), "bin"),
+		}
+
+		runner.MockCommitMessage("test commit message")
+
+		res := runner.MustQuery("bash", "-c", "$GIT_EDITOR output; cat output")
```
kevgo, line 45: Rather than shelling out to Bash and then again to `cat`, could this test read the created file directly from Go, for example using `os.ReadFile`? This would also nicely document/test the expected path of the file that is supposed to exist.

test/subshell/test_runner_test.go
```diff
@@ -32,6 +32,20 @@ func TestMockingRunner(t *testing.T) {
 		must.EqOp(t, "foo called with: bar", res)
 	})
 
+	t.Run("MockCommitMessage", func(t *testing.T) {
+		t.Parallel()
+		runner := subshell.TestRunner{
+			WorkingDir: t.TempDir(),
+			HomeDir:    t.TempDir(),
+			BinDir:     filepath.Join(t.TempDir(), "bin"),
+		}
```
kevgo, line 41: You could save a few roundtrips to the filesystem during and then again after the test (to clean up the created temp dirs).

```go
dir := t.TempDir()
runner := subshell.TestRunner{
	WorkingDir: dir,
	HomeDir:    dir,
	BinDir:     filepath.Join(dir, "bin"),
}
```

test/subshell/test_runner_test.go
kevgo: What does the call `$GIT_EDITOR output` do here?

test/subshell/test_runner_test.go
legeana: Didn't realise `t.TempDir()` returns a new tempdir every time. Thanks!

test/subshell/test_runner_test.go
legeana: Done in #4453.

PR #4445, opened 2024-12-29
Add always-merge ship strategy

legeana:
Tests are a modified copy of fast-forward tests.

Currently undo only restores the feature branch, similar to fast-forward
strategy. There is no undo support for the parent branch like in
squash-merge.

Undoing merge commits generally should be avoided. Reverting a merge
creates a new revert commit that reverts the data, but the original
commits are still part of the history. As a result any subsequent merge
will skip commits that have been merged, even though their data has been
reverted, unless the user modified the old commits in some way, e.g. by
rebasing or amending them.

While this is unlikely to be a problem in modern workflows where users
often sync their branches, we should err on the side of not doing
something that can lead to confusion potentially far in the future.

See
https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt
for more information.

See #4381.

### DIFF
--- a/features/ship/always_merge/current_branch/unconfigured.feature
+++ b/features/ship/always_merge/current_branch/unconfigured.feature
@@ -1,0 +1,15 @@
+@messyoutput
+Feature: ask for missing configuration information
+
+  Scenario: unconfigured
+    Given a Git repo with origin
+    And Git Town is not configured
+    And Git setting "git-town.ship-strategy" is "always-merge"
+    When I run "git-town ship" and enter into the dialog:
+      | DIALOG      | KEYS  |
+      | main branch | enter |
+    And the main branch is now "main"
+    And Git Town prints the error:
+      """
+      cannot ship the main branch
+      """
--- a/features/config/setup/change_git_metadata.feature
+++ b/features/config/setup/change_git_metadata.feature
@@ -32,7 +32,7 @@ Feature: change existing information in Git metadata
       | enable push-new-branches                  | down enter             |
       | disable the push hook                     | down enter             |
       | new-branch-type                           | down enter             |
-      | set ship-strategy to "fast-forward"       | down enter             |
+      | set ship-strategy to "fast-forward"       | down down enter        |
       | disable ship-delete-tracking-branch       | down enter             |
       | save config to Git metadata               | down enter             |
 
--- a/features/ship/always_merge/current_branch/additional_commits/before_undo/offline.feature
+++ b/features/ship/always_merge/current_branch/additional_commits/before_undo/offline.feature
@@ -1,0 +1,38 @@
+Feature: partially undo an offline ship using the always-merge strategy after additional commits to main
+
+  Background:
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the commits
+      | BRANCH  | LOCATION      | MESSAGE        |
+      | feature | local, origin | feature commit |
+    And the current branch is "feature"
+    And offline mode is enabled
+    And Git setting "git-town.ship-strategy" is "always-merge"
+    When I run "git-town ship" and close the editor
+    And I add commit "additional commit" to the "main" branch
+
+  Scenario: undo
+    When I run "git-town undo"
+    Then Git Town runs the commands
+      | BRANCH | COMMAND                                       |
+      | main   | git branch feature {{ sha 'feature commit' }} |
+      |        | git checkout feature                          |
+    And Git Town prints:
+      """
+      cannot reset branch "main"
+      """
+    And Git Town prints:
+      """
+      it received additional commits in the meantime
+      """
+    And the current branch is now "feature"
+    And these commits exist now
+      | BRANCH  | LOCATION | MESSAGE                |
+      | main    | local    | feature commit         |
+      |         |          | Merge branch 'feature' |
+      |         |          | additional commit      |
+      | feature | origin   | feature commit         |
+    And the initial branches and lineage exist now
--- a/features/ship/always_merge/current_branch/additional_commits/before_undo/online.feature
+++ b/features/ship/always_merge/current_branch/additional_commits/before_undo/online.feature
@@ -1,0 +1,29 @@
+Feature: partially undo an online ship using the always-merge strategy after additional commits to the main branch
+
+  Background:
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the commits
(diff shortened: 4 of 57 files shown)


features/ship/always_merge/current_branch/unconfigured.feature
legeana, line 1: This test is weird, need to look more into it and sister tests.

kevgo:
Since it came up, here is an overview of how the undo system works. I'm happy to provide more background if needed, just let me know.

- each Git Town command captures a snapshot of the Git repo before and after running
- snapshots consist of all local and remote Git branches and their SHAs, all local and global Git configuration, and the Git stash size
- both snapshots get stored in the persistent runstate on disk for the command
- when the user runs `git town undo`: Git Town calculates the diff between the stored before and after snapshot and executes the Git commands to reset the Git repo to the before state
  - reset non-perennial branches: `git reset --hard <before-SHA>`, then `git push --force`
  - reset perennial branches: `get revert <after-SHA>`, then `git push`
- plenty of checks to ensure the reset is safe:
  - only reverts branches that the command actually changed
  - make sure no other commits that were created separately get reverted

features/ship/always_merge/current_branch/unconfigured.feature
legeana: Oh, never mind, it queries the main branch, and the test verifies that the dialogue happens.

legeana:
This is now blocked on #4443.

features/ship/always_merge/current_branch/unconfigured.feature
kevgo: The purpose of this test is to verify a safeguard against accidentally forgetting to commit changes that should be included in the shipped branch. Here's a potential scenario: while preparing a branch for shipping, I make a few last-minute adjustments. After verifying that all tests pass, I run `git town ship`. In this situation, the command should refuse to proceed because the uncommitted changes likely belong to the current branch and should be committed before shipping.

legeana:
> Since it came up, here is an overview of how the undo system works. I'm happy to provide more background if needed, just let me know.
> 
> * each Git Town command captures a snapshot of the Git repo before and after running
> * snapshots consist of all local and remote Git branches and their SHAs, all local and global Git configuration, and the Git stash size
> * both snapshots get stored in the persistent runstate on disk for the command
> * when the user runs `git town undo`: Git Town calculates the diff between the stored before and after snapshot and executes the Git commands to reset the Git repo to the before state
>   
>   * reset non-perennial branches: `git reset --hard <before-SHA>`, then `git push --force`
>   * reset perennial branches: `get revert <after-SHA>`, then `git push`
> * plenty of checks to ensure the reset is safe:
>   
>   * only reverts branches that the command actually changed
>   * make sure no other commits that were created separately get reverted

Thanks! This matches my understanding. My biggest concern with merge reversals is that it isn't recommended, so I don't think we should do it. But I am open to other opinions.

features/ship/always_merge/current_branch/unconfigured.feature
legeana: Are you sure you are talking about this test case? I am pretty sure the feature under test here is the dialogue:

```
    When I run "git-town ship" and enter into the dialog:
      | DIALOG      | KEYS  |
      | main branch | enter |
```

kevgo:
Yes, this feature introduces a new situation: undoing a merge commit created by `git town ship` on a perennial branch, where we cannot reset the branch to an earlier commit. Thanks for the link explaining the problem!

The prime directive for Git Town is to never lose data, and never leave a user's repo in an inconsistent state, even in unusual circumstances. If there is risk that something does not work, then it's better to not do anything and let the user sort out the problem.

So, the right thing to do here seems to not change the merge commit on the perennial branch and tell the user why.

features/ship/always_merge/current_branch/unconfigured.feature
kevgo: Apologies, I was talking about at the test case above this one 🤣 

You're right, this test ensures that `git town ship` doesn't ship the main branch. In this scenario the tool starts up without knowing what the main branch is, then identifies it later based on user input during the configuration dialog.

legeana:
> So, the right thing to do here seems to not change the merge commit on the perennial branch and tell the user why.

If I understand you correctly you are proposing to add a warning?

kevgo:
Yes. To me the right thing to do here seems: don't revert the merge commit, and notify the user that the merge commit wasn't reverted, and why. How does that sound to you? I'm open to other ideas.

legeana:
> Yes. To me the right thing to do here seems: don't revert the merge commit, and notify the user that the merge commit wasn't reverted, and why. How does that sound to you? I'm open to other ideas.

I have an idea on how to implement this: similar to `RunState.UndoablePerennialCommits` we can add `RunState.UnsafeToUndoPerennialMergeCommits`. Then I can register the merge commit in the `MergeAlwaysProgram` and print warnings if `UnsafeToUndoPerennialMergeCommits` are present in the `UndoProgram`.

@kevgo WDYT?

kevgo:
This approach could work, but I have some concerns about the added complexity. The warning addresses a relatively rare use case, and it might not justify the overhead introduced by implementing it.

It might be better to omit the warning for non-undoable merge commits from this PR. My earlier comments were more about brainstorming the ideal user experience, but I'm perfectly okay with taking small, incremental steps toward that goal. We can revisit the idea of displaying a warning in a future PR to evaluate whether the additional complexity is worth it.

Regarding the current behavior: I believe `git town undo` already handles things correctly by not attempting to undo merge commits on perennial branches. To make `git town undo` revert such commits, they would need to be registered as undoable at the time of their creation. Currently, this happens only [when using squash-merge](https://github.com/git-town/git-town/blob/516cfb874202b22305ad079f2fbc22ee413942bb/internal/vm/opcodes/merge_squash_program.go#L52-L54), because it's clear at that point that the commit isn't a merge commit.

For the always-merge strategy, your implementation should simply avoid registering the merge commit it creates as undoable.

If we decide to implement the warning message later, it could be handled when compiling the undo program. The logic for this includes a [condition](https://github.com/git-town/git-town/blob/516cfb874202b22305ad079f2fbc22ee413942bb/internal/undo/undobranches/branch_changes.go#L64) that determines whether a commit on a perennial branch is undoable. If a commit is found to be non-undoable, we could append a warning message to the [messages displayed at the end of the program](https://github.com/git-town/git-town/blob/ddcd0c67dbb1e185ac866ace134d2692dd47a683/internal/cli/print/footer.go#L17-L19). The benefit of this approach is that it would cover all non-undoable commits on perennial branches.

kevgo:
I have just shipped infrastructure that notifies the user about all non-undoable commits on perennial branches: https://github.com/git-town/git-town/pull/4455. Hopefully this solves the notification problem and this branch can ship faster.

legeana:
@kevgo this is ready for review.

kevgo (approved):
Very well done!

PR #4446, opened 2024-12-29
Remove "The best way to change this setting"

stephenwade:
This PR removes all lines that say "The best way to change this setting is via the setup assistant" in the preferences documentation.

I disagree that the best way to change settings is via the setup assistant. When I want to update a setting, I don't want my config file rewritten with all the defaults explicitly stated and my comments removed.

Probably the best way to create a config file is using the setup assistant, but not to update one.

I will wait for review to merge this PR since others may disagree.


kevgo:
Good catch, I agree with your take on this. I was probably getting a bit carried away when I created the setup assistant 😄

PR #4447, opened 2024-12-30
Add `the uncommitted file still exists` checks to undo scenarios

legeana:
If an uncommitted file is created, we may as well test that it isn't lost during undo.


stephenwade:
Good idea, thanks!

PR #4448, opened 2024-12-30
Add `Git Town runs no commands` for `nothing to undo` scenarios

legeana:
And order then consistently [`commands check`, `message check`]. Commands check is more important.

```
    Then Git Town runs no commands
    And Git Town prints:
       """
       nothing to undo
       """
```


kevgo:
I like this order!

PR #4450, opened 2024-12-30
Document configuration hierarchy

kevgo:
Implements action item in https://github.com/git-town/git-town/pull/4425#issuecomment-2561998380.

### DIFF
--- a/website/src/preferences.md
+++ b/website/src/preferences.md
@@ -1,9 +1,27 @@
 # Preferences
 
 You can see all preferences via the [config](commands/config.md) command and
-change them via the [setup assistant](commands/config-setup.md).
+change them via the [setup assistant](commands/config-setup.md) or manually.
+Configuration data exists on multiple levels:
 
-Git Town can store preferences in two places:
+1. Team-wide configuration settings go into the
+   [configuration file](configuration-file.md). These settings apply to all Git
+   Town users working on the respective repository.
+2. Each developer can configure their preferred Git Town settings for all
+   repositories on their machine using global Git metadata. These settings
+   override (1). For example, if I always want to use the `rebase`
+   [sync-feature-strategy](https://www.git-town.com/preferences/sync-feature-strategy.html)
+   in all my repositories, I would run:
 
-- [configuration file](configuration-file.md)
-- Git metadata: as entries in the local or global Git configuration
+   ```command-summary
+   git config --global git-town.sync-feature-strategy rebase.
+   ```
+3. User and repo specific configuration settings go into local Git metadata,
+   which takes precedence over (1) and (2). For example, if I want `rebase` as
+   the default strategy for all my repositories, except in the `foo` repo I want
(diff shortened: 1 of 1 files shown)


website/src/preferences.md
```diff
@@ -1,9 +1,27 @@
...
+   repositories on their machine using global Git metadata. These settings
+   override (1). For example, if I always want to use the `rebase`
+   [sync-feature-strategy](https://www.git-town.com/preferences/sync-feature-strategy.html)
+   in all my repositories, I would run:
 
-- [configuration file](configuration-file.md)
-- Git metadata: as entries in the local or global Git configuration
+   ```command-summary
+   git config --global git-town.sync-feature-strategy rebase
+   ```
```
kevgo, line 18: @stephenwade, this new type of code block doesn't wrap like the others, and it seems to be due to the absence of square brackets. Would there be a different identifier—other than `command-summary`—that could be applied here to make this example Git call wrap similarly to the command summaries? This exceeds my CSS-Fu.

image

stephenwade: > this new type of code block doesn't wrap like the others, and it seems to be due to the absence of square brackets.

That's correct, "command-summary" is a single-purpose block for the summary at the top of command pages, and it starts wrapping at the first `[` or `<`.

Suggestion: We could add a similar type of code block with a more generic name like "wrap" and wrap every word (separated by spaces), instead of trying to match a command name. On a narrow screen, this code block would wrap like:

```
git config --global
    git-town.sync-feature-strategy
    rebase
```

Would this work the way you want?

kevgo: You definitely have a stronger aesthetic sense than I do, so I trust your judgment on this! Your suggestion sounds great.

Here's a dopey idea—no worries if it’s not feasible or would add unnecessary complexity. It’s also perfectly fine to start with a simple implementation and revisit this idea later if it makes sense. Would it be possible to make this look like a shell command split across multiple lines using backslashes? For example:

```bash
git config --global \
    git-town.sync-feature-strategy \
    rebase
```

Ideally, when someone copies and pastes this, it would paste as a single line without the backslashes or line breaks. Just a suggestion—feel free to disregard if it's not practical or would break accessibility!

stephenwade: I looked into this and found it's not possible with CSS - we could target the first line but not any lines by position (I&nbsp;would want to use CSS to add the `\` to all but the last line, so it is not included in copy-paste). So the solution would have to involve JavaScript that runs whenever the page is resized.

Detecting element widths and adding content in the right place is somewhat complicated, and I don't think it is worth it to achieve the backslashes. The code block will copy and paste on one line, since the wrapping is achieved with CSS `text-wrap` instead of any actual line breaks, so I think it is good enough without adding complexity.

kevgo: Thanks for looking into this! As I said, it was a dopey idea :smile: Let's go with your idea.

stephenwade: Implemented in #4474.

PR #4451, opened 2024-12-30
Add configdomain.UseMessage

legeana:
UseMessage is a variant type that supports the following options:
- `EditDefaultMessage`
- `UseCustomMessage(CommitMessage)`
- `UseDefaultMessage`

As Go doesn't support algebraic enums we can instead use helper functions `Is<Type>()`. Compared to Go style const enums we lose the linter's exhaustive check, so we have to use a default clause instead.

One of the advantages of this approach is we don't have to think about `useDefaultMessage` value when we explicitly set a custom commit message, e.g. in `test/cucumber/steps.go`. That value would be unused anyway.

Alternatives considered:
1. `configdomain.UseDefaultMessage(bool)`. This name is confusing when used together with a custom message. For example `Commit(message, UseDefaultMessage(true))` is ambiguous whether a default message or a custom message is used.
2. `configdomain.EditCommitMessage(bool)`. This separates default message and editing. But currently we only care about editing default messages, so the caller needs to handle more complexity.
3. `configdomain.EditCommitMessage(int)`. This allows us to declare custom behaviours such as `EditDefaultMessage`, `NoEditDefaultMessage` or even `AlwaysEditMessage`, `EditCustomMessage` if they become necessary. However negative `NoEditDefaultMessage` doesn't read well, and if we just call it `UseDefaultMessage` we go back to (1): `Commit(message, UseDefaultMessage)`: do we use the message or the default message?

Looking at the issue at hand we always care about the combination of the message and default message editing behaviour. We only care about the default message editing if the message itself is None. So it is much simpler to think of these two arguments as if it is a composite type, hence UseMessage.

See https://github.com/git-town/git-town/pull/4443#discussion_r1899236258.

### DIFF
--- a/internal/config/configdomain/use_message.go
+++ b/internal/config/configdomain/use_message.go
@@ -1,0 +1,98 @@
+package configdomain
+
+import (
+	"github.com/git-town/git-town/v17/internal/git/gitdomain"
+	. "github.com/git-town/git-town/v17/pkg/prelude"
+)
+
+// UseMessage indicates how to set commit message.
+// The following variants are supported:
+// - EditDefaultMessage
+// - UseCustomMessage(CommitMessage)
+// - UseDefaultMessage
+//
+// See `--message`, `--edit` and `--no-edit` flags in `git commit` and `git merge`.
+type UseMessage struct {
+	customMessage      Option[gitdomain.CommitMessage]
+	editDefaultMessage bool
+}
+
+// EditDefaultMessage indicates that the default message should be edited.
+func EditDefaultMessage() UseMessage {
+	return UseMessage{
+		customMessage:      None[gitdomain.CommitMessage](),
+		editDefaultMessage: true,
+	}
+}
+
+// UseCustomMessage indicates that the custom message should be used as is without opening an editor.
+func UseCustomMessage(message gitdomain.CommitMessage) UseMessage {
+	return UseMessage{
+		customMessage:      Some(message),
+		editDefaultMessage: false, // This value is unused.
+	}
+}
+
+// UseCustomMessageOr returns UseCustomMessage(message) if message is Some, or other.
+func UseCustomMessageOr(message Option[gitdomain.CommitMessage], other UseMessage) UseMessage {
+	if m, has := message.Get(); has {
+		return UseCustomMessage(m)
+	}
+	return other
+}
+
+// UseDefaultMessage indicates that the default message should be used as is without opening an editor.
+func UseDefaultMessage() UseMessage {
+	return UseMessage{
+		customMessage:      None[gitdomain.CommitMessage](),
+		editDefaultMessage: false,
+	}
+}
+
+// UseMessageWithFallbackToDefault returns UseMessage based on an optional message and fallbackToDefault.
+// - If message is Some(...) then this is equivalent to UseCustomMessage(...).
+// - If message is None and fallbackToDefault then this is equivalent to UseDefaultMessage().
+// - If message is None and !fallbackToDefault then this is equivalent to EditDefaultMessage().
+func UseMessageWithFallbackToDefault(message Option[gitdomain.CommitMessage], fallbackToDefault bool) UseMessage {
+	return UseMessage{
+		customMessage:      message,
+		editDefaultMessage: !fallbackToDefault,
+	}
+}
+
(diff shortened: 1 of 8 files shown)


kevgo:
Thanks for implementing this, I think it makes the intent of the code a lot more clear, and I agree with the direction you are taking here.

It's unfortunate that modeling this type of data is so cumbersome in Go, but at least this complexity is contained. I think Go would be a much more ergonomic and capable, yet still simple language if it had Rust-style enums, Rust's `?` operator for error handling, and Rust-style pattern matching.

Btw, I don't seem to get notifications when the `draft` status of pull requests gets removed, so please comment when this branch is ready for review. Looking forward to merging this in!

legeana:
> It's unfortunate that modeling this type of data is so cumbersome in Go, but at least this complexity is contained. I think Go would be a much more ergonomic and capable, yet still simple language if it had Rust-style enums, Rust's `?` operator for error handling, and Rust-style pattern matching.

That is my main issue with Go as well to be honest.

> Btw, I don't seem to get notifications when the draft status of pull requests gets removed, so please comment when this branch is ready for review. Looking forward to merging this in!

Sounds good, I will just leave an explicit message when the code is ready for review in the future.

legeana:
@kevgo this is now ready for review.

The outstanding question is what is your policy on dead code? Currently there are 3 functions that are unreachable:
- `EditDefaultMessage` - unused, but logically is part of the enum interface. Removing it might be confusing.
- `UseCustomMessageOr` is used in #4445.
- `UseDefaultMessage` is used in #4443.

internal/config/configdomain/use_message.go
```diff
@@ -0,0 +1,98 @@
...
+// - If message is None and !fallbackToDefault then this is equivalent to EditDefaultMessage().
+func UseMessageWithFallbackToDefault(message Option[gitdomain.CommitMessage], fallbackToDefault bool) UseMessage {
+	return UseMessage{
+		customMessage:      message,
+		editDefaultMessage: !fallbackToDefault,
+	}
+}
+
+// GetCustomMessage returns a copy of the custom message and IsCustomMessage().
+func (self *UseMessage) GetCustomMessage() (gitdomain.CommitMessage, bool) {
```
kevgo, line 64: Not feeling strongly about this, but am curious whether this method could just return `Option[gitdomain.CommitMessage]`?

internal/config/configdomain/use_message_test.go
```diff
@@ -0,0 +1,149 @@
...
+		must.True(t, isCustom)
+		must.Eq(t, want, message)
+		must.Eq(t, want, m.GetCustomMessageOrPanic())
+		// Type checks.
+		must.True(t, m.IsCustomMessage())
+		must.False(t, m.IsEditDefault())
+		must.False(t, m.IsUseDefault())
+	})
+
+	t.Run("UseCustomMessageOrWithSome", func(t *testing.T) {
```
kevgo, line 61: There seem to be multiple tests that verify `configdomain.UseCustomMessageOr`. How do you feel about moving them into a nested `t.Run` block?

```go
t.Run("UseCustomMessageOr", func(t *testing.T) {
  t.Run("with some", func(t *testing.T) { ... })
  t.Run("with none", func(t *testing.T) { ... })  
})
```

What I like about this setup is that it helps get an overview of the various test cases, and spot missing ones.

kevgo (approved):
Thanks for modeling the data so close to the domain! You can disable false deadcode positives in the [Makefile](https://github.com/git-town/git-town/blob/516cfb874202b22305ad079f2fbc22ee413942bb/Makefile#L131).

internal/config/configdomain/use_message.go
legeana: While this looks similar, I disagree. If we just return the inner `customMessage` this starts to look like a direct access to the private state.

Instead the boolean here indicates whether the UseMessage instance is CustomMessage or not. And if it is logically it contains non-optional CommitMessage.

internal/config/configdomain/use_message_test.go
legeana: Done.

internal/config/configdomain/use_message.go
kevgo: Okay, thanks for the background. Makes sense.

kevgo:
Thanks for implementing this so well!

PR #4452, opened 2024-12-30
Fix reliance on order in MergeNoFastForward unit tests

legeana:
Use SliceContainsAll instead of Eq.


kevgo:
Nice, I was getting flakiness around this test case as well on my machine. Surprising that the order in which Git returns commits has such a variance here. Great this is fixed now!

PR #4453, opened 2024-12-30
Improve MockCommitMessage test

legeana:
- Cache `t.TempDir()` at the start of the test. We don't need more than
  one TempDir per test.
- Use `os.ReadFile()` to read the file instead of `cat`.
- Add a comment explaining what the assert section of the test is
  checking.

This commit is a response to feedback in #4444.

### DIFF
--- a/test/subshell/test_runner_test.go
+++ b/test/subshell/test_runner_test.go
@@ -34,16 +34,26 @@ func TestMockingRunner(t *testing.T) {
 
 	t.Run("MockCommitMessage", func(t *testing.T) {
 		t.Parallel()
+		dir := t.TempDir()
 		runner := subshell.TestRunner{
-			WorkingDir: t.TempDir(),
-			HomeDir:    t.TempDir(),
-			BinDir:     filepath.Join(t.TempDir(), "bin"),
+			WorkingDir: dir,
+			HomeDir:    dir,
+			BinDir:     filepath.Join(dir, "bin"),
 		}
 
 		runner.MockCommitMessage("test commit message")
 
-		res := runner.MustQuery("bash", "-c", "$GIT_EDITOR output; cat output")
-		must.Eq(t, "test commit message", res)
+		// Simulate Git calling the mock editor configured by MockCommitMessage and
+		// verify its effect.
+		// MockCommitMessage creates a custom editor that the runner makes available
+		// via the GIT_EDITOR environment variable. We verify the following:
+		// - GIT_EDITOR is available and executable.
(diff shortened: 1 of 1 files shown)


legeana:
@kevgo please take a look.

test/subshell/test_runner_test.go
```diff
@@ -34,16 +34,20 @@ func TestMockingRunner(t *testing.T) {
...
+			WorkingDir: dir,
+			HomeDir:    dir,
+			BinDir:     filepath.Join(dir, "bin"),
 		}
 
 		runner.MockCommitMessage("test commit message")
 
-		res := runner.MustQuery("bash", "-c", "$GIT_EDITOR output; cat output")
-		must.Eq(t, "test commit message", res)
+		// Verify that `$GIT_EDITOR file` sets the file's contents to the given commit message.
```
kevgo, line 46: This is a pretty clever test because it bypasses Git altogether! :+1: A comment like this would help people like me follow what it does:

```suggestion
		// Simulate Git calling the executable configured in the $GIT_EDITOR environment variable when it needs to display an editor to the user.
```

kevgo (approved):
Thanks, I now get what this test does. Well done!

test/subshell/test_runner_test.go
legeana: I propose a slightly more verbose comment, please take another look.

test/subshell/test_runner_test.go
```diff
@@ -34,16 +34,25 @@ func TestMockingRunner(t *testing.T) {
...
+			BinDir:     filepath.Join(dir, "bin"),
 		}
 
 		runner.MockCommitMessage("test commit message")
 
-		res := runner.MustQuery("bash", "-c", "$GIT_EDITOR output; cat output")
-		must.Eq(t, "test commit message", res)
+		// Simulate Git calling the mock editor configured by MockCommitMessage and
+		// verify its effect.
+		// MockCommitMessage sets the GIT_EDITOR environment variable to force Git to
```
kevgo, line 48: Technically `MockCommitMessage` doesn't set this environment variable. It sets the `gitEditor` property on its `TestRunner` instance. `TestRunner.QueryWithCode` then sets the environment variable in the subshell it calls into, if the property is set.

kevgo (commented):
I'm fine with whatever you want to write here, just making suggestions. :slightly_smiling_face:

kevgo (approved):
Ship it! :rocket:

PR #4455, opened 2024-12-30
Notify the user about all non-undoable commits on perennial branches

kevgo:
Inspired by #4445


internal/undo/undobranches/branch_changes.go
```diff
@@ -65,6 +68,8 @@ func (self BranchChanges) UndoProgram(args BranchChangesUndoProgramArgs) program
 			result.Add(&opcodes.CheckoutIfNeeded{Branch: branch})
 			result.Add(&opcodes.CommitRevertIfNeeded{SHA: change.After})
 			result.Add(&opcodes.PushCurrentBranchIfNeeded{CurrentBranch: branch})
+		} else {
+			args.FinalMessages.Add(fmt.Sprintf(messages.UndoCannotRevertCommitOnPerennialBranch, change.After))
```
legeana, line 72: This is awesome, thank you!

I tested that this works on #4445 but I couldn't add cucumber tests yet (see https://github.com/legeana/git-town/pull/1) because I didn't find a way to parameterise the messages. I understand `{{ sha ... }}` is implemented in datatables only and we may need to do something else here if we want to test this. On the other hand we can just verify the end state, and this is good enough.

kevgo: For now you could use the step `Then Git Town prints something like:` with a regex. Example: https://github.com/git-town/git-town/blob/8a59dfef86cd95cdbe1f797927e3f5a152ebfcf8/features/status/status.feature#L27-L33

PR #4464, opened 2025-01-01
Add undo warning test for the always-merge shipping strategy

legeana:
See #4381 and #4455.

### DIFF
--- a/features/ship/always_merge/current_branch/undo_warning.feature
+++ b/features/ship/always_merge/current_branch/undo_warning.feature
@@ -1,0 +1,22 @@
+Feature: git-town undo prints a warning message for a merge commit
+
+  Background:
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the current branch is "feature"
+    And the commits
+      | BRANCH  | LOCATION      | MESSAGE        |
+      | feature | local, origin | feature commit |
+    And Git setting "git-town.ship-strategy" is "always-merge"
+    # "I ran" verifies the exit status.
+    And I ran "git-town ship -m 'feature done'"
+
+  Scenario: undo
+    When I run "git-town undo"
+    Then Git Town prints something like:
+      # ".*" is the merge commit created by `git town ship`.
+      """
+      Cannot undo commit ".*" because it is on a perennial branch
+      """


features/ship/always_merge/current_branch/undo_warning.feature
```diff
@@ -0,0 +1,22 @@
...
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the current branch is "feature"
+    And the commits
+      | BRANCH  | LOCATION      | MESSAGE        |
+      | feature | local, origin | feature commit |
+    And Git setting "git-town.ship-strategy" is "always-merge"
+    When I run "git-town ship" and close the editor
+    Then the current branch is now "main"
```
kevgo, line 14: What is the purpose of this step in the context of this test, which verifies the message printed to the user?

kevgo (approved):
LGTM. I like how this test verifies just one aspect (the message printed to the user). Might be a way to get a handle on the combinatorial explosion that the end-to-end test suite starts running into. The current end-to-end tests try to describe the complete user experience, which combines more and more features with each other, leading to quite a bit of redundancy.

features/ship/always_merge/current_branch/undo_warning.feature
legeana: Are you asking about the verification that we are on the main branch? This is a very simple way to make sure the command actually executed. This verifies that the test makes sense.

If the command hasn't executed, then the developer may incorrectly focus on debugging nonsense output from `undo` instead of focusing on why `git-town ship` didn't work. I added a comment for clarification, PTAL.

legeana:
> I like how this test verifies just one aspect (the message printed to the user).

That's typically the preferred way of writing tests, but without a good overall understanding of all the tests and their reasoning it would be hard to rework the entire cucumber test suite.

I absolutely agree with you on the point that this **is** the way to reduce the test suite size. However for integration tests there is a value in testing all possible combinations to avoid incorrect corner case handling, so we do need to find the right balance as well.

features/ship/always_merge/current_branch/undo_warning.feature
```diff
@@ -0,0 +1,25 @@
...
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the current branch is "feature"
+    And the commits
+      | BRANCH  | LOCATION      | MESSAGE        |
+      | feature | local, origin | feature commit |
+    And Git setting "git-town.ship-strategy" is "always-merge"
+    When I run "git-town ship" and close the editor
```
kevgo, line 13: Nitpick: the past tense helps signal that this is not the `act` phase.

```suggestion
    When I ran "git-town ship" and close the editor
```

features/ship/always_merge/current_branch/undo_warning.feature
kevgo: Yes, I was wondering what the step to verify the current branch does here.

This gives me an idea: the `I ran` step should always fail if the exit code is non-zero. I'll implement this in a separate PR.

kevgo:
> > I like how this test verifies just one aspect (the message printed to the user).
> 
> That's typically the preferred way of writing tests, but without a good overall understanding of all the tests and their reasoning it would be hard to rework the entire cucumber test suite.
> 
> I absolutely agree with you on the point that this **is** the way to reduce the test suite size. However for integration tests there is a value in testing all possible combinations to avoid incorrect corner case handling, so we do need to find the right balance as well.

Thanks for the inspiration and guidance! I have created https://github.com/git-town/git-town/issues/4466 to track this idea.

features/ship/always_merge/current_branch/undo_warning.feature
```diff
@@ -0,0 +1,25 @@
...
+    # Make sure that "git-town ship" actually worked.
+    # If this fails then the debugging should focus on Background instead of
+    # Scenarios below.
+    Then the current branch is now "main"
```
kevgo, line 17: I just merged #4468, which updates the behavior of `I ran` steps to fail if the executed command exits with a non-zero status. Since Git Town reliably signals failure using exit statuses, this enhancement should make it safe to remove this step and comment while maintaining confidence that the preceding command executed successfully and completely. Let me know if you have any concerns or edge cases where this might not apply!

```suggestion
```

legeana: `and close the editor` is not currently supported, but it isn't hard to rewrite it to `-m "merge message"` instead.

kevgo: I wanted to suggest that anyways since it executes faster.

legeana: Done. PTAL.

features/ship/always_merge/current_branch/undo_warning.feature
```diff
@@ -0,0 +1,22 @@
...
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the current branch is "feature"
+    And the commits
+      | BRANCH  | LOCATION      | MESSAGE        |
+      | feature | local, origin | feature commit |
+    And Git setting "git-town.ship-strategy" is "always-merge"
+    # "I ran" verifies the exit status.
+    When I ran "git-town ship -m 'feature done'"
```
kevgo, line 14: I think the `I ran` step is part of the `Arrange` phase. In Gherkin (the language used in these Cucumber-based end-to-end tests) the `When` keyword is for steps in the `Act` phase.

```suggestion
    And I ran "git-town ship -m 'feature done'"
```

kevgo (approved):
This is good to merge in from my perspective. Thanks for adding this test!

features/ship/always_merge/current_branch/undo_warning.feature
legeana: Done.

PR #4465, opened 2025-01-01
Document merge reversal for always-merge ship strategy

legeana:
See #4381.

### DIFF
--- a/website/src/preferences/ship-strategy.md
+++ b/website/src/preferences/ship-strategy.md
@@ -24,6 +24,12 @@ The `always-merge` ship strategy creates a merge commit via `git merge --no-ff`.
 This strategy allows visually grouping related feature commits together which
 may aid in understanding project history in certain situations.
 
+It is not generally recommended to revert merge commits, so `git town undo` will
+not create a merge reversal commit if the merge commit has been pushed already.
+See
+[howto/revert-a-faulty-merge.txt](https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt)
+in the official Git documentation for more information.
+
 ### fast-forward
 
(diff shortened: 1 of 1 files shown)


website/src/preferences/ship-strategy.md
```diff
@@ -24,6 +24,12 @@ The `always-merge` ship strategy creates a merge commit via `git merge --no-ff`.
 This strategy allows visually grouping related feature commits together which
 may aid in understanding project history in certain situations.
 
+It is not generally recommended to revert merge commits, so `git town undo` will
+not create a merge reversal commit if the merge commit has been pushed already.
+See
+[howto/revert-a-faulty-merge.txt](https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt)
```
kevgo, line 30: How do you feel about also adding this link to the warning message printed on the CLI?

kevgo (approved):
Great idea to add this link to the documentation!

website/src/preferences/ship-strategy.md
legeana: I am not familiar with the undo code enough to figure out the right way to do that.

I see the following places where `UndoCannotRevertCommitOnPerennialBranch` is logged:
- [1](https://github.com/git-town/git-town/blob/main/internal/undo/undobranches/branch_changes.go#L72)
- [2](https://github.com/git-town/git-town/blob/main/internal/undo/undobranches/branch_changes.go#L102)

At no point it is clear that this is related to a merge commit. Are you suggesting to check a commit type around (1) and (2)?

kevgo: Great point! You're absolutely right, Git Town cannot determine whether a commit it can't undo is a merge commit or something else. Please disregard my earlier suggestion.

PR #4469, opened 2025-01-01
git-town-status-reset: Fix wrong path

kevgo:
This command was using a path shortcut that doesn't seem to work.


kevgo:

PR #4474, opened 2025-01-02
New "wrap" code block type on website

stephenwade:
600 px page:

500 px page:

400 px page:


kevgo:
This looks great, and thanks for updating all the other code blocks as well!

PR #4475, opened 2025-01-02
Improve preprocessor code on website

stephenwade:
Improve comments, simplify markup a bit, and make a tokenizer instead of doing it all with regular expressions.

### DIFF
--- a/tools/mdbook_preprocessor/mdbook_preprocessor.js
+++ b/tools/mdbook_preprocessor/mdbook_preprocessor.js
@@ -144,34 +144,83 @@ function processCommandSummary(code) {
     code
       .split("\n")
       .map(line => {
-        /**
-         * This regex matches the command in the line. The command is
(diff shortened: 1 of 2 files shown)


kevgo (commented):
This is a solid improvement, the code is now more readable and its intent more clear!

Some of the logic now reaches levels of complexity that make it non-trivial to reason about whether it behaves correct in all edge cases. How do you feel about adding unit tests for the `tokenize` and `extractCommand` functions? In my experience the time to set this up pays for itself within hours, because debugging, developing confidence in the code, and maintenance become so much easier. 

I don’t see an urgent requirement for tests at this stage, but I do think the need for them is starting to emerge.

If you’d like, I’d be happy to help set up an initial test suite or guide you through the process. Let me know if this is something you'd prefer to tackle now or address down the road!

stephenwade:
Unit tests for these utility functions are implemented in #4487.

PR #4478, opened 2025-01-02
Factor UNIT_TEST_DIRS out of Makefile unit, unit-all and unit-race

legeana:
unit-race specifically was now out of sync with the rest.


kevgo:
Thanks, this is a nice improvement, especially given that there will me more linters in the future!

PR #4479, opened 2025-01-02
Stash changes away with commit message

kevgo:
Stashing open changes away with a commit message allows users to identify Git stash entries created by Git Town.


kevgo:

PR #4480, opened 2025-01-02
Pop stash even if encountering conflicts

kevgo:
When the uncommitted changes that got stashed away conflicts with the branch that Git Town ends on, it leaves the stashed away changes on the stash. Technically the user should run `git town continue` to try to pop the stash again and finish the command. In practice, this is super easy to forget, and when using Git Town for a while the Git stash accumulates entries. This PR changes the behavior of Git Town commands so that they never leave the entry they created on the stash.

Following the principle of least surprise, this improves the ergonomics of Git Town commands.


kevgo:

PR #4486, opened 2025-01-03
Update Node to v22

stephenwade:
Node >= v21 is required to use glob syntax with the Node test runner. See #4487.


kevgo:
Nice, all the other tooling still works with Node 22!

PR #4491, opened 2025-01-03
tools/tests_sorted: add isTestFunc() and tRunSubtestNames()

legeana:
Both will be useful for implementing tools/tests_sorted linter.

Also return Extract error as a separate return value instead of mixing it up with a `matcher.Result`. These are two different types covering two distinct situations: code error vs not-a-match.


kevgo:
Thanks, good progress!

PR #4493, opened 2025-01-04
Update spelling of Bitbucket Data Center

stephenwade:
The product name is "Data Center", not "Datacenter".

See: https://www.atlassian.com/enterprise/data-center/bitbucket

~~@kevgo Ideally we would allow `bitbucket-data-center` as a valid hosting platform value, but I don't know how to do it without breaking old configs. Have we ever deprecated a specific config value before? Maybe we should just change it and special case `bitbucket-datacenter` to also parse to `bitbucket-data-center`.~~

### DIFF
--- a/features/propose/edge_cases/unsupported_hosting_service.feature
+++ b/features/propose/edge_cases/unsupported_hosting_service.feature
@@ -15,7 +15,7 @@ Feature: unsupported hosting platform
 
       This command requires hosting on one of these services:
       * Bitbucket
-      * Bitbucket datacenter
+      * Bitbucket Data Center
       * GitHub
       * GitLab
       * Gitea
--- a/features/repo/edge_cases/unsupported_hosting_service.feature
+++ b/features/repo/edge_cases/unsupported_hosting_service.feature
@@ -9,7 +9,7 @@ Feature: unsupported hosting platform
 
       This command requires hosting on one of these services:
       * Bitbucket
-      * Bitbucket datacenter
+      * Bitbucket Data Center
       * GitHub
       * GitLab
       * Gitea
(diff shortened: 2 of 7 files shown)


kevgo (approved):
Thanks for correcting this! :raised_hands: 

We have updated many config settings over the years. See [these two lists of deprecated config settings](https://github.com/git-town/git-town/blob/01bbcf564d46defcc68f5c9be5b45b96b30166e3/internal/config/configdomain/key.go#L174-L190).

Before the Git Town config file, when all configuration was stored in Git metadata, doing so was super easy: Git Town automatically updates old config settings to the new format. The change is completely transparent to the user.

However, now that we have a dedicated config file, silent updates are no longer feasible or advisable. To modernize configuration settings while maintaining backward compatibility, we’d need to follow a staged, non-atomic refactoring approach:

1. **Introduce the new variant:** Add the new config option, keep the old config option working. Release with bump of the minor app version.
2. **Deprecate the old variant:** The old config option still works, but results in a warning to users who still use it, asking them to update it.
3. **Remove the old variant:** When all users have transitioned to the new variant, remove the deprecated value in a major version bump.

When modernizing things this way, major version bumps ideally only remove deprecated functionality. Users who don't get any deprecation warnings are compatible with the new major version of the product.

For Git Town we cannot (and don't want to) verify step 3. So we typically stop at step 1 and remain liberal in the configurations we accept, leaving deprecated options functional indefinitely. 

Fixing the documentation to the correct name was absolutely the right call here. However, I don't see a compelling reason to change the actual configuration value. The current pattern of `platform name - dash - platform variant` feels appropriate. Ideally both  components (platform name and platform variant) should be concise, i.e. single words.

Given that most Git Town users on Bitbucket Data Center will either consult the documentation or rely on the setup assistant, the existing setup should suffice, no need to introduce additional complexity.

What do you think about this solution?

stephenwade:
Sounds good enough to me. I appreciate your thoughtful review!

PR #4494, opened 2025-01-04
tools/tests_sorted: add lintFuncDecl and tests

legeana:
Unlike tools/structs_sorted Issues is returned directly. These slices will be concatenated at a higher level. This API is much simpler to read than passing around a mutable pointer to a slice.

### DIFF
--- a/tools/tests_sorted/issue.go
+++ b/tools/tests_sorted/issue.go
@@ -1,0 +1,32 @@
+package main
+
+import (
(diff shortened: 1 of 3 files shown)


legeana:
@kevgo please take a look.

kevgo:
Thanks for the ping, and apologies that this PR fell through the cracks! This is starting to look really nice!

PR #4496, opened 2025-01-05
Docs: Update `go get` URLs to use correct version

stephenwade:
I discovered this when debugging why the Go Report Card was out of date. It turns out that they fetch the latest info the same way that `go get` does, which results in an outdated version being analyzed.

- https://pkg.go.dev/github.com/git-town/git-town shows v7.3.0+incompatible
- https://goreportcard.com/report/github.com/git-town/git-town shows v6.0.2+incompatible
- Changing `github.com/git-town/git-town` -> `github.com/git-town/git-town/v17` fixes both issues.


kevgo:
These two updates are pretty important. Thanks!

Issue #1665, opened 2021-07-08
Windows Defender AV is alerting on v7.5 installed with scoop as being 'Zpevdo' malware


rohancragg:
Threat name: Trojan:Win32/Zpevdo.B
path: %USERPROFILE%\scoop\apps\git-town\7.5.0\git-town.exe
SHA: 058b733479bb67412703bb697fecb424b5e752a5

Also 17 other security vendors are flagging it as malicious:
https://www.virustotal.com/gui/file/713f7a4332fba878068077c44bfc9f027776eba3146265addb826d064acc9b73/detection

kevgo:
Thanks for letting us know!

kevgo:
Have you ruled out a malware infection on your machine that spread to git-town.exe?

kevgo:
Executables written in Go are often falsely flagged as malicious by virus scanning programs. See https://go.dev/doc/faq#virus for more background. More discussion also at https://www.reddit.com/r/golang/comments/s1bh01/goexecutables_and_windows_defender.

PR #4508, opened 2025-01-07
tools/tests_sorted: make PositionalFields a normal function

legeana:
The iterator interface is unnecessary. A simple function is much simpler to use, test and understand.

### DIFF
--- a/tools/tests_sorted/matcher/matcher.go
+++ b/tools/tests_sorted/matcher/matcher.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"go/ast"
 	"go/token"
-	"iter"
 	"strconv"
 )
 
@@ -58,7 +57,7 @@ type FieldListMatcher interface {
 	Match(fields *ast.FieldList) Result
 }
 
-// PositionalFields yields (name, field) for each name in ast.FieldList in logical order.
+// PositionalFields returns (name, field) for each name in ast.FieldList in logical order.
 // For example func(a, b string, c int) will result in
 // - fields[0]: ast.Field{Names: []{"a", "b"}}
 // - fields[1]: ast.Field{Names: []{"c"}}
@@ -66,18 +65,14 @@ type FieldListMatcher interface {
 // - ("a", fields[0])
 // - ("b", fields[0])
 // - ("c", fields[1])
-func PositionalFields(fields *ast.FieldList) iter.Seq2[int, PositionalField] {
-	return func(yield func(int, PositionalField) bool) {
-		i := 0
-		for _, field := range fields.List {
-			for _, name := range field.Names {
-				if !yield(i, PositionalField{Name: name, Field: field}) {
-					return // Stopping early.
-				}
-				i++
-			}
+func PositionalFields(fields *ast.FieldList) []PositionalField {
+	var posFields []PositionalField
+	for _, field := range fields.List {
+		for _, name := range field.Names {
+			posFields = append(posFields, PositionalField{Name: name, Field: field})
 		}
 	}
+	return posFields
 }
 
 // IdentSelectorMatcher matches ast.SelectorExpr where both the selector
@@ -148,10 +143,7 @@ type FieldListPrefixMatcher struct {
 }
 
 func (self *FieldListPrefixMatcher) Match(fields *ast.FieldList) Result {
-	var posFields []PositionalField
-	for _, field := range PositionalFields(fields) {
-		posFields = append(posFields, field)
-	}
+	posFields := PositionalFields(fields)
 	for i, fieldMatcher := range self.Prefix {
 		if i >= len(posFields) {
 			return fmtResult("not enough fields, want at least %d, got %d", len(self.Prefix), len(posFields))
--- a/tools/tests_sorted/matcher/matcher_test.go
+++ b/tools/tests_sorted/matcher/matcher_test.go
@@ -37,18 +37,12 @@ func TestPositionalFields(t *testing.T) {
 	must.NoError(t, err)
 	funcType := expr.(*ast.FuncType)
 
-	var positionalIndices []int
-	var positionalFields []PositionalField
-	for i, field := range PositionalFields(funcType.Params) {
-		positionalIndices = append(positionalIndices, i)
-		positionalFields = append(positionalFields, field)
-	}
+	posFields := PositionalFields(funcType.Params)
 
-	must.Eq(t, []int{0, 1, 2}, positionalIndices)
-	must.Eq(t, len(positionalFields), 3)
-	must.Eq(t, "a", positionalFields[0].Name.Name)
-	must.Eq(t, "b", positionalFields[1].Name.Name)
-	must.Eq(t, "c", positionalFields[2].Name.Name)
+	must.Eq(t, len(posFields), 3)
+	must.Eq(t, "a", posFields[0].Name.Name)
+	must.Eq(t, "b", posFields[1].Name.Name)
+	must.Eq(t, "c", posFields[2].Name.Name)
 }
 
 func TestIdentSelectorMatcher(t *testing.T) {


kevgo (approved):
I agree. Iterators feel like a performance optimization in Go. Probably not needed for such a small tool.

kevgo:
I just meant to say that to me, iterators really shine when one can take advantage of lazy evaluation: in scenarios where creating data for another loop iteration is costly, and the loop terminates early.

For our case, creating an entire slice with the data for all loop iterations seems simpler without noteworthy downsides.

legeana:
> I just meant to say that to me, iterators really shine when one can take advantage of lazy evaluation: in scenarios where creating data for another loop iteration is costly, and the loop terminates early.

Yea, and Go isn't really designed for iterator chaining anyway... My instinct is to create iterator instead of returning collections because usually iterators are more flexible. Not really the case in Go as I see.

kevgo:
I completely agree. A part of me would love to port Git Town over to Rust - just as we have ported Git Town from Bash to Go 7 years ago (that's where some of the technical debt comes from). However, the effort required just doesn't seem justifiable—especially considering how challenging it is to [mix Rust with Go](https://vercel.com/blog/how-turborepo-is-porting-from-go-to-rust) and [what Joel Spolsky has to say about this](https://www.joelonsoftware.com/2000/04/06/things-you-should-never-do-part-i). Instead we’ve incorporated some of the aspects that make Rust so enjoyable to work with into the Git Town codebase, like option types, exhaustive struct and enum checking, clear data ownership, and many linters. It’s a step in the right direction but in the end it's still Go. On the good side, Go as a language pretty much fades into the background once one gets used to its style of (intentionally) keeping everything extremely simple.

legeana:
> However, the effort required just doesn't seem justifiable

I agree about the full rewrite. I don't know much about Go/Rust interoperability, but my instinct tells me that it is feasible, but quite a big undertaking. 


```
$ wc -l (find . -name '*.go' | grep -v '^./vendor/')
  48946 total

$ wc -l (find . -name '*.go' | grep -v '^./vendor/' | grep -v '_test\.go' )
  32380 total
```

So naively assuming 1:1, this is 32k lines of code to rewrite. The biggest challenge I see isn't really the rewrite itself, but maintaining the sliding API between the parts. We could probably split the codebase into a few bigger parts that can be reasonably isolated such as `internal/git/commands` can be migrated really easily and transparently. Then we could probably create an FFI translation layer for `gitdomain` and `configdomain` to migrate individual components.

But yea, I will be frank while I would love to, I won't have energy to do that. It's far better to focus on making good helper types, and occasionally rant about how the neighbour's grass is greener 🤣

kevgo:
Running `make stats` shows some statistics for this codebase (run `make` to see all available commands):

```
───────────────────────────────────────────────────────────────────────────────
Language                 Files     Lines   Blanks  Comments     Code Complexity
───────────────────────────────────────────────────────────────────────────────
Gherkin Specificati…       801     34212     2472        38    31702        916
Go                         755     48990     4141      1896    42953       3928
Markdown                    86      5444     1576         0     3868          0
YAML                        12       365       30        32      303          0
JSON                         8       148        3         0      145          0
JavaScript                   3       316       29       147      140         20
Powershell                   3        83       10        19       54          2
Plain Text                   2        19        6         0       13          0
SVG                          2         2        0         0        2          0
Shell                        2       190       22        18      150         14
TOML                         2        18        2         0       16          0
Handlebars                   1        34        4         0       30          0
License                      1        21        4         0       17          0
Makefile                     1       189       33         3      153         26
TypeScript Typings           1        24        3         4       17          3
nuspec                       1        38        0         0       38          0
───────────────────────────────────────────────────────────────────────────────
Total                     1681     90093     8335      2157    79601       4909
───────────────────────────────────────────────────────────────────────────────
Estimated Cost to Develop (organic) $2,676,453
Estimated Schedule Effort (organic) 19.99 months
Estimated People Required (organic) 11.89
───────────────────────────────────────────────────────────────────────────────
```

Your estimates are spot on! The tests need to be translated too, so it's the full 43k LOC.

Everything I heard from people who combined Rust and Go using CGO discourages me because there are so many issues: [Turborepo experience report](https://vercel.com/blog/how-we-continued-porting-turborepo-to-rust), [Go team member discouraging using CGO](https://dave.cheney.net/2016/01/18/cgo-is-not-go). So I think a rewrite in small steps is off the table. That leaves the all-at-once approach, which might look something like this:
1. Refactor the Go codebase to align as closely as possible with how it would work in Rust. This step is largely complete and has made Go so pleasant to work with that I'm okay continuing with it for the foreseeable future.
2. Wait for LLM-based translation tools to mature to a point where they can handle large-scale codebases like this effectively. 
3. Use the LLM to translate each file from Go to Rust.
4. Let the LLM address as many issues in the translated codebase as it can, then manually resolve the remaining problems. This could involve either further refactoring the Go code to improve translation fidelity or fixing the Rust codebase directly.
5. Manually replace non-translatable components, such as connectors to the GitHub API and installers.

Of course, this is still a monumental amount of work (at least one person-year) and probably never happens because that person-year is better spent adding features. In the meantime, I’ll continue letting Go teach me how to implement things in the simplest possible way, improving my Zen in the process. 🧘

legeana:
> Gherkin Specificati…       801     34212     2472        38    31702        916

I disagree. I think migrating this is going to be the easiest. See https://github.com/cucumber-rs/gherkin.
We just need to reimplement `steps.go`.

> Use the LLM to translate each file from Go to Rust.

I am somewhat sceptical on this approach.

> combined Rust and Go using CGO

I think for this specific project a better approach would be starting a subprocess and feeding the state as arguments / stdin. `git-town` already has json serialisation of the vm state, we could use that, and create an interopability level between various components. That would be a performance hit during the migration but would allow us to maintain a separation level between the components.

kevgo:
I’ve been considering the option of using separate Go and Rust binaries as well. The main challenge with this approach is distribution. Git Town is available through various package managers—Homebrew, MacPorts, FreshPorts, Nix, Arch, Chocolatey, Scoop, and a native Windows installer—many of which are maintained by third parties. Ensuring all these ecosystems can reliably compile both Rust and Go binaries, and package them together, introduces significant complexity to the distribution process.

legeana:
> I’ve been considering the option of using separate Go and Rust binaries as well. The main challenge with this approach is distribution. Git Town is available through various package managers—Homebrew, MacPorts, FreshPorts, Nix, Arch, Chocolatey, Scoop, and a native Windows installer—many of which are maintained by third parties. Ensuring all these ecosystems can reliably compile both Rust and Go binaries, and package them together, introduces significant complexity to the distribution process.

That makes sense if `git-town` doesn't own the distribution. One of the ideas I had was to build `git-town` as a single binary by embedding all the components to preserve the single binary concept. But I can see how this would make things more complex.

On the other hand, while I haven't looked into CGO issues myself, I've had positive experience encoding/decoding data at FFI level in Java/C++ boundary in the past which greatly reduces the complexity. So maybe if building CGO is feasible, serialisation could help us make it a bit more manageable.

kevgo:
Preserving the single binary concept by embedding all components into a single executable is an interesting idea. Are you envisioning the primary binary writing out and invoking another binary? That could work, but raises challenges around cross-compilation (automation around that wouldn't work out of the box anymore) and potentially flagging as malware. We've encountered false-positive virus reports in the past due to the embedded garbage collector's unconventional memory access patterns on Windows: #1665

Maybe the Rust binary could download its Go counterpart at runtime.

Thanks for sharing your good experiences with FFI. Was that primarily on the server side or for desktop applications? How were these apps packaged? The Vercel team faced several hurdles with this approach for their CLI tool (see links shared earlier) and ultimately had to integrate a third language (Zig) to make cross-compilation work.

Thinking about that makes me wonder whether one could sidestep many of these pitfalls by transpiling the Go components into [web assembly](https://go.dev/wiki/WebAssembly) and running that inside the Rust binary using a WASM runtime for Rust, like [wasmtime](https://github.com/bytecodealliance/wasmtime) or [wasmer](https://github.com/wasmerio/wasmer). This would be relatively straightforward, except for sharing complex data structures between Rust and WASM. The entire WASM ecosystem and tooling are pretty early-stage technologies.

AI might be able to create a prototype of a Rust app that executes an embedded WASM payload and interacts with it via JSON messages.

legeana:
Let's continue in #4521.

PR #4510, opened 2025-01-07
Skip syncing perennial branches based on detected sync status

kevgo:
Perennial branches are unique in the sense that they don't get changes merged in from their parent. This means we know at the time we compile the sync program whether we even need to sync a perennial branch or not. This PR eliminates unnecessary syncs of perennial branches.

This reduces the time to run `make cuke` from 22.5 sec to 21.5 sec. :muscle: 

TODO: also apply this to contribution and observed branches?


kevgo:
This is now available as part of [Git Town 17.3](https://github.com/git-town/git-town/releases/tag/v17.3.0). Enjoy!

PR #4511, opened 2025-01-07
Enable linting in tools/tests_sorted

legeana:
- Add missing t.Parallel calls.
- Use %w in fmt.Errorf to wrap errors.
- Disable ireturn in matcher.
- Use string concatenation instead of fmt.Sprintf.
- Use matcher_test package for unit tests.
- Disable prealloc linter. It's generally not recommended as it is a premature optimisation to follow this without a good reason.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,8 @@ lint-all: lint tools/rta@${RTA_VERSION}  # runs all linters
 	@(cd tools/stats_release && ../rta golangci-lint run)
 	@echo lint tools/structs_sorted
 	@(cd tools/structs_sorted && ../rta golangci-lint run)
+	@echo lint tools/tests_sorted
+	@(cd tools/tests_sorted && ../rta golangci-lint run --disable prealloc)
 	@echo lint tools/lint_steps
 	@(cd tools/lint_steps && ../rta golangci-lint run)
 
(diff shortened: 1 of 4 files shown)


kevgo:
Thanks for enabling the linters!

In my experience, the prealloc linter has been quite reliable—I’ve never encountered false positives with it. I recall reading that preallocating slices to their known size is considered a best practice, as it can reduce the number of allocations and minimize copying between backing arrays. Do you have any resources or links that explain why preallocating might be considered a premature optimization?

That said, for the Git Town codebase, this likely isn't a critical concern given the limited amount of data we handle. I'm fine with not using the prealloc linter if you prefer to leave it disabled.

legeana:
> In my experience, the prealloc linter has been quite reliable—I’ve never encountered false positives with it. I recall reading that preallocating slices to their known size is considered a best practice, as it can reduce the number of allocations and minimize copying between backing arrays. Do you have any resources or links that explain why preallocating might be considered a premature optimization?
> 
> That said, for the Git Town codebase, this likely isn't a critical concern given the limited amount of data we handle. I'm fine with not using the prealloc linter if you prefer to leave it disabled.

Let's continue in #4514.

PR #4513, opened 2025-01-07
Optimize syncing contribution and observed branches

kevgo:
Since contribution and observed branches don't merge in changes from their parent, their sync status is known at the time the VM program is compiled. This allows to determine whether they need to be synced at all at compile time.


kevgo:
This is now available as part of [Git Town 17.3](https://github.com/git-town/git-town/releases/tag/v17.3.0). Enjoy!

PR #4516, opened 2025-01-08
Fix CHANGELOG headings

FirelightFlagboy:
The `17.x.x` (`17.0.0` excluded) headings where not consistent with the rest of the document


stephenwade:
Thanks!

PR #4520, opened 2025-01-09
tools/tests_sorted: re-enable prealloc for consistency

### LINKED ISSUE #4514: Let's talk about prealloc linter
### Your message to us

Original messages by @kevgo consolidated by @legeana for a discussion in the same thread.

https://github.com/git-town/git-town/pull/4511#issuecomment-2576150680

In my experience, the prealloc linter has been quite reliable—I’ve never encountered false positives with it. I recall reading that preallocating slices to their known size is considered a best practice, as it can reduce the number of allocations and minimize copying between backing arrays. Do you have any resources or links that explain why preallocating might be considered a premature optimization?

That said, for the Git Town codebase, this likely isn't a critical concern given the limited amount of data we handle. I'm fine with not using the prealloc linter if you prefer to leave it disabled.

https://github.com/git-town/git-town/pull/4403#issuecomment-2576253870

I’d like to request that we don't disable the prealloc linter across the entire repo. The end-to-end test suite currently calls the Git Town executable over 1600 times. At this scale, even small inefficiencies can compound, making it worthwhile to tackle some of the low-hanging fruit around performance.

For context, there was a time when running make cuke took 18 minutes. Through targeted, data-driven optimizations, we've managed to bring that down to just 22 seconds today (on Apple Silicon running Asahi Linux—macOS remains slower due to filesystem bottlenecks). This makes it possible to lean on the end-to-end tests as part of development.

Here are a couple of examples of optimizations that made a meaningful impact on the test suite’s performance:

- https://github.com/git-town/git-town/pull/1441 found that creating a Git repo from scratch for each test was a major bottleneck. The solution involved setting up a fully configured Git environment once at the start of the test suite and cloning it for each test.
- https://github.com/git-town/git-town/pull/1447 found that disabling the garbage collector significantly reduces test suite runtime on CI (which is a severely performance constrained environment).

I'm curious about how you see this, and if I'm missing something. Let me know if you’d like more background or additional details! 🙂

legeana:
Closes #4514.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ lint-all: lint tools/rta@${RTA_VERSION}  # runs all linters
 	@echo lint tools/structs_sorted
 	@(cd tools/structs_sorted && ../rta golangci-lint run)
 	@echo lint tools/tests_sorted
-	@(cd tools/tests_sorted && ../rta golangci-lint run --disable prealloc)
+	@(cd tools/tests_sorted && ../rta golangci-lint run)
 	@echo lint tools/lint_steps
 	@(cd tools/lint_steps && ../rta golangci-lint run)
 
--- a/tools/tests_sorted/Makefile
+++ b/tools/tests_sorted/Makefile
@@ -3,6 +3,5 @@ help:  # prints all available targets
 
 test:  # runs all tests
 	go test ./...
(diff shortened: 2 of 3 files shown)


kevgo:
Thanks 🙏

PR #4549, opened 2025-02-04
setup assistant: add ff-only strategy for perennial branches

kevgo:


kevgo:
This is now available as part of [Git Town 18.0](https://github.com/git-town/git-town/releases/tag/v18.0.0).

PR #4611, opened 2025-03-26
hack/append/prepend: --propose flag

### LINKED ISSUE #4376: hack/append/prepend: --commit and --propose flags
The [boy scouts rule](https://innolution.com/resources/glossary/boy-scout-rule) is a  best practice that tooling like Git Town should make easy to perform, so that it happens more often. From https://egoless.engineering:

> We also needed to create the strong expectation that you’d leave things in a better state as you messed with other people’s stuff. That made people overjoyed to have you messing with their stuff. 

The way this is often done is to commit the cleanup into the same branch. Then either ship it as part of the branch, or extract the cleanup commit into its own branch. The former has the problem of mixing concerns, the latter is cumbersome because it requires running several commands: create the cleanup commit, create a new branch, move the commit, and propose the branch.

The most straightforward solution would be to commit the cleanup directly into a separate branch. Right now that also requires multiple commands to do. It would be nice if it was possible using a single command.

```zsh
> git town hack new-branch --commit
git checkout -b new-branch main
git commit
git checkout <old branch>
```

- `--commit` runs `git commit` on the new branch (requires stashed changes)
- `--propose` automatically proposes the new branch
- `--commit-message` commits using the given commit message (implies `--commit`)
- `--commit-all` stages all changes before committing (implies `--commit`)

kevgo:
resolves #4376


ruudk:
🤩 Can't wait to try this out. Friendly ping for a new tag 🥳

kevgo:
Thanks for the ping! I tried to get something else done for this release, but it's more complex, so I'm going to roll this out now.

kevgo:
It's available now: https://github.com/git-town/git-town/releases/tag/v18.2.0

PR #4715, opened 2025-04-14
add sha to the commit to beam dialog

pradeepmurugesan:
Fixed #4519 

Truncated the sha as the whole sha doesn't look nice. Hard coded the length to 7 as in github. 

Can be changed. Kindly let me know your thoughts.. 

looks like

```
gt prepend test1 --beam

Select the commits to beam into branch "test1"
0 > [ ] add sha to the commit to beam dialog (1637f07)
```

Issue #4519, opened 2025-01-09
Improve commit picker (git town prepend --beam)


ruudk:
### Your message to us

First of all, the new `--beam` functionality for `git town prepend` is awesome 😍 

Currently it looks like this:
![Screenshot 2025-01-09 at 14 46 58@2x](https://github.com/user-attachments/assets/55242e17-4432-4488-b732-a39003f6ce96)

Would it be an idea to improve this a bit? 

Maybe good to add the short commit sha in front of it.

And if possible, add a diff viewer.

This is what I see when I use my [git hack-pick](https://github.com/git-town/git-town/issues/3667#issuecomment-2222804895) script:
![Screenshot 2025-01-09 at 14 49 19@2x](https://github.com/user-attachments/assets/2e55f2fe-d117-4ad7-912f-5db58a9bd752)

kevgo:
Guidance for implementation: 

- in the commit picker dialog, change the [text for dialog entries](https://github.com/git-town/git-town/blob/main/internal/cli/dialog/commits_to_beam.go#L23) from commit title to commit title + SHA 
- debug the commit picker dialog by running `git town debug commits-to-beam 8`

kevgo:
I believe this wraps up the work for this ticket. Showing a diff preview feels out of scope.

Git Town needs to run reliably across a wide range of environments, including some fairly constrained ones like Windows terminals, editor-integrated emulators, and CI systems that render console output in the browser. Many of these environments already have trouble rendering even basic UI elements like checkboxes.

While a diff preview might offer some value in a few scenarios, it risks degrading the experience for a broader set of users by producing noisy or unreadable output. On balance, it seems better to leave this feature out.

Also worth noting: for the preview to be truly useful, we'd likely need to support scrolling, searching, and possibly paging. All of which add significant complexity. Given our current bandwidth, it makes more sense to prioritize higher-impact improvements.

PR #4719, opened 2025-04-16
Format SHA in beam dialog with `rev-parse --short`

stephenwade:
Ref: https://github.com/git-town/git-town/issues/4519, #4715

> We should also use `git rev-parse --short` to shorten the hash, instead of hard-coding 7.
> 
> 1. Git automatically picks a length for short hashes based on how many objects are in the repository; 7 characters isn't enough to guarantee uniqueness for large repos. For example, in this repo, Git uses 9 characters. See: [git/git@e6c587c](https://github.com/git/git/commit/e6c587c733b4634030b353f4024794b08bc86892)
> 2. This would also cause Git Town to respect the [core.abbrev](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreabbrev) setting.
> 
> I was going to suggest just reading the core.abbrev setting, but while looking into it I learned that Git automatically picks a good length if the setting is unspecified, so I think we should just let Git do the work of abbreviating commits.

https://github.com/git-town/git-town/pull/4715#issuecomment-2808287936

Closes #4726

### DIFF
--- a/internal/cli/dialog/commits_to_beam.go
+++ b/internal/cli/dialog/commits_to_beam.go
@@ -5,25 +5,27 @@ import (
 
 	"github.com/git-town/git-town/v19/internal/cli/dialog/components"
 	"github.com/git-town/git-town/v19/internal/cli/dialog/components/list"
+	"github.com/git-town/git-town/v19/internal/git"
 	"github.com/git-town/git-town/v19/internal/git/gitdomain"
 	"github.com/git-town/git-town/v19/internal/messages"
 )
 
-const (
-	commitsToBeamTitle = `Select the commits to beam into branch %q`
-	shaLength          = 7
-)
+const commitsToBeamTitle = `Select the commits to beam into branch %q`
 
 // lets the user select commits to beam to the target branch
-func CommitsToBeam(commits []gitdomain.Commit, targetBranch gitdomain.LocalBranchName, inputs components.TestInput) (gitdomain.Commits, bool, error) {
+func CommitsToBeam(commits []gitdomain.Commit, targetBranch gitdomain.LocalBranchName, git git.Commands, querier gitdomain.Querier, inputs components.TestInput) (gitdomain.Commits, bool, error) {
 	if len(commits) == 0 {
 		return gitdomain.Commits{}, false, nil
 	}
 	entries := make(list.Entries[gitdomain.Commit], len(commits))
 	for c, commit := range commits {
+		shortSHA, err := git.ShortenSHA(querier, commit.SHA)
+		if err != nil {
+			return gitdomain.Commits{}, false, err
+		}
 		entries[c] = list.Entry[gitdomain.Commit]{
 			Data: commit,
-			Text: fmt.Sprintf("%s (%s)", commit.Message.String(), commit.SHA.TruncateTo(shaLength).String()),
+			Text: fmt.Sprintf("%s (%s)", commit.Message.String(), shortSHA),
 		}
 	}
 	selection, aborted, err := components.CheckList(entries, []int{}, fmt.Sprintf(commitsToBeamTitle, targetBranch), "", inputs)
--- a/internal/cli/dialog/commits_to_beam.go
+++ b/internal/cli/dialog/commits_to_beam.go
@@ -25,7 +25,7 @@ func CommitsToBeam(commits []gitdomain.Commit, targetBranch gitdomain.LocalBranc
 		}
 		entries[c] = list.Entry[gitdomain.Commit]{
 			Data: commit,
-			Text: fmt.Sprintf("%s (%s)", commit.Message.String(), shortSHA),
+			Text: fmt.Sprintf("%s %s", shortSHA, commit.Message.String()),
 		}
 	}
 	selection, aborted, err := components.CheckList(entries, []int{}, fmt.Sprintf(commitsToBeamTitle, targetBranch), "", inputs)
--- a/internal/git/commands_test.go
+++ b/internal/git/commands_test.go
@@ -1355,6 +1355,29 @@ func TestBackendCommands(t *testing.T) {
 		})
 	})
 
+	t.Run("ShortenSHA", func(t *testing.T) {
+		t.Parallel()
+		runtime := testruntime.Create(t)
+		branch := gitdomain.NewLocalBranchName("branch")
+		runtime.CreateBranch(branch, initial.BranchName())
+		runtime.CreateCommit(testgit.Commit{
+			Branch:      branch,
+			FileContent: "file1",
+			FileName:    "file1",
+			Message:     "first commit",
+		})
+		commits, err := runtime.Commands.CommitsInBranch(runtime.TestRunner, "branch", Some(gitdomain.NewLocalBranchName("initial")))
+		must.NoError(t, err)
+		cmds := git.Commands{
+			CurrentBranchCache: &cache.WithPrevious[gitdomain.LocalBranchName]{},
+			RemotesCache:       &cache.Cache[gitdomain.Remotes]{},
+		}
+		have, err := cmds.ShortenSHA(runtime, commits[0].SHA)
+		must.NoError(t, err)
+		must.True(t, len(commits[0].SHA.String()) == 40)
+		must.True(t, len(have.String()) == 7)
+	})
+
 	t.Run("StashEntries", func(t *testing.T) {
 		t.Parallel()
 		t.Run("some stash entries", func(t *testing.T) {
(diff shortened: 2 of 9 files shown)


internal/cli/dialog/commits_to_beam.go
```diff
@@ -15,15 +15,19 @@ const (
 )
 
 // lets the user select commits to beam to the target branch
-func CommitsToBeam(commits []gitdomain.Commit, targetBranch gitdomain.LocalBranchName, inputs components.TestInput) (gitdomain.Commits, bool, error) {
+func CommitsToBeam(commits []gitdomain.Commit, targetBranch gitdomain.LocalBranchName, formatSHA func(gitdomain.SHA) (string, error), inputs components.TestInput) (gitdomain.Commits, bool, error) {
```
pradeepmurugesan, line 18: can `format` be part of the `SHA` instead of passing it as a function ? like `abbreviated()`, so we can just call `commit.SHA.abbreviated` ?

internal/cli/dialog/commits_to_beam.go
```diff
@@ -15,15 +15,19 @@ const (
...
 	}
 	entries := make(list.Entries[gitdomain.Commit], len(commits))
 	for c, commit := range commits {
+		sha, err := formatSHA(commit.SHA)
+		if err != nil {
+			return gitdomain.Commits{}, false, err
+		}
 		entries[c] = list.Entry[gitdomain.Commit]{
 			Data: commit,
-			Text: fmt.Sprintf("%s (%s)", commit.Message.String(), commit.SHA.TruncateTo(shaLength).String()),
```
pradeepmurugesan, line 26: Nit: we can also remove the `shaLength` declared above

internal/cli/dialog/commits_to_beam.go
stephenwade: I'm not sure. The difficulty is that it needs to invoke `git`, so it needs a git.Commands and a gitdomain.Runner, or a closure with access to those. I'll wait to see what @kevgo suggests before making changes.

kevgo:
```suggestion
func CommitsToBeam(commits []gitdomain.Commit, targetBranch gitdomain.LocalBranchName, git git.Commands, inputs components.TestInput) (gitdomain.Commits, bool, error) {
```

kevgo: I think this would introduce a circular dependency, which is not allowed in Go (for good reasons: to avoid tight coupling). `SHA` exists in the `gitdomain` package. The `git` package depends on the `gitdomain` package. This means we cannot add anything from the `git` package to the `gitdomain` package.

pradeepmurugesan: Would be nice if we can add this `abbreviatedSha` as part of the `gitdomain.Commit` struct as an `optional` .  Almost all the places where we construct the above struct should have access to the `gitCommands`

stephenwade:
> My proposal is to keep the code that truncates the full-length SHAs (now using `git rev-parse --short` instead of `SHA.TruncateTo`) where it is now, in [the code that prepares the data to display in the dialog](https://github.com/git-town/git-town/blob/main/internal/cli/dialog/commits_to_beam.go?rgh-link-date=2025-04-16T19%3A12%3A25Z). Can you just run `git rev-parse --short` there? You would need to provide the `git.Commands` instance (typically `repo.Git`) as an additional argument to the `CommitsToBeam` function.

This was my first attempt. I added `gitCommands git.Commands, backend gitdomain.RunnerQuerier` as additional arguments to `CommitsToBeam` and called `gitCommands.SHAForRef(backend, commit.SHA)` inside `CommitsToBeam`. But I ran into an issue with internal/cmd/debug/enter_commits_to_beam.go, which uses `CommitsToBeam` and doesn't have a Commands or a RunnerQuerier to pass in.

stephenwade:
I reverted #4339 in this PR because it was preventing `--short` from working the way I expected. I've opened #4722 to try resolving that another way.

kevgo:
Please take https://github.com/git-town/git-town/pull/4726 as a reference for how to run Git commands inside a dialog.

internal/cli/dialog/commits_to_beam.go
```diff
@@ -5,25 +5,27 @@ import (
...
 	entries := make(list.Entries[gitdomain.Commit], len(commits))
 	for c, commit := range commits {
+		shortSHA, err := git.ShortenSHA(querier, commit.SHA)
+		if err != nil {
+			return gitdomain.Commits{}, false, err
+		}
 		entries[c] = list.Entry[gitdomain.Commit]{
 			Data: commit,
-			Text: fmt.Sprintf("%s (%s)", commit.Message.String(), commit.SHA.TruncateTo(shaLength).String()),
+			Text: fmt.Sprintf("%s %s", shortSHA, commit.Message.String()),
```
kevgo, line 28: Good call, I agree that it reads better without the parentheses.

internal/git/commands_test.go
```diff
@@ -1355,6 +1355,29 @@ func TestBackendCommands(t *testing.T) {
...
+			FileName:    "file1",
+			Message:     "first commit",
+		})
+		commits, err := runtime.Commands.CommitsInBranch(runtime.TestRunner, "branch", Some(gitdomain.NewLocalBranchName("initial")))
+		must.NoError(t, err)
+		cmds := git.Commands{
+			CurrentBranchCache: &cache.WithPrevious[gitdomain.LocalBranchName]{},
+			RemotesCache:       &cache.Cache[gitdomain.Remotes]{},
+		}
+		have, err := cmds.ShortenSHA(runtime, commits[0].SHA)
```
kevgo, line 1375: This should work:

```suggestion
		have, err := runtime.ShortenSHA(runtime, commits[0].SHA)
```

Background: `runtime` is a [TestCommands](https://github.com/git-town/git-town/blob/main/test/commands/test_commands.go#L26-L31) instance, which [embeds](https://gobyexample.com/struct-embedding) `git.Commands`. It's imported as `prodgit.Commands` to avoid conflicting namespaces). `ShortenSHA` comes from that embedded type.

Please let me know if this wasn't obvious at first glance. Happy to make any of this more obvious/explicit in the code. I've been looking at this code for so long that I might be missing things that aren't as clear to fresh eyes.

internal/git/commands_test.go
```diff
@@ -1355,6 +1355,29 @@ func TestBackendCommands(t *testing.T) {
...
+		must.NoError(t, err)
+		must.True(t, len(commits[0].SHA.String()) == 40)
+		must.True(t, len(have.String()) == 7)
```
kevgo, line 1378: This test could also verify that the shortened SHA has the same content as the beginning of the long SHA.

```suggestion
		must.True(t, len(have.String()) == 7)
		must.EqOp(t, have.String(), commits[0].SHA.String()[0:7])
```

kevgo (approved):
Excellent work, thanks so much for implementing this! :pray:

internal/git/commands_test.go
stephenwade: It wasn't obvious, although maybe that's due to my unfamiliarity with the language. I was mostly stumbling around until the red squiggles in the editor went away 😅

Usage of other Commands usually looks like passing a `Runner` to a `Commands` (e.g. `args.Git.Commit(args.Frontend, ...)`), so it didn't occur to me that `runtime` might satisfy both interfaces. I just assumed `runtime` was a `Runner` and looked for an example of a `Commands` elsewhere in the file to copy/paste.

> Happy to make any of this more obvious/explicit in the code. I've been looking at this code for so long that I might be missing things that aren't as clear to fresh eyes.

I would expect the `Commands`, `Runner`, etc. to be separate structs in test code like they are in other code. That might have made it easier for me to put the pieces together when writing this test.

kevgo: Thanks for taking the time to provide this feedback. You make very good points. I'm going to implement this.

kevgo: Btw, the codebase is intentionally type-heavy so that most bugs manifest as red squiggles in the editor. That way, issues are easy to spot and fix during development or at lint time.

I’ve made the Git access in test code a bit more explicit in #4734 to help with discoverability.

On separating the `subshell.TestRunner` from `TestCommands`: I'm a bit hesitant to split them up. In production, we separate the command logic from the runner because there are two different runner types: one that runs Git commands and captures output (backend), which is used to discover the state the Git repo is in. And one that streams output directly to the terminal (frontend), which is used to execute Git commands that change the state of the Git repo in the name of the user. That distinction doesn’t really apply in test code, where there’s always just a single runner.

So, there's a tradeoff between mirroring the production architecture vs keeping the test setup simpler. In cases like this, I lean toward simplicity, so I’d prefer to keep the test runner bundled with `TestCommands` for now. I'll keep what you said in the back of my mind, though!

PR #4726, opened 2025-04-17
Prototype of beam dialog that shortens SHAs using Git

kevgo:
Here is a quick spike that demonstrates how to inject Git access into a dialog.

This isn't finished, it needs unit tests and polish.

@stephenwade I hope you find this useful while implementing #4719


kevgo:
It would be nice if https://github.com/git-town/git-town/blob/main/internal/git/commands_test.go contains a unit test for `Commands.ShortenSHA`.

Forgot to mention it, you can run `make install && git-town debug commits-to-beam 7` to debug the dialog!

Issue #4716, opened 2025-04-14
git town config - github token & adding a --show-origin flag


pradeepmurugesan:
**Github token**

- github token is fetched in the following order
    - Environment variables
      - $GITHUB_TOKEN
      - $GITHUB_AUTH_TOKEN
    - Config file

This happens when the connector is instantiated.  I had the correct token in the config file , but there was another token in `.zshrc` for a different reason. my `git town propose` always resulted in 404 as it was using the token from `GITHUB_TOKEN` env var. 

my `git town config` always returned the one from the file. 

It would be nice if we can avoid calling the `GetAPIToken` when creating the new connector

```
case configdomain.ForgeTypeGitHub:
		var err error
		connector, err = github.NewConnector(github.NewConnectorArgs{
			APIToken:  github.GetAPIToken(config.NormalConfig.GitHubToken),
			Log:       log,
			RemoteURL: remoteURL,
		})
		return Some(connector), err
```

instead we call `GetAPIToken` while constructing the actual `NormalConfigData` so it's easier to figure the above problem. 

**--show-origin**

Would also be nice if we can add something like `--show-origin` like github which can print out where the config is from. Could be a separate issue as this can be worked independently.

kevgo:
Great ideas! I agree with you that `GetAPIToken` should be called when constructing the config data structures. The current behavior is buggy.

I think it's okay if `git town config` always shows the origin, i.e. behaves as if `--show-origin` is always enabled.

pradeepmurugesan:
I have assigned this to myself. Hope that is ok.  

I am exploring the `config` package now and trying to understand how the configs are fetched from different location.  I also see a `envconfig` sub package for environment variables. 

I will try to come up with an approach for the github api token fix.

kevgo:
I have implemented a spike of the remaining feature for this ticket: showing where config values come from. After exploring it, I don't think the value justifies the added complexity.

To do this properly, we'd need to replace all config values with a generic container that carries both the value and a description of its origin. Every piece of logic that provides, merges, or consumes config data would then need to wrap and unwrap these containers.

Even with that in place, we couldn't reliably distinguish between CLI flags like `-m` vs `--merge`, since the flag parser doesn’t expose which variant was used.

On top of that, the actual utility is pretty limited. Even the motivating use case doesn't strictly need it. You can already see which tokens Git Town uses and trace where they come from in your environment.

The extra infrastructure would also make the code harder to maintain and more intimidating for contributors.

[...]

PR #4728, opened 2025-04-18
GitHub token config gh 4716

pradeepmurugesan:
fixes a part of #4716. github api token is added at the time of config construction

### DIFF
--- a/internal/config/envconfig/load.go
+++ b/internal/config/envconfig/load.go
@@ -1,0 +1,11 @@
+package envconfig
+
+import (
+	"github.com/git-town/git-town/v19/internal/config/configdomain"
+)
+
+func Load() configdomain.PartialConfig {
+	partialConfig := configdomain.EmptyPartialConfig()
+	partialConfig.GitHubToken = GithubAPIToken()
+	return partialConfig
+}
--- a/internal/config/envconfig/load_test.go
+++ b/internal/config/envconfig/load_test.go
@@ -1,0 +1,70 @@
+package envconfig_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/git-town/git-town/v19/internal/config/configdomain"
+	"github.com/git-town/git-town/v19/internal/config/envconfig"
+	"github.com/git-town/git-town/v19/pkg/prelude"
+	"github.com/shoenig/test/must"
+)
+
+func TestLoad(t *testing.T) {
+	tokenText := "my-github-token"
+	authTokenText := "my-github-auth-token"
+
+	tests := []struct {
+		name        string
+		githubToken *string
+		authToken   *string
+		want        prelude.Option[configdomain.GitHubToken]
+	}{
+		{
+			name:        "loads from GITHUB_TOKEN when both are set",
+			githubToken: &tokenText,
+			authToken:   &authTokenText,
+			want:        prelude.Some(configdomain.GitHubToken(tokenText)),
+		},
+		{
+			name:        "loads from GITHUB_AUTH_TOKEN if GITHUB_TOKEN is empty",
+			githubToken: ptr(""),
+			authToken:   &authTokenText,
+			want:        prelude.Some(configdomain.GitHubToken(authTokenText)),
+		},
+		{
+			name:        "loads from GITHUB_TOKEN only",
+			githubToken: &tokenText,
+			authToken:   nil,
+			want:        prelude.Some(configdomain.GitHubToken(tokenText)),
+		},
+		{
+			name:        "returns none when no env is set",
+			githubToken: nil,
+			authToken:   nil,
+			want:        prelude.None[configdomain.GitHubToken](),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Clear environment
+			os.Unsetenv("GITHUB_TOKEN")
+			os.Unsetenv("GITHUB_AUTH_TOKEN")
+
+			if tt.githubToken != nil {
+				t.Setenv("GITHUB_TOKEN", *tt.githubToken)
+			}
+			if tt.authToken != nil {
+				t.Setenv("GITHUB_AUTH_TOKEN", *tt.authToken)
+			}
+
+			cfg := envconfig.Load()
+			must.Eq(t, tt.want, cfg.GitHubToken)
+		})
+	}
+}
+
+func ptr(s string) *string {
+	return &s
+}
--- a/internal/execute/open_repo.go
+++ b/internal/execute/open_repo.go
@@ -10,6 +10,7 @@ import (
 	"github.com/git-town/git-town/v19/internal/config"
 	"github.com/git-town/git-town/v19/internal/config/configdomain"
 	"github.com/git-town/git-town/v19/internal/config/configfile"
+	"github.com/git-town/git-town/v19/internal/config/envconfig"
 	"github.com/git-town/git-town/v19/internal/config/gitconfig"
 	"github.com/git-town/git-town/v19/internal/git"
 	"github.com/git-town/git-town/v19/internal/git/gitdomain"
@@ -88,6 +89,7 @@ func OpenRepo(args OpenRepoArgs) (OpenRepoResult, error) {
 		Access:        configGitAccess,
 		ConfigFile:    configFile,
 		DryRun:        args.DryRun,
+		EnvConfig:     envconfig.Load(),
 		FinalMessages: finalMessages,
 		GitVersion:    gitVersion,
 		GlobalConfig:  globalConfig,
--- a/internal/config/configdomain/partial_config.go
+++ b/internal/config/configdomain/partial_config.go
@@ -46,6 +46,7 @@ type PartialConfig struct {
 func EmptyPartialConfig() PartialConfig {
 	return PartialConfig{
 		Aliases: Aliases{},
+		Lineage: NewLineage(),
 	} //exhaustruct:ignore
 }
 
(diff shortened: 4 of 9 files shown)


internal/config/envconfig/load_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"os"
+	"testing"
+
+	"github.com/git-town/git-town/v19/internal/config/configdomain"
+	"github.com/git-town/git-town/v19/internal/config/envconfig"
+	"github.com/git-town/git-town/v19/pkg/prelude"
+	"github.com/shoenig/test/must"
+)
+
+func TestLoad(t *testing.T) {
```
kevgo, line 13: This is how you disable the `tparallel` linter for this entire test function and all its embedded subtests:

```suggestion
//nolint:tparallel
func TestLoad(t *testing.T) {
```

internal/config/envconfig/load_test.go
```diff
@@ -0,0 +1,69 @@
...
+		// Call your Load function
+		cfg := envconfig.Load()
+
+		// Assert expected value in the config
+		got, _ := cfg.GitHubToken.Get()
+		must.Eq(t, text, got.String())
+	})
+
+	//nolint:parallelTest // setting env vars
+
```
kevgo, line 55: No need for this anymore if you implement the suggestion above.

```suggestion
```

internal/config/envconfig/load_test.go
```diff
@@ -0,0 +1,69 @@
...
+		os.Unsetenv("GITHUB_TOKEN")
+
+		// Call your Load function
+		cfg := envconfig.Load()
+
+		// Assert expected value in the config
+		got := cfg.GitHubToken
+		must.Eq(t, prelude.None[configdomain.GitHubToken](), got)
+	})
+}
```
kevgo, line 69: This is a really nice test suite, and I'm perfectly happy to merge it the way it is right now. If you want to learn a nice trick, you could simplify these tests into a [table-driven test](https://go.dev/wiki/TableDrivenTests).

internal/config/envconfig/load.go
```diff
@@ -0,0 +1,11 @@
+package envconfig
+
+import (
+	"github.com/git-town/git-town/v19/internal/config/configdomain"
+)
+
+func Load() configdomain.PartialConfig {
```
kevgo, line 7: Just wanted to say that it's a brilliant approach to return a `PartialConfig` instance that contains all the data from env variables here and treat the environment like another source for config data!

This opens up the possibility to set Git Town settings via env variables in the future, if the need for that ever arises!

internal/execute/open_repo.go
```diff
@@ -7,6 +7,8 @@ import (
 	"os/exec"
 	"runtime"
 
+	"github.com/git-town/git-town/v19/internal/config/envconfig"
+
```
kevgo, line 11: Let's get rid of the empty line here, all imports should be one block.

```suggestion
```

kevgo (commented):
Very nice cleanup of the architecture, big thanks for doing this! :pray: You have removed one of the last remaining architecture smells from the time when Git Town was written in Bash many years ago.

This PR is very well done and overall good to go. I just have a number of smaller suggestions to polish details. I can't wait to merge this!

internal/config/envconfig/load_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/git-town/git-town/v19/internal/config/configdomain"
+	"github.com/git-town/git-town/v19/internal/config/envconfig"
+	"github.com/git-town/git-town/v19/pkg/prelude"
+	"github.com/shoenig/test/must"
+)
+
+func TestLoad(t *testing.T) {
+	const text = "my-github-token"
+	const alternateText = "my-github-auth-token"
```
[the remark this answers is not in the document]
pradeepmurugesan: anything with token gets flagged by the ci 😞 

```
tools/rta golangci-lint run
Error: internal/config/envconfig/load_test.go:15:8: G101: Potential hardcoded credentials (gosec)
	const tokenText = "my-github-token"
	      ^
Error: internal/config/envconfig/load_test.go:16:8: G101: Potential hardcoded credentials (gosec)
	const authTokenText = "my-github-auth-token"
```

internal/config/envconfig/load.go
pradeepmurugesan: Yeah the existing architecture allowed me to extend this nicely without touching the other configs :-). I really like how the separation of concern is handled in this repo. The boundaries of the domains are defined super clearly/strictly ❤️.

One Q I have in mind is should the env config takes precedence over everything else? I mean if usually we would override things for the repository with local config right ? 

Should we just inverse the priority ?

internal/config/envconfig/load_test.go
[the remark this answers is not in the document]
kevgo: I have just disabled scanning for hard-coded credentials in https://github.com/git-town/git-town/pull/4746 because that linter is way too loose and produces too many false positives.

internal/config/envconfig/load_test.go
kevgo: It's not you, it's golangci-lint, a sometimes flimsy third-party framework for running dozens on third-party Go linters concurrently. There are two linters at play here: `tparallel` and `paralleltest`. I have found a combination that doesn't produce warnings on CI. 

Btw, to run the linters on your machine, simply run `make lint`.

internal/config/envconfig/load.go
kevgo: > Yeah the existing architecture allowed me to extend this nicely without touching the other configs :-). I really like how the separation of concern is handled in this repo. The boundaries of the domains are defined super clearly/strictly ❤️.

Thanks for recognizing this! It took _a lot_ of refactoring to identify the various domains and concerns and model them orthogonal to each other. I think that's time well spent, because it makes maintaining this codebase doable, predictable, and fun!

kevgo: > One Q I have in mind is should the env config takes precedence over everything else? I mean if usually we would override things for the repository with local config right ?

That's a good question. I think the env config applies to all Git repos and multiple apps (the GitHub CLI and Git Town both use `GITHUB_TOKEN`), but the Git config applies only to Git Town and sometimes only to a single repo. Hence the Git config should override the env config, because more specific overrides more generic.

The env config should override the config file, though. Because the config file applies to all users, while the env config applies only to one user.

internal/config/envconfig/load_test.go
kevgo: Forgot to mention, I added the fix to your branch. The tests are green now.

internal/config/envconfig/load.go
stephenwade: > Hence the Git config should override the env config, because more specific overrides more generic.

Hmm, I disagree. Environment variables can be as specific as per-command, and we can't tell where the environment variable comes from. I think if anything is set in the environment, it should override everything else.

For example, running `GIT_CONFIG_PARAMETERS="'color.ui=never'" git status` will show status output without color, even if the Git config for this repo has `color.ui = always`.

pradeepmurugesan: Cool. I think the current behaviour is fine then 👍

internal/config/envconfig/load_test.go
pradeepmurugesan: I have updated the test to use the table driven tests. Kindly have a look 🙏

internal/config/envconfig/load.go
kevgo: Yes, you were right all along! It just took me a while to reach your level of insight :smile:

internal/config/envconfig/load_test.go
```diff
@@ -0,0 +1,70 @@
...
+			// Clear environment
+			os.Unsetenv("GITHUB_TOKEN")
+			os.Unsetenv("GITHUB_AUTH_TOKEN")
+
+			if tt.githubToken != nil {
+				t.Setenv("GITHUB_TOKEN", *tt.githubToken)
+			}
+			if tt.authToken != nil {
+				t.Setenv("GITHUB_AUTH_TOKEN", *tt.authToken)
+			}
```
kevgo, line 60: Nitpick: You could avoid resetting the env variables first if you set them to a value later anyways.


```suggestion

			if tt.githubToken != nil {
				t.Setenv("GITHUB_TOKEN", *tt.githubToken)
			} else {
				os.Unsetenv("GITHUB_TOKEN")
			}
			if tt.authToken != nil {
				t.Setenv("GITHUB_AUTH_TOKEN", *tt.authToken)
			} else {
				os.Unsetenv("GITHUB_AUTH_TOKEN")
			}
```

internal/config/envconfig/load.go
pradeepmurugesan: fixed with this [commit](https://github.com/git-town/git-town/pull/4728/commits/abaeda0af3379249dacbd82b7e7bac879c31ab87). Had to instantiate Lineage in the `EmptyPartialConfig` to avoid `panic: assignment to entry in nil map`

pradeepmurugesan:
> Very well done! I leave it up to you to optionally implement the last suggestion, and then merge this PR yourself when you are done. Thanks again for this nice improvement! 🚀

I will leave the test as it is. Unsetting the env in the else block hides the test set up a bit..  Kindly let me know your thoughts. 

Had to fix the ordering though since we want the env variable to override everything else.. Fixed in the [commit](panic: assignment to entry in nil map). 

Kindly let me know if you are happy with the recent commit and the above comment regarding the test setup. Will merge 🤞

kevgo:
Looks good to me, ship it! :speedboat:

PR #4733, opened 2025-04-18
Improve code

stephenwade:
I'm not sure why this showed up in VS Code but not in CI.

Screenshot 2025-04-18 at 6 52 49 PM


kevgo (approved):

kevgo:
Nice simplification! Yes, CI currently [doesn't run](https://github.com/git-town/git-town/blob/b6bf3df0bdf32ac480af116c7a397d56f53cfc04/Makefile#L61) the `go-staticcheck` linter because the precompiled binaries it distributes are compiled with an older version of Go, which makes it incompatible with the more modern syntax in our codebase. I'll set up compilation from source for this linter, so that we can re-enable it!

PR #4737, opened 2025-04-19
Use consistent variable name for git arguments

stephenwade:
Most functions in this file assemble the arguments to `git` in a variable named `args`. This PR updates the two functions that use a different variable name to match the others.


kevgo (approved):
Thanks for cleaning this up!

PR #4740, opened 2025-04-19
Simplify removing Git config

stephenwade:
This PR simplifies RemoveLocalGitConfiguration. Instead of removing only `git-town-branch` keys that were parsed, it removes all keys that start with `git-town-branch.`.

I'm not sure this is necessary, but it feels cleaner. Since the first part of the function removes all `git-town.*` keys, the second part should remove all `git-town-branch.*` keys.

Side-note: it's a shame that `git config --remove-section git-town-branch` doesn't work. It only looks for keys with two levels (e.g. `git-town-branch.abc`, not `git-town-branch.abc.parent`).


kevgo (approved):
This is so much better, thanks for cleaning this up!

kevgo:
The low-level configuration data (`configdomain.SingleSnapshot`) has only been exposed to the rest of the codebase relatively recently. There may be additional areas where leveraging it could help simplify or clean up existing logic.

PR #4741, opened 2025-04-19
Simplify checking for unexpected Git Town configuration

stephenwade:
Anything matched by `git config --get-regex git-town-branch` would already have been found by
`git config --get-regex git-town`.


kevgo (approved):
Nice catch! This must be a leftover from before we used regexes here.

PR #4742, opened 2025-04-20
Remove unused functions from test_commands.go

stephenwade:
I can't figure out how to run deadcode on this file, but ideally this would've been detected by the linters.

### DIFF
--- a/test/commands/test_commands.go
+++ b/test/commands/test_commands.go
@@ -353,11 +353,6 @@ func (self *TestCommands) LineageTable() datatable.DataTable {
 	return result
 }
 
-// SetGitAlias sets the Git alias with the given name to the given value.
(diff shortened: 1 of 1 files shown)


kevgo:
Nice cleanup, thanks! The [Makefile](https://github.com/git-town/git-town/blob/b3eb33fbc0c3ee3d7c1754c744633cd2ed81d160/Makefile#L161-L189) gives an example for calling deadcode. I think it can only run on entire packages since it needs to see the entire call graph to know which code isn't used. 

In my experience it catches some dead code but not all. Which is still better than nothing. And it struggles to see when code is used in tests, even when using the `-test` switch. I suspect part of the problem is that Go has no concept of visibility only within a module. I'll try moving the `test` folder into the [internal folder](https://go.dev/doc/go1.4#internalpackages) to see if that makes a difference.

PR #4744, opened 2025-04-20
Use `-c rebase.updateRefs=false` instead of checking Git version

stephenwade:
We already did this in `Rebase`, but not in `RebaseOnto` or `RemoveCommit`.

Other options would be:

1. Use `git -c rebase.updateRefs=false`, which wouldn't require checking the Git version (I verified with Git 2.30).
2. Update the minimum required Git version to 2.36 (the fact that no one has reported this as an issue probably means very few people are using 4-year-old versions of Git with Git Town).


kevgo (approved):
Thanks for adding this! I’m good with either approach. Personally, I slightly prefer the `-c rebase.updateRefs=false` syntax since it avoids the version check. The only reason I didn’t go with it originally is that I wasn’t aware of it. Feel free to go with whichever option you prefer and merge when ready. Appreciate the improvements!

PR #4749, opened 2025-04-21
Use `git status -z` instead of parsing user-facing output

stephenwade:
This PR updates usages of `git status` to use the `-z` flag, which outputs file status information in a machine-readable format separated by nulls (`\x00`). Git Town now parses this output instead of user-facing output.

This stops Git Town from being tricked by certain file names like "working tree clean" or "rebase in progress".


internal/git/commands.go
```diff
@@ -519,11 +520,26 @@ func (self *Commands) HasLocalBranch(runner gitdomain.Runner, name gitdomain.Loc
 	return runner.Run("git", "show-ref", "--quiet", "refs/heads/"+name.String()) == nil
 }
 
-func (self *Commands) HasMergeInProgress(runner gitdomain.Runner) bool {
-	err := runner.Run("git", "rev-parse", "-q", "--verify", "MERGE_HEAD")
+func (self *Commands) HasMergeInProgress(querier gitdomain.Querier) bool {
+	_, err := querier.Query("git", "rev-parse", "-q", "--verify", "MERGE_HEAD")
```
kevgo, line 524: If you don't use the output returned by `querier.Query`, then you can also keep this as a `runner.Run`.

internal/git/commands.go
```diff
@@ -519,11 +520,26 @@ func (self *Commands) HasLocalBranch(runner gitdomain.Runner, name gitdomain.Loc
...
 	return err == nil
 }
 
+func (self *Commands) HasRebaseInProgress(querier gitdomain.Querier) bool {
+	gitDir, err := querier.QueryTrim("git", "rev-parse", "--absolute-git-dir")
+	if err != nil {
+		return false
+	}
+	for _, rebaseDirName := range []string{"rebase-merge", "rebase-apply"} {
```
kevgo, line 533: This is cool that it detects a rebase by looking at the files that the user needs to edit when doing a rebase. Let's hope that these file names don't change in future Git versions.

kevgo (approved):
This is a really great improvement! We already use zero bytes when loading the Git configuration data. Definitely a good idea to use this in other places as well.

Please note that if verbose mode is enabled, the terminal output will [print two newlines instead of the zero byte](https://github.com/git-town/git-town/blob/961eac19f65cb5d4bafd88f48c19e4bc149b4774/internal/subshell/backend_runner.go#L79) for better readability.

I leave it up to you to merge this when you are ready.

internal/git/commands.go
stephenwade: Since I need to call it from `RepoStatus` which only has a `Querier`, it was easier to make this function also take a `Querier`.

internal/git/commands.go
stephenwade: It's weirdly difficult to tell if a rebase is in progress! I tried looking for `REBASE_HEAD` at first, but it still exists after a rebase is finished. Checking for the presence of the directories `.git/rebase-merge` or `.git/rebase-apply` is how https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh detects if a rebase is in progress.

I think it's really unlikely that this will change in the future because it would break other clients who use this method to detect a rebase. I ran into this PR from VS Code where they also check these folders: https://github.com/microsoft/vscode/pull/57651/files

internal/git/commands.go
kevgo: Okay, thanks for the background. Good as is then.

kevgo: On second thought... Can you update `RepoStatus` to receive a [RunnerQuerier](internal/git/gitdomain/runner.go) instead of a `Querier`? Then you should be able to call `HasMergeInProgress` with a `Runner`. `RunnerQuerier` implements both the `Runner` and the `Querier` interface.

stephenwade: https://github.com/git-town/git-town/pull/4755

PR #4751, opened 2025-04-22
Generate opcodes list with `make fix`

stephenwade:
This PR adds a few missing opcodes to the opcodes package.

I noticed that `BranchDeleteIfEmptyAtRuntime` was missing because I got an error when running `git town continue` today. I decided to write a quick and dirty generator, which added the following opcodes to the list:

- `BranchDeleteIfEmptyAtRuntime`
- `CherryPick`
- `MergeFastForward`
- `MergeParent`
- `MergeSquashAutoUndo`


tools/generate_opcodes_all.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# This script generates internal/vm/opcodes/all.go.
+
+set -e
+
+temp_file=$(mktemp)
+
+(
+	# shellcheck disable=SC2016
+	echo '// Generated by generate_opcodes_all.sh. Run `make fix` to update. DO NOT EDIT.'
```
kevgo, line 11: You could consider replacing the repeated echo statements with a [heredoc](https://linuxize.com/post/bash-heredoc). That said, the current version is quite readable, totally fine to leave as-is if you prefer.

tools/generate_opcodes_all.sh
```diff
@@ -0,0 +1,27 @@
...
+	echo '// All provides all existing opcodes.'
+	echo '// This is used to iterate all opcode types.'
+	echo 'func All() []shared.Opcode {'
+	echo '	return []shared.Opcode{'
+	cat internal/vm/opcodes/*.go | grep '^type .* struct .*$' | sed -e 's/type /\t\t\&/' -e 's/ .*/{},/'
+	echo '	} //exhaustruct:ignore'
+	echo '}'
+) >"$temp_file"
+
+mv "$temp_file" internal/vm/opcodes/all.go
```
kevgo, line 26: Curious, why not update `internal/vm/opcodes/all.go` directly? I don't see any risks of it ending up in an inconsistent state, so wondering if there's a reason for going the detour through the temp file.

tools/generate_opcodes_all.sh
```diff
@@ -0,0 +1,27 @@
...
+	echo
+	echo 'package opcodes'
+	echo
+	echo 'import "github.com/git-town/git-town/v19/internal/vm/shared"'
+	echo
+	echo '// All provides all existing opcodes.'
+	echo '// This is used to iterate all opcode types.'
+	echo 'func All() []shared.Opcode {'
+	echo '	return []shared.Opcode{'
+	cat internal/vm/opcodes/*.go | grep '^type .* struct .*$' | sed -e 's/type /\t\t\&/' -e 's/ .*/{},/'
```
kevgo, line 21:
```suggestion
	cat internal/vm/opcodes/*.go | grep '^type .* struct .*$' | sed -e 's/type /\t\t\&/' -e 's/ .*/{},/' | sort
```

On my machine it generates the opcodes in a non-alphabetic order. Sorting them eliminates this problem.

image

kevgo (approved):
Massive thanks for doing this! Maintaining this list of opcodes manually has been an ongoing headache, and I constantly forget to update it. Really good idea to automate this! :100:

tools/generate_opcodes_all.sh
stephenwade: I prefer the echo statements so I'll leave it as-is 🙂

tools/generate_opcodes_all.sh
stephenwade: I think I was trying to be extra cautious, but I agree that it's not necessary. I'll remove the detour.

tools/generate_opcodes_all.sh
stephenwade: Done.

stephenwade: Turns out, Ubuntu in the dev container and Ubuntu on GitHub Actions `sort` differently! I generated the file first in the dev container, but when I run the same script from Terminal.app, I get the same result as GitHub and as you.

I added an option to use the most basic collation (`LC_COLLATION=C sort`) which makes them behave the same way.

PR #4757, opened 2025-04-23
Use rev-parse instead of show-ref

stephenwade:
Either works, but `rev-parse` feels more accurate. `show-ref`'s main purpose is to show all the refs that match a string, although it requires exact matches with `--verify`. `rev-parse`'s purpose is to look up details of a reference.

https://stackoverflow.com/a/44994778


kevgo:
I agree, thanks for improving this! :+1:

PR #4758, opened 2025-04-23
Update CurrentBranchDuringRebase to avoid parsing `git branch`

stephenwade:


internal/git/commands.go
```diff
@@ -852,20 +858,25 @@ func (self *Commands) UnstageAll(runner gitdomain.Runner) error {
 }
 
 func (self *Commands) currentBranchDuringRebase(querier gitdomain.Querier) (gitdomain.LocalBranchName, error) {
-	output, err := querier.QueryTrim("git", "branch", "--list")
```
kevgo, line 855: Good idea to extract this into its own method!

internal/git/commands.go
```diff
@@ -852,20 +858,25 @@ func (self *Commands) UnstageAll(runner gitdomain.Runner) error {
...
+	gitDir, err := self.gitDirectory(querier)
 	if err != nil {
-		return "", err
+		return gitdomain.LocalBranchName(""), err
```
kevgo, line 863: In Go, constants without type are untyped. When using an untyped constant expression, it takes on the type of the variable it is assigned to. So `""` will result in a `gitdomain.LocalBranchName("")` here. https://go.dev/blog/constants provides some background.

internal/git/commands_test.go
```diff
@@ -1410,29 +1410,4 @@ func TestBackendCommands(t *testing.T) {
...
-			must.Eq(t, want, have)
-		})
-		t.Run("branch name is three words", func(t *testing.T) {
-			t.Parallel()
-			give := "* (no branch, rebasing the feature branch)"
-			have := git.ParseActiveBranchDuringRebase(give)
-			want := gitdomain.NewLocalBranchName("the feature branch")
-			must.Eq(t, want, have)
-		})
-	})
```
kevgo, line 1437: I just realized there's a cleaner way to test this, and would also validate your new implementation. I just shipped [#4760](https://github.com/git-town/git-town/pull/4760) with these improvements.

I'm holding off on syncing your branch myself to avoid interfering with your changes. When you do pull in the latest main, you might hit some merge conflicts. You should be able to keep all your production code changes and safely discard mine during the merge.

Just a heads up: the `currentBranchDuringRebase` method is now capitalized to make it testable, so it moved to a different spot in the file.

kevgo (approved):
This is a really nice improvement! I'm very glad you were able to get rid of my super hacky parsing of complex porcelain output. :fire:

PR #4759, opened 2025-04-23
Replace `git branch` in test code

stephenwade:


pkg/asserts/branch_exists.go
```diff
@@ -9,9 +9,8 @@ import (
 
 func BranchExists(t *testing.T, dir, expectedBranch string) {
 	t.Helper()
-	cmd := exec.Command("git", "branch")
+	cmd := exec.Command("git", "rev-parse", "--verify", "-q", "refs/heads/"+expectedBranch)
```
kevgo, line 12: To fix the linter warning:

```suggestion
	cmd := exec.Command("git", "rev-parse", "--verify", "-q", "refs/heads/"+expectedBranch) //nolint:gosec
```

stephenwade:
@kevgo Please let me know if this approach is okay or if I should return data in a different structure.

internal/test/commands/test_commands.go
```diff
@@ -355,28 +355,50 @@ func (self *TestCommands) LineageTable() datatable.DataTable {
...
 // LocalBranches provides the names of all branches in the local repository,
 // ordered alphabetically.
-func (self *TestCommands) LocalBranches() (gitdomain.LocalBranchNames, error) {
-	output, err := self.QueryTrim("git", "branch")
+func (self *TestCommands) LocalBranches() (gitdomain.LocalBranchNames, gitdomain.LocalBranchNames, error) {
+	forEachRefFormat := strings.Join(
+		[]string{
+			"%(refname:lstrip=2)", // the branch name (without refs/heads/)
+
+			" ", // a literal space
```
kevgo, line 363: Brilliant idea to use literate programming style here to document the various parts of this complex string. A possible simplification is to use string concatenation:

```suggestion
	forEachRefFormat := "" +
		"%(refname:lstrip=2)" + // the branch name (without refs/heads/)

		" " + // a literal space
```

kevgo (commented):
Thanks so much for getting rid of the porcelain commands! They got us where we are now, but using plumbing commands is so much better. I hope that get to a point where we can remove the ugly override to [always use us_en](https://github.com/git-town/git-town/blob/a24520790624299535e27726f57a8a0dbe30bede/internal/subshell/backend_runner.go#L54) in backend Git commands, so that our users can see verbose Git output in their native language!

internal/test/commands/test_commands.go
```diff
@@ -355,28 +355,50 @@ func (self *TestCommands) LineageTable() datatable.DataTable {
 
 // LocalBranches provides the names of all branches in the local repository,
 // ordered alphabetically.
-func (self *TestCommands) LocalBranches() (gitdomain.LocalBranchNames, error) {
-	output, err := self.QueryTrim("git", "branch")
+func (self *TestCommands) LocalBranches() (gitdomain.LocalBranchNames, gitdomain.LocalBranchNames, error) {
```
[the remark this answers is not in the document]
stephenwade: TIL about named returns. I'll use that to make the function signature clearer.

internal/test/commands/test_commands.go
```diff
@@ -355,28 +355,50 @@ func (self *TestCommands) LineageTable() datatable.DataTable {
...
+	forEachRefFormat := strings.Join(
+		[]string{
+			"%(refname:lstrip=2)", // the branch name (without refs/heads/)
+
+			" ", // a literal space
+
+			// "+" if the branch is checked out in another worktree
+			"%(if)", "%(HEAD)", "%(then)", // If the branch is checked out in the current worktree
+			// show nothing
```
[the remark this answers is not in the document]
stephenwade: Thanks for the suggestion! I don't know why I didn't think to put the indicator first; that does indeed make parsing simpler.

Branch names are more constrained and can't contain control characters or whitespace, so splitting on newlines and spaces is safe here (since I don't actually output the worktree path, only a single letter if a worktree path is present).

internal/test/commands/test_commands.go
stephenwade: I would prefer to keep the current approach. To me, `, //` looks nicer than `+ //`.

stephenwade:
> I hope that we get to a point where we can remove the ugly override to [always use us_en](https://github.com/git-town/git-town/blob/a24520790624299535e27726f57a8a0dbe30bede/internal/subshell/backend_runner.go#L54) in backend Git commands, so that our users can see verbose Git output in their native language!

My goal is to be able to remove that and the `color.ui=never` preference so that Git Town is resilient to however the user wants to configure Git. 😄

internal/test/commands/test_commands.go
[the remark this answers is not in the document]
kevgo: They are usually not a good idea since they introduce problems. But in this case they seem a good fit.

internal/test/commands/test_commands.go
kevgo: No problem!

internal/test/commands/test_commands.go
```diff
@@ -355,28 +355,52 @@ func (self *TestCommands) LineageTable() datatable.DataTable {
...
-		line = strings.Trim(line, "* ")
-		line = strings.TrimSpace(line)
-		result = append(result, gitdomain.NewLocalBranchName(line))
+		marker := line[0]
+		branch := line[1:]
+		switch marker {
+		case 'H':
+			fallthrough
+		case ' ':
+			allBranches = append(allBranches, gitdomain.NewLocalBranchName(branch))
```
kevgo, line 385: You can group multiple case statements together like this:

```suggestion
		case 'H', ' ':
			allBranches = append(allBranches, gitdomain.NewLocalBranchName(branch))
```

internal/test/commands/test_commands.go
```diff
@@ -355,28 +355,52 @@ func (self *TestCommands) LineageTable() datatable.DataTable {
...
+		case 'W':
+			allBranches = append(allBranches, gitdomain.NewLocalBranchName(branch))
```
kevgo, line 387: The line `allBranches = append(allBranches, gitdomain.NewLocalBranchName(branch))` exists in all case statements. You can extract it to above the switch statement.

kevgo (approved):
Very nicely done! I am pretty excited about using these plumbing commands everywhere in the backend. This makes Git Town much more idiomatic, robust, and stable in the long term!

PR #4785, opened 2025-04-24
Improve unit test for BranchesSnapshot

stephenwade:
I'm working on replacing the call to `git branch` in BranchesSnapshot. As part of that process, and inspired by #4760, I rewrote the unit tests that target ParseVerboseBranchesOutput to target BranchesSnapshot instead.

~~@kevgo I'm getting test failures on "branch is deleted at the remote", and I'm not sure how to trigger that status in a unit test. I would appreciate your review to see if I am taking the right approach, if there are things I should improve, and to help fix the test failure. I'll finish rewriting the tests after this feedback.~~


kevgo (commented):
Thanks for giving the unit tests some attention and for flagging the blocker! Your test code looks solid and idiomatic, nicely done!

To help reproduce the issue, here’s a Bash script that simulates what the unit tests do:

```bash
mkdir test
cd test

# create origin repo
git init origin
cd origin
git commit --allow-empty -m "Initial commit"

# create branch "branch"
git checkout -b branch
touch file
git add .
git commit -m "Branch commit"

# create local clone
cd ..
git clone origin local
cd local
git checkout branch
echo "LOCAL REMOTES"
git remote -v

# origin deletes branch "branch"
cd ../origin
git checkout main
git branch -D branch
echo "ORIGIN AFTER BRANCH DELETED"
git branch -vva

# local sees the branch as deleted
cd ../local
echo "LOCAL BEFORE FETCH"
git branch -vva
git fetch --prune
echo "LOCAL AFTER FETCH"
git branch -vva
```

Output (cleaned up):

```
LOCAL REMOTES
origin	/home/kevlar/test/origin (fetch)
origin	/home/kevlar/test/origin (push)

ORIGIN AFTER BRANCH DELETED
* main 02780fd Initial commit

LOCAL BEFORE FETCH
* branch                c3a2aeb [origin/branch] Branch commit
  remotes/origin/HEAD   -> origin/branch
  remotes/origin/branch c3a2aeb Branch commit
  remotes/origin/main   02780fd Initial commit

LOCAL AFTER FETCH
* branch              c3a2aeb [origin/branch: gone] Branch commit
  remotes/origin/HEAD -> origin/main
  remotes/origin/main 02780fd Initial commit
```

The key takeaway here is a bit hard to spot: running `git fetch` alone doesn’t pick up that the branch got deleted, but `git fetch --prune` does. Only with this flag does Git update the local refs to show that the remote branch is gone.

internal/git/commands_test.go
```diff
@@ -82,6 +82,323 @@ func TestBackendCommands(t *testing.T) {
...
+				origin := testruntime.Create(t)
+				local := testruntime.Clone(origin.TestRunner, t.TempDir())
+				origin.CreateAndCheckoutFeatureBranch("branch", initial.Location())
+				origin.CreateCommit(testgit.Commit{
+					Branch:      "branch",
+					FileContent: "content",
+					FileName:    "file",
+					Message:     "commit on parent-only branch",
+				})
+				local.Fetch()
```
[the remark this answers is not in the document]
stephenwade: > we need a separate method that runs `git fetch --prune`

Since Git Town runs `git fetch --prune`, would it be okay to modify the `Fetch` method to use `--prune` instead of making a new method?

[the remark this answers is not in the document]
kevgo: Yes, good idea!

internal/git/commands_test.go
```diff
@@ -315,6 +315,499 @@ func TestBackendCommands(t *testing.T) {
 		})
 	})
 
+	t.Run("BranchesSnapshot", func(t *testing.T) {
+		t.Parallel()
+		t.Run("recognizes the branch names", func(t *testing.T) {
```
kevgo, line 320:
```suggestion
		t.Run("recognizes the active branch names", func(t *testing.T) {
```

internal/git/commands_test.go
```diff
@@ -315,6 +315,499 @@ func TestBackendCommands(t *testing.T) {
...
+				rebaseInProgress, err := runtime.Git.HasRebaseInProgress(runtime)
+				must.NoError(t, err)
+				must.True(t, rebaseInProgress)
+				snapshot, err := runtime.Git.BranchesSnapshot(runtime)
+				must.NoError(t, err)
+				must.Eq(t, None[gitdomain.LocalBranchName](), snapshot.Active)
+			})
+		})
+
+		t.Run("recognize the branch sync status", func(t *testing.T) {
```
kevgo, line 370:
```suggestion
		t.Run("recognizes the branch sync status", func(t *testing.T) {
```

internal/git/commands_test.go
```diff
@@ -315,6 +315,499 @@ func TestBackendCommands(t *testing.T) {
...
+			t.Parallel()
+			t.Run("branch is ahead of its remote branch", func(t *testing.T) {
+				t.Parallel()
+				origin := testruntime.Create(t)
+				origin.CreateCommit(testgit.Commit{
+					Branch:      initial,
+					FileContent: "content",
+					FileName:    "file",
+					Message:     "commit on both parent and child",
```
kevgo, line 379:
```suggestion
					Message:     "local and origin commit",
```

internal/git/commands_test.go
```diff
@@ -315,6 +315,499 @@ func TestBackendCommands(t *testing.T) {
...
+				})
+				local := testruntime.Clone(origin.TestRunner, t.TempDir())
+				local.CreateCommit(testgit.Commit{
+					Branch:      initial,
+					FileContent: "content 2",
+					FileName:    "file",
+					Message:     "commit only on child",
```
kevgo, line 386:
```suggestion
					Message:     "local commit",
```

kevgo: Just a proposal to simplify all the commit messages in the entire test segment.

internal/git/commands_test.go
```diff
@@ -315,6 +315,499 @@ func TestBackendCommands(t *testing.T) {
...
+						gitdomain.BranchInfo{
+							LocalName:  Some(initial),
+							LocalSHA:   Some(commits[0].SHA),
+							SyncStatus: gitdomain.SyncStatusLocalOnly,
+							RemoteName: None[gitdomain.RemoteBranchName](),
+							RemoteSHA:  None[gitdomain.SHA](),
+						},
+					},
+				}
+				have, _ := runtime.Git.BranchesSnapshot(runtime)
```
kevgo, line 660: I prefer to verify the error returned by the function under test.

```suggestion
				have, err := runtime.Git.BranchesSnapshot(runtime)
				must.NoError(t, err)
```

internal/git/commands_test.go
```diff
@@ -873,332 +1364,6 @@ func TestBackendCommands(t *testing.T) {
 		}
 	})
 
-	t.Run("ParseVerboseBranchesOutput", func(t *testing.T) {
```
kevgo, line 876: Since you delete the tests for `ParseVerboseBranchesOutput`, can you also delete that function?

internal/git/commands_test.go
```diff
@@ -447,14 +940,12 @@ func TestBackendCommands(t *testing.T) {
 			Branch:      "branch",
 			FileContent: "branch content",
 			FileName:    "file",
-			Locations:   testgit.Locations{testgit.LocationLocal},
```
kevgo, line 450: Could you clarify the motivation for removing this line? It seems unrelated to the other changes in this pull request.

kevgo (approved):
Thanks for adding these tests!

internal/git/commands_test.go
stephenwade: It didn't seem necessary and the tests still passed without it, so I thought to clean it up. That should probably happen in a different PR, though, so the change can be discussed on its own. I'll revert it here.

internal/git/commands_test.go
stephenwade: I'm going to delete it in another PR that changes the branch parsing. For now, `ParseVerboseBranchesOutput` is still used by `BranchesSnapshot`.

internal/git/commands_test.go
kevgo: Yes, I have [disabled](https://github.com/git-town/git-town/blob/main/Makefile#L93) the [exhaustruct](https://github.com/GaijinEntertainment/go-exhaustruct) linter for test code a while ago, to allow tests to be more concise.

There are probably a whole number of struct fields in tests that could be deleted as a result. I agree that's a whole other can of worms that is better done in a separate PR.

PR #4811, opened 2025-04-30
Use `git show -s` instead of `git log -n 1`

stephenwade:


kevgo (approved):
Looks good to me! What's the advantage of using `git show --no-patch` over `git log -n 1`? https://github.com/git/git/blob/master/command-list.txt lists both as porcelain commands.

stephenwade:
You're right, they're both porcelain commands, but also the way we use them is safe because we pass a format string.

It's a really minor nitpick, but I want to change it because `git log` is meant to show info about many commits (which we have to limit with `-n 1`). I would prefer to use `git show`, the command meant to show details about one commit.

kevgo:
Thanks for the background, makes sense to me!

PR #4824, opened 2025-05-01
Make test command CommitsInBranch more robust

stephenwade:
This PR updates test command CommitsInBranch to split the pieces of each line with a null byte instead of a pipe character.


kevgo (commented):
Git Town is becoming bulletproof! :+1:

PR #4825, opened 2025-05-01
Remove unused code

stephenwade:


kevgo:
So nice that we don't have this type of conditional behavior anymore!

PR #4735, opened 2025-04-19
Set color.ui=never when running git commands

stephenwade:
Fixes #4717.

Issue #4717, opened 2025-04-14
Panic when 'git branch' command outputs in color


terheyden:
### Steps to reproduce

MacBook Pro running Sequioa 15.4. It's a pretty fresh machine, so I was installing 'git town' on it.
When I run 'git town config setup -v' I get an error:

```
panic: "\x1b[m" is not a valid Git SHA
```

This happens both with Apple git (pre-installed) as well as with brew installed git. I think it has to do with the 'git branch' command outputting in color -- see attached screenshot. Thanks!!

![Image](https://github.com/user-attachments/assets/8c7f978d-3f7c-4cac-bc62-2bf173c107da)

### Command output

```shell
(verbose) git stash list

(verbose) git branch -vva --sort=refname
* master                          1adfd272aaae76850789b6106354492c1f05d3cb [origin/master] Merge pull request #64 from Mikrz/stop_ignoring_environment
  remotes/origin/HEAD             -> origin/master
  remotes/origin/completion       4e52e2edf8497cd8c977417d7f460f6229f2f829 Add zsh completion for fzm
  remotes/origin/demo             fe035ddd4c244193527c2d8c454c9290c3ae575a Add new demo
  remotes/origin/master           1adfd272aaae76850789b6106354492c1f05d3cb Merge pull request #64 from Mikrz/stop_ignoring_environment
  remotes/origin/revert-41-master 3168e9c9b23d2f0543dba94334718d98b9fe5ae0 Revert "added possibilty to paste a directory"
panic: "\x1b[m" is not a valid Git SHA

goroutine 1 [running]:
github.com/git-town/git-town/v18/internal/git/gitdomain.NewSHA(...)
	github.com/git-town/git-town/v18/internal/git/gitdomain/sha.go:16
github.com/git-town/git-town/v18/internal/git.ParseVerboseBranchesOutput({0x1400021e900, 0x2c1})
	github.com/git-town/git-town/v18/internal/git/commands.go:959 +0xa74
github.com/git-town/git-town/v18/internal/git.(*Commands).BranchesSnapshot(0x14000235a80, {0x104af3130, 0x14000130798})
	github.com/git-town/git-town/v18/internal/git/commands.go:116 +0x1d0
[...]

PR #4840, opened 2025-05-02
Remove unnecessary color.ui override

stephenwade:
Now that we're not parsing Git porcelain output in Git Town, we can safely run Git commands without caring about the user's `color.ui` setting.

See #4717, #4735

### DIFF
--- a/internal/subshell/backend_runner.go
+++ b/internal/subshell/backend_runner.go
@@ -52,7 +52,6 @@ func (self BackendRunner) execute(executable string, args ...string) (string, er
 	}
 	env := subProcess.Environ()
 	env = append(env, "LC_ALL=C")
-	env = append(env, `GIT_CONFIG_PARAMETERS='color.ui=never'`)
 	subProcess.Env = env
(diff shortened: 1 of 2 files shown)


internal/subshell/backend_runner.go
```diff
@@ -52,7 +52,6 @@ func (self BackendRunner) execute(executable string, args ...string) (string, er
 	}
 	env := subProcess.Environ()
 	env = append(env, "LC_ALL=C")
```
kevgo, line 54: I _think_ you can also remove this line, which sets the language to us_en for backend commands. This line was added to fix bugs like https://github.com/git-town/git-town/issues/2469.

A simple reproduction of the bug is in the ticket: `LANG=fr_FR git status`

kevgo (approved):
Woohoo!! :tada:

internal/subshell/backend_runner.go
stephenwade: I'm planning to remove this in a future PR.

PR #4841, opened 2025-05-02
Run CI on all PRs

stephenwade:


kevgo:
Just out of curiosity, on which PRs did CI not run before this change?

stephenwade:
Any PRs not against `main`. Specifically, I wanted CI to run against #4842 and #4844, which had to include changes from #4840 before it was merged.

kevgo:
I see. I had limited CI to run only for "top-level" pull requests that are about to be merged, to keep pressure on CI low. But CI is free for open-source projects, so I'm okay with using this more generic setting and see how it goes.

PR #4842, opened 2025-05-02
Run tests with color.ui=always

stephenwade:
This PR probably shouldn't be merged, but it proves that Git Town works even if the user has set `color.ui` to `always`.


kevgo (commented):
This is so cool! I think this should be an end-to-end test for each Git Town command.

This test would help prevent porcelain commands from creeping back into the codebase in the future.

And I consider being able to run with enforced Git colors a requirement for Git Town now.

Here is how to enable colors in tests:

```cucumber
And Git setting "color.ui" is "always"
```

PR #4862, opened 2025-05-05
Add tests for running all Git Town commands with `color.ui=always`

### LINKED ISSUE #4856: Add an end-to-end test for running each Git Town command with enforced colors
> This is so cool! I think this should be an end-to-end test for each Git Town command.
> 
> This test would help prevent porcelain commands from creeping back into the codebase in the future.
> 
> And I consider being able to run with enforced Git colors a requirement for Git Town now.
> 
> Here is how to enable colors in tests:
> 
> ```cucumber
> And Git setting "color.ui" is "always"
> ```
> 
> _Originally posted by @kevgo in https://github.com/git-town/git-town/pull/4842#pullrequestreview-2811761002_

stephenwade:
closes #4856


stephenwade:
@kevgo Can you help me figure out the lint failure?

```
tools/rta node tools/node_modules/.bin/gherkin-lint
/home/runner/work/git-town/git-town/features/sync/current_branch/feature_branch/compress_sync_strategy/color.ui/always.feature
  24    Step "Given I add this commit to the current branch:" should use And instead of Given     use-and
```

I copied this file from features/sync/current_branch/feature_branch/compress_sync_strategy/happy_path.feature, and it works fine there (and in 5 other files), but it fails in my new test.

kevgo:
@stephenwade I have fixed it in https://github.com/git-town/git-town/pull/4862/commits/699b5368de79bd105a3206dcb8d27e6fe9151b66.

The `gherkinlint` linter verifies, amongst other things, that the order of steps in `.feature` files is (1) `Given`, (2) `When`, and finally (3) `Then`. Which is equivalent to (1) setup the test environment, (2) run the code under test, and (3) verify the changes that the code under test made.

Some tests, like the one that failed, exercise more complex usage scenarios with steps in this order: (1) (2) (3)  (1) (2) (3)  (1) (2) (3)  (1) (2) (3). The solution is to simply tell `gherkinlint` to not look at those files.

PR #4863, opened 2025-05-05
End-to-end tests for different languages

kevgo:


kevgo:
@stephenwade FYI Google Gemini and I found the right incantation to run Git in Spanish on GitHub Actions.

PR #4871, opened 2025-05-07
Support pull requests comments

Ydot19:


kevgo (commented):
Thanks for implementing this nice feature - great work! :pray:

You're definitely heading in the right direction. My main feedback at this stage is to approach this using the "spike and stabilize" method: start by getting a simple prototype working (the spike), and once that’s in place, we can refine it (stabilize) by adding the necessary checks, patterns, and types to make it production-ready.

In general, I find that good abstractions and patterns emerge naturally from working code rather than being designed upfront.

This approach should help us keep sight of the bigger picture as we iterate. I'm excited to see the next iteration of this!

kevgo:
You can run `make fix` to fix the linter errors.

internal/forge/forgedomain/proposal_comments.go
```diff
@@ -0,0 +1,49 @@
+package forgedomain
+
+type ProposalCommentSearchOptions struct {
+}
+
+type ProposalCommentSortBy string
+
+const (
+	ProposalCommentSortByCreatedAt ProposalCommentSortBy = "created_at"
+	ProposalCommentSortByUpdatedAt ProposalCommentSortBy = "updated_at"
```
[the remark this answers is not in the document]
yaadata: Yes this sounds reasonable!

internal/config/configdomain/proposal_lineage.go
```diff
@@ -0,0 +1,13 @@
+package configdomain
+
+type ProposalLineage string
+
+const (
+	ProposalLineageNone            ProposalLineage = "none"
+	ProposalLineageComment         ProposalLineage = "comment"
+	ProposalLineageTerminalDisplay ProposalLineage = "terminal-display"
```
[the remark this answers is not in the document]
yaadata: This is good to know, I was not thinking of it from that vantage point but makes sense!

kevgo:
@Ydot19 please ping me when you want me to take another look at this. Thanks!

ceilfors:
Just wanted to say that I'm incredibly excited by this PR and looking forward to it getting merged!

kevgo:
@Ydot19 Nice work so far! Let me know if there's anything I can do to help get this PR across the finish line.

yaadata:
@kevgo Sorry I have been going through a rush at work / life but it is my focus to get this feature complete by July 4th weekend.

I last touched this PR about 2 weeks go but I want to break down my thought process. Below are sections
- Realizations (what I have learned from opening up this pull-request)
- Questions (questions I have for you @kevgo)
- Design (my current design approach)
- Next Steps

to help parse this large response

### Realizations

- Displaying the lineage on a proposal is best done on a proposal comment. The Proposal Body is not a good location to display this information because if you create future child proposals at a later time, it is hard to update and won't play nice for organizations using proposal templates for the proposal body.
> This brings up interesting challenges of how to reliably update the same comment if it still exists.

- Displaying the lineage on proposal creation is not a reasonable workflow because creating a proposal relies on opening up the browser and completing the flow on the browser. There is no hook mechanism for listening to proposal. The approach I am considering is moving the ability to create the proposal comment as a part of the git-town sync command, initially behind a flag called `-l` (--lineage)

- Tangentially, we should be displaying the same lineage information on the terminal (something like `git-town branch -p`, shows the lineage as it would appear in a proposal). I am currently considering this as the v0 to prove that lineage information is correctly being generated. This I want to introduce in a separate pull-request.


### Current Questions:

- This method seems to be the preferred way to get branch lineage. Let me know if there is another way that is preferred 
```golang
func (self Lineage) BranchLineageWithoutRoot(branch gitdomain.LocalBranchName, perennialBranches gitdomain.LocalBranchNames) gitdomain.LocalBranchNames {
	if self.Parent(branch).IsNone() {
		result := gitdomain.LocalBranchNames{}
		if !perennialBranches.Contains(branch) {
			result = append(result, branch)
		}
		return append(result, self.Descendants(branch)...)
	}
	return append(append(self.AncestorsWithoutRoot(branch), branch), self.Descendants(branch)...)
}
```
- How do you propose we store the comment id containing the lineage comment for updates to lineage information. I am currently going to implement a 'dumb' approach, which is to store a hidden mark down text and search the first 10 comments on a proposal for this text (see the design section for more on this). If we find this hidden text in a comment, we will update the comment; otherwise, we will create a new lineage comment

### Design

#### Commands

```zsh
git-town sync -p
```
Purpose: Syncs and creates the lineage comment in proposals

```zsh
git-town sync -s -p
```
Purpose: Syncs and creates the lineage comment across the entire stack of proposals

```zsh
git-town branch -p
```
Shows the lineage comment in the terminal for existing proposals. If a branch has no proposal, it will show that there are no proposals for the branch

#### Comment Format:

```txt
<!---GIT-TOWN GENERATED--->
This proposal is part of a stack. 

main
    ↳ branch_name_a ➜  [#prosopalID](link to proposal)
        ↳  branch_name_b ➜  [#prosopalID](link to proposal)
            ↳ branch_name_c ➜  [#prosopalID](link to proposal)  ***CURRENT*** 
```

NOTE: 
- The current surrounded by *** denotes that the reader is viewing that proposal
- The hidden text is used to the existing lineage comment if we need to update the information
- If there is a branch that exists locally but there is no proposal for it, it will not appear in the proposal comment. Example: take the comment above, if there is also a branch called `branch_name_d` that exists locally but no proposal exists on the forge, then it will not appear in the comment



### Test Cases 

These are the cases I am considering prior to considering this feature ready. Let me know if there are others
- Generating the base case (see the design section). This is the happy case where a proposal and its entire lineage of proposals are created at the same time.
- Generate comments when there is a perennialBranch in the lineage. If there is a perennial Branch, the perennial branch will not have a proposal linked to it in the lineage comment

Example:
```txt
<!---GIT-TOWN GENERATED--->
This proposal is part of a stack. 

main
    development (perennial branch)
       ↳ branch_name_a ➜  [#prosopalID](link to proposal)
           ↳  branch_name_b ➜  [#prosopalID](link to proposal)
               ↳ branch_name_c ➜  [#prosopalID](link to proposal)  ***CURRENT*** 
```

- Update lineage comments when new proposals are created that are part of the lineage of branch. For example, say we have two branches (part_a and part_b) that are created and you are currently viewing part_b's proposal
BEFORE:
```txt
<!---GIT-TOWN GENERATED--->
This proposal is part of a stack. 

main
    ↳ part_a ➜  [#prosopalID](link to proposal)
        ↳  part_b ➜  [#prosopalID](link to proposal) ***CURRENT*** 
```

If a part_c proposal is created at a later time, running `git-town sync -p` should update this comment to be the following

AFTER:
```txt
<!---GIT-TOWN GENERATED--->
This proposal is part of a stack. 

main
    ↳ part_a ➜  [#prosopalID](link to proposal)
        ↳  part_b ➜  [#prosopalID](link to proposal) ***CURRENT*** 
            ↳ part_c ➜  [#prosopalID](link to proposal) 
```

Similarly, if you update the entire stack, the lineage comment for part_a should also be updated

- Not displaying branches in the lineage comment that do not have a proposal


### Next Steps:
- This pull-request will focus only on enabling `git-town sync -p` for the initial comment creation
- Separate pull-request to enable updating an existing lineage comment on a proposal
- Separate pull-request:  Enabling `git-town sync -s -p`
- Separate pull-request: Enabling `git-town branch -p`

kevgo:
@Ydot19 thanks for the update, and for all your great work and thoughtful ideas on this!

I recommend aligning your implementation as closely as possible with how the [Git Town GitHub Action](https://github.com/git-town/action) already does this. There are a few strong reasons for doing so:

- The team maintaining the GitHub Action has already explored and iterated on multiple approaches before landing on the current design. Leveraging their experience lets us stand on the shoulders of giants here.

- It ensures a consistent UX and interoperability. Whether somebody uses this feature via the Git Town CLI or through the GitHub Action, the lineage formatting and behavior should be the same. For example, if a team isn't using the GitHub Action yet, but someone starts including the lineage in proposals through the CLI, and later the team adopts the action, the transition should be seamless without visual or behavioral differences.

I think that embedding the stack into the proposal body has strong UX advantages for many (most?) users, so we should try to make this happen even if it doesn't work in some edge cases. If some of those edge cases are a big enough problem, we can layer on additional functionality to make things work there as well. You are right that one such edge case are GitHub pull request templates. A couple of potential paths forward here:

- Convince the affected teams to include Git Town's invisible placeholder into their PR templates. If they find the lineage useful, they'll have no problem doing this. 
- Users manually insert the placeholder when creating proposals. Not ideal, but also not impossible.
- Optionally, when we get user feedback from the real world around this, also support displaying the stack lineage in comments.

I also really like your idea of updating the lineage during `git town sync`. That seems like a natural fit. I think a preference would be a good way to control this behavior consistently in all affected commands like `git town sync`, `git town ship`, `git town delete`, `git town append`, `git town prepend`, etc.

Let's implement the lineage functionality in each of those commands one by one (using separate proposals), so that we keep momentum and the respective PRs remain easier to review.

Finally, for the lineage data: the `Lineage.BranchLineageWithoutRoot` method returns the right structure for this use case. To include the root branch in the output (which makes sense for display purposes), consider introducing a new method that preserves it, like `Lineage.BranchLineage`.

yaadata:
@kevgo I do believe using a proposal comment similar to graphite (here is one I pulled off line as an example) is going to be the preferable approach.

The way that graphite team does it is very similar to how the git-town action works in the proposal body. I do think we should stick with proposal comments because this is already on the action's team roadmap to pivot to a comment https://github.com/git-town/action/issues/28

tranhl:
@y

tranhl:
@Ydot19 

> Displaying the lineage on a proposal is best done on a proposal comment. 

I don't necessarily agree with this. Each team is unique with their own set of processes & preferences, so it's ultimately up to them to decide what is "best". The plan for https://github.com/git-town/action/issues/28 is to support both the proposal body and comment as a visualisation target via config -- IMO the CLI should do the same.

### Notes on Design

I think the key design challenge when syncing lineage via CLI will be aligning forge lineage state with local lineage stage. Proposals can be modified/merged/closed independently from Git Town, which means the local lineage may not represent the actual state of the repository. This means that the forge is ultimately the source of truth when it comes to lineage data. If we want to sync lineage visualisations via CLI, the solution we come up with will need to take this into account.

Tangential to aligning forge & local state, another design challenge will be ensuring our solution works when multiple contributors work on the same stack/hierarchy. Storing comment IDs locally may work with a sole contributor, but depending on how it's implemented this may not adapt correctly for multiple contributors. In my experiencing, leveraging the forge as the source of truth is the usually the best bet here. My first instinct in this scenario would be to pull proposal & comment data upfront before executing a sync and then search for the invisible anchor in order to locate the visualisation.

I also question the DX improvement syncing via CLI would provide 🤔. Pulling forge data in order to generate correct lineage visualisations is going to introduce additional networking overhead which slows down syncing process. A big advantage of Git Town is how snappy due to it mostly wrapping the Git CLI. Some questions I find myself asking here are:

- Would our efforts perhaps be better directed towards creating forge-specific integrations (like the GitHub Action) that offload this to the forge?
- What would be the advantages of implementing this via CLI instead?

yaadata:
The GitHub actions approach works if you are the owner of the repository or on a team/org that is okay with updating there PR template for this use case.

In my mind, the graphite approach allows individuals to gently introduce stacking to their proposals before getting buy-in about the tool via the comments approach.

Ultimately, I won't die on this hill. If we want consistency I can do exactly what the action does via cli OR if we don't want this functionality in git-town I can also forego this functionality

tranhl:
It seems like I've come across as against syncing via CLI overall -- that was not my intention!  The need for a non-invasive alternative to the GitHub Action is pretty clear. What I'm really saying is that the solution we come up with needs to be 1. behave correctly and 2. introduce minimal performance impact to the core Git Town experience.

I don't think 1. can be achieved without pulling forge data directly, but doing this without impacting 2. is not clear to me yet.

kevgo:
> Forge vs CLI for updating stack lineage embedded in proposals

I agree that a forge-based solution - like our GitHub Action - is the preferred approach, and it's what Git Town recommends when feasible. The CLI-based approach is for situations where setting up the GitHub Action isn't possible: when decision-makers aren't (yet) on board, or when the team uses a forge that doesn't (yet) have a supported CI job.

The CLI-based stack lineage in proposals is also a natural way to raise visibility of Git Town in teams who aren't familiar with it.

> Stack lineage in proposal body vs. comments

There are valid use cases for either approach, and ideally we should support both. https://github.com/git-town/action/issues/28 is about _adding_ support for displaying stack lineage in a comment. It doesn't _remove_ the existing behavior of including it in the PR body.

> Updating lineage embedded in proposals via CLI requires (1) accurate lineage info from the forge, and (2) introduces additional API calls that may slow down the CLI

Totally agree. One way to handle this might be to configure which approach the team uses via a configuration setting. That way, teams who use the GitHub Action won't be impacted by the additional overhead. 

One the other hand, keeping the local Git Town client up to date on lineage changes made on the forge sounds like a good thing in general...

yaadata:
Thank you for the clarity. I will revise this proposal to update the proposal body similar to the github action for functional and ui parity.

Let me work through a functioning prototype then we can discuss the more thorny topics/issues with this proposal

yaadata:
@kevgo I now have a working feature for this change. I want to add more tests but this PR is already at 800+ lines so this can be a separate changes to follow

Please feel free to pilot the changes and take it for a test drive.

yaadata:
I have more branches that build of this branch that adds more enhancements and fixes to this feature. Because I don't have access to make branches to the git-town repository I cannot show the proposals queued off this branch

I have a mirror of this pull-request on my fork https://github.com/Ydot19/git-town/pull/2 and it shows more prs ready for review

Likely the next enhancement I will make is make an update to a stack of proposals in one command.

kevgo:
You can run `make fix` to fix the linter errors.

yaadata:
@kevgo it seems that the generation script is not behaving the same on my local machine because there is no diff I see.

Please see https://github.com/git-town/git-town/pull/5283#discussion_r2236948622

yaadata:
updated and responded in the other thread https://github.com/git-town/git-town/pull/5283#discussion_r2237613394

internal/config/configdomain/lineage.go
```diff
@@ -80,6 +80,16 @@ func (self Lineage) BranchLineageWithoutRoot(branch gitdomain.LocalBranchName, p
 	return append(append(self.AncestorsWithoutRoot(branch), branch), self.Descendants(branch)...)
 }
 
+// BranchLineage method returns a complete lineage chain for a given branch, including the branch itself,
+// all its ancestors, and all its descendants.
+func (self Lineage) BranchLineage(branch gitdomain.LocalBranchName) gitdomain.LocalBranchNames {
+	if self.Parent(branch).IsNone() {
+		return append(append(gitdomain.LocalBranchNames{}, branch), self.Descendants(branch)...)
```
kevgo: Simplification:

```suggestion
		return append(gitdomain.LocalBranchNames{branch}), self.Descendants(branch)...)
```

internal/config/configdomain/lineage.go
```diff
@@ -80,6 +80,16 @@ func (self Lineage) BranchLineageWithoutRoot(branch gitdomain.LocalBranchName, p
...
+	}
+
+	return append(append(self.Ancestors(branch), branch), self.Descendants(branch)...)
+}
```
kevgo: Could you simplify this method to this one?

```suggestion
func (self Lineage) BranchLineage(branch gitdomain.LocalBranchName) gitdomain.LocalBranchNames {
	return append(append(self.Ancestors(branch), branch), self.Descendants(branch)...)
}
```

internal/config/configdomain/proposal.go
```diff
@@ -0,0 +1,8 @@
+package configdomain
+
+// indicates whether a Git Town command should sync/display the lineage of a proposal
+type Proposal bool
```
kevgo: Suggestion: name this `UpdateLineageInProposal`

internal/vm/opcodes/proposal_lineage_create.go
```diff
@@ -0,0 +1,135 @@
...
+	. "github.com/git-town/git-town/v21/pkg/prelude"
+)
+
+type ProposalLineageCreate struct {
+	Branch                  gitdomain.LocalBranchName
+	ProposalLineageIn       configdomain.ProposalLineageIn
+	undeclaredOpcodeMethods `exhaustruct:"optional"`
+}
+
+func (self *ProposalLineageCreate) Run(args shared.RunArgs) error {
```
kevgo: The purpose of opcodes is to implement the [command pattern](https://en.wikipedia.org/wiki/Command_pattern) so that Git Town can suspend and resume execution. Opcodes should not implement complex business logic themselves, but rather call business logic that exists somewhere else. Can you please extract the business logic for creating the proposal stack data from this method into the `config` package, and call that business logic here in this opcode?

internal/cmd/branch.go
```diff
@@ -11,30 +11,47 @@ import (
...
 	branchHelp = `
 Git Town's equivalent of the "git branch" command.`
 )
 
 func branchCmd() *cobra.Command {
+	addProposalLineageFlag, readProposalFlag := flags.Proposal("Display the proposal stack lineage based on the current branch")
 	addVerboseFlag, readVerboseFlag := flags.Verbose()
 	cmd := cobra.Command{
-		Use:   "branch",
+		Use:   branchCommand,
```
kevgo: Let's keep each PR laser‑focused on a single concern. This one should only add the proposal‑lineage section to proposal bodies, nothing more.

The tweak that extracts a string into a constant belongs in its own PR so we can discuss that refactor independently. I'd much rather review 10 small, focused PRs than one omnibus change set that touches 10 different things. Refactors like the constant extraction don't even need a stacked branch, a standalone branch is perfect here. Creating many different branches is super easy with Git Town :slightly_smiling_face:

internal/cmd/branch.go
```diff
@@ -11,30 +11,47 @@ import (
...
+	branchCommand = "branch"
+	branchDesc    = `
+Display hierarchy either for local branches or an existing proposal of stack.
+	`
 	branchHelp = `
 Git Town's equivalent of the "git branch" command.`
 )
 
 func branchCmd() *cobra.Command {
+	addProposalLineageFlag, readProposalFlag := flags.Proposal("Display the proposal stack lineage based on the current branch")
```
kevgo: Can you please describe why a Git Town user would want to run `git town branch --proposal`? 

I understand the motivation for including the stack lineage when proposing a branch, and for updating it during a sync. But I’m not seeing the value of the change you're making to this file. Can you please clarify the benefit?

internal/cmd/sync/cmd.go
```diff
@@ -138,6 +142,14 @@ func executeSync(cliConfig cliconfig.CliConfig, syncAllBranches configdomain.All
 	if data.remotes.HasRemote(data.config.NormalConfig.DevRemote) && data.shouldPushTags && data.config.NormalConfig.Offline.IsOnline() {
 		runProgram.Value.Add(&opcodes.PushTags{})
 	}
+
+	if proposal {
```
kevgo: Very well done! You really get how Git Town executes business logic using opcodes! :+1:

internal/cmd/sync/cmd.go
```diff
@@ -164,7 +176,7 @@ func executeSync(cliConfig cliconfig.CliConfig, syncAllBranches configdomain.All
 		Backend:                 repo.Backend,
 		CommandsCounter:         repo.CommandsCounter,
 		Config:                  data.config,
-		Connector:               None[forgedomain.Connector](),
+		Connector:               data.connector,
```
kevgo, line 225: Great catch to always provide the connector instance here!

internal/config/configdomain/proposal_stack_lineage_builder_noop.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/git-town/git-town/v21/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v21/internal/git/gitdomain"
+	. "github.com/git-town/git-town/v21/pkg/prelude"
+)
+
+// noopProposalStackLineageBuilder used by the stack lineage builder
+// when the forge connector does not support finding proposals.
+type noopProposalStackLineageBuilder struct{}
```
kevgo: You shouldn't need this noop-builder. Only build the stack lineage using the real builder if you need it.

internal/vm/opcodes/proposal_lineage_create_test.go
```diff
@@ -0,0 +1,85 @@
...
+			expected:       "Description\n\n<!-- branch-stack -->\n### New lineage\n<!-- branch-stack-end -->\n\nMore content",
+		},
+		{
+			name:           "multiple runs produce same result (idempotent)",
+			currentBody:    "Initial content",
+			lineageContent: "### Stack info",
+			expected:       "Initial content\n\n<!-- branch-stack -->\n### Stack info\n<!-- branch-stack-end -->",
+		},
+		{
+			name:           "body with marker at the end",
```
kevgo: The body doesn't have a marker at the end in this test.

internal/vm/opcodes/proposal_lineage_create.go
```diff
@@ -0,0 +1,135 @@
...
+
+		stackLineage := builder.Build(
+			configdomain.WithCurrentBranch(self.Branch),
+			configdomain.WithCurrentBranchIndicator(":point_left:"),
+			configdomain.WithIndentMarker("-"),
+			configdomain.WithProposalLineageIn(self.ProposalLineageIn),
+			configdomain.WithStringBeforeStackDisplay("\n-------------------------\n"),
+			configdomain.WithStringBeforeStackDisplay(explainationText),
+			configdomain.WithStringAfterStackDisplay("\n\nStack generated by [Git-Town](https://github.com/git-town/git-town)\n"),
+		)
```
kevgo: I think using the builder pattern here is overkill. All you use it for here is creating an instance of `proposalStackLineageBuildOptions`. You call each method exactly once to set one field of the struct. Can you instead just create the `proposalStackLineageBuildOptions` instance here directly? The advantage is that the [exhaustruct linter](https://github.com/GaijinEntertainment/go-exhaustruct) will then ensure that you provide all fields.

Something like this:

```go
stackLineage := proposalStackLineageBuildOptions{
	afterStackDisplay      []string{"\n\nStack generated by [Git-Town](https://github.com/git-town/git-town)\n"},
	beforeStackDisplay     []string{"\n-------------------------\n", explainationText}
	currentBranch          self.Branch,
	currentBranchIndicator ":point_left:",
	indentMarker           "-",
	location               self.ProposalLineageIn
}

**Review by kevgo — COMMENTED** [review]
Thanks for the great work on this! It's really coming together and getting close to being shippable!

I've already extracted and merged the connector upgrades from this PR in a separate PR.

I've left some comments with suggestions on how we can further optimize the implementation of the remaining pieces.

Please reach out if you have any questions or see things differently. Thanks!

File: internal/vm/opcodes/proposal_lineage_create.go (hunk 13 above)
Comment (yaadata) [reply internal/vm/opcodes/proposal_lineage_create.go]: yep makes sense. I have an affinity for the builder pattern in cli's but this makes sense. Will change

File: internal/cmd/branch.go (hunk 7 above)
Comment (yaadata) [reply internal/cmd/branch.go]: I will split this out and I can be convinced with not including it at all. For development it made it nice to see the output in the terminal but I will for now remove it unless we see a github issue requesting it

File: internal/vm/opcodes/proposal_lineage_create.go (hunk 6 above)
Comment (yaadata) [reply internal/vm/opcodes/proposal_lineage_create.go]: This makes sense, so should the ops command rely be something like `ProposalUpdateBody`

It shouldn't be linked to lineage but a general operation that updates the proposal body to some new content. This also makes rollbacks easier if you know both the starting and ending body contents

**Comment by kevgo** [comment]
To help get this PR shipped, I have just shipped #5291. It provides a configuration setting for how lineage should be added to proposals. If this setting is set to `configdomain.ProposalsShowLineageCLI`, then the following Git Town CLI commands should update the lineage in the proposal:
- propose
- sync
- swap
- append
- prepend
- delete
- ship
- merge
- swap

Let's implement the changes for each of these commands as a separate pull request.

I'm not convinced that there should be CLI flags that configure whether the Git Town command should update lineage in proposals or not. This leads to inconsistent behavior: 
- I create a proposal via `git town propose --lineage`
- but then when syncing I forget to add `--lineage`, and now the proposal displays an outdated lineage. 

Although a user error, this makes Git Town look bad. Good tooling should make it easy to do the right thing, and make it hard/impossible to do the wrong thing. I therefore think, in this case a config option is the right way forward.

Testing this out is easy:

```bash
git config set git-town.proposals-show-lineage cli  # enable the feature for testing
git propose                                         # creates a proposal with lineage
git prepend other-branch                            # updates the lineage in the existing proposal
git config unset git-town.proposals-show-lineage    # disable the feature
```

internal/config/configdomain/proposal.go
yaadata: are you thinking that we can get this information from the config updates that you merged this morning?

internal/config/configdomain/lineage.go
yaadata: I should be transparent here that this function doesn't actually accomplish the requirements for tracking stack lineage. I have another branch that show's the new way for tracking stack lineage that mimics what the user expects (similar to git-town branch) but tailored for proposals https://github.com/Ydot19/git-town/pull/1 (I will soon open the pr in this repository after i have addressed the other comments, I couldn't before because github doesn't allow me to stack proposals if the original branch is from my fork)


for this pr, I will add these changes but wanted to put at the back of your mind that there is an amendment coming to this function's usage

internal/vm/opcodes/proposal_lineage_create.go
kevgo: Yes, ideally there is only a `ProposalUpdateBody` opcode. It takes the new body as a parameter (via a field) and updates the proposal to have this body. Great idea!

internal/config/configdomain/proposal.go
kevgo: Yes, please use the configuration entry I created this morning.

internal/cmd/branch.go
yaadata: https://github.com/git-town/git-town/pull/4871#discussion_r2239878501

internal/config/configdomain/lineage.go
kevgo: Thanks for clarifying this. Please feel free to change this PR in any way that addresses the insights that surface while building and reviewing it.

internal/messages/en.go
```diff
@@ -204,31 +203,32 @@ Please upgrade to the new format: create.new-branch-type = "prototype"`
...
+	PerennialRegexPrompt                    = "Perennial regex: "
+	PerennialRegexResult                    = "Perennial regex: %s\n"
+	PreviousCommandFinished                 = "The previous Git Town command (%s) finished successfully.\n"
+	PreviousCommandProblem                  = "The last Git Town command (%s) hit a problem %v ago.\n"
+	ProposalLineageUnsupportedForBranchType = "Proposal stack lineage unsupported for branch type %s"
+	ProposalMultipleFromFound               = "found %d proposals for branch %q"
+	ProposalMultipleFromToFound             = "found %d proposals from branch %q to branch %q"
+	ProposalNoNumberGiven                   = "no proposal number given"
+	ProposalNoParent                        = "branch %q has no parent and can therefore not be proposed"
+	ProposalNotFoundForUpdate               = "cannot find proposal to update for branch %s\n"
```
yaadata: New message

internal/messages/en.go
```diff
@@ -332,10 +332,9 @@ and will be removed in future versions of Git Town.`
 	UnfinishedRunStateSkip                  = "Skip the current branch and continue the \"%s\" command on the next branch"
 	UnfinishedRunStateUndo                  = "Undo the previous \"%s\" command"
 	UnknownBranchType                       = "Unknown branch type: %s\n"
-
-	ValueInvalid = "invalid value for %s: %q. Please provide either \"yes\" or \"no\""
-
-	WalkAllOrStack = "please provide either --all or --stack"
-	WalkDone       = "Branch walk done."
-	WalkNoDryRun   = "there is no dry-run mode for walking through branches on your shell, please call with a command to run on each branch"
+	UpdateProposalBodyUnsupported           = "the Git Town driver for your forge does not support updating the proposal body"
```
yaadata, line 334: New message

internal/vm/opcodes/proposal_update_body.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"errors"
+
+	"github.com/git-town/git-town/v21/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v21/internal/messages"
+	"github.com/git-town/git-town/v21/internal/vm/shared"
+)
+
+type ProposalUpdateBody struct {
```
yaadata, line 11: In a follow up proposal, I will add the ability to undo the proposal body update

internal/vm/opcodes/proposal_lineage_create_test.go
yaadata: updated in another set of tests after refactor

internal/vm/opcodes/proposal_lineage_create.go
yaadata: updated

internal/config/configdomain/proposal_stack_lineage_builder_noop.go
yaadata: done

internal/cmd/sync/cmd.go
yaadata: Now I use the CliConfig

internal/cmd/branch.go
yaadata: done!

yaadata:
@kevgo want to give this another look?

yaadata:
>  
```zsh
git config set git-town.proposals-show-lineage cli  # enable the feature for testing
git propose                                         # creates a proposal with lineage
git prepend other-branch                            # updates the lineage in the existing proposal
git config unset git-town.proposals-show-lineage    # disable the feature
```

~do you have a special setup that you're running locally because I don't have access to setting config~

```zsh
~/repositories/git-town on  support-pull-requests-comments [$] 
🕙[ 10:19:00 PM ] ❯ git-town config set git-town.proposals-show-lineage cli

Error: unknown command "set" for "git-town config"

```

~I was able to test by remove a conditional if on this particular config value~



EDIT:
I see that you made your git-town commands available to git and this are the git tags themselves

kevgo:
It's `git config set git-town.proposals-show-lineage cli`, without "town".

internal/cmd/sync/cmd.go
```diff
@@ -43,6 +44,7 @@ When run on a feature branch:
 - pulls updates for the current branch
 - merges the parent branch into the current branch
 - pushes the current branch
+- updates proposal(s)
```
kevgo:
```suggestion
- updates branch lineage in proposals
```

internal/cmd/sync/cmd.go
```diff
@@ -162,6 +164,40 @@ func executeSync(args executeSyncArgs) error {
 	if data.remotes.HasRemote(data.config.NormalConfig.DevRemote) && data.shouldPushTags && data.config.NormalConfig.Offline.IsOnline() {
 		runProgram.Value.Add(&opcodes.PushTags{})
 	}
+
+	if args.cliConfig.ProposalsShowLineage.GetOrDefault() == configdomain.ProposalsShowLineageCLI {
```
kevgo: This uses the setting no matter where it comes from (CLI, env variables, Git metadata, or config file):

```suggestion
	if data.config.NormalConfig.AutoResolve.ShouldAutoResolve() {
```

internal/cmd/sync/cmd.go
```diff
@@ -162,6 +164,40 @@ func executeSync(args executeSyncArgs) error {
...
+		explainationText := "Current dependencies on/for this pull-request\n\n"
```
kevgo:
```suggestion
		explanationText := "Current dependencies on/for this pull-request\n\n"
```

internal/cmd/sync/cmd.go
```diff
@@ -162,6 +164,40 @@ func executeSync(args executeSyncArgs) error {
...
+		}
+		proposalStackLineageBuilder := configdomain.NewProposalStackLineageBuilder(&proposalStackLineageArgs)
+		if builder, hasBuilder := proposalStackLineageBuilder.Get(); hasBuilder {
+			if branchProposalData, hasBranchProposalData := builder.GetProposal(data.initialBranch).Get(); hasBranchProposalData {
+				runProgram.Value.Add(&opcodes.ProposalUpdateBody{
+					Proposal:    branchProposalData,
+					UpdatedBody: configdomain.ProposalBodyUpdateWithStackLineage(branchProposalData.Body.GetOrDefault(), builder.Build(&proposalStackLineageArgs).GetOrDefault()),
+				})
+			} else {
+				fmt.Printf(messages.ProposalNotFoundForUpdate, data.initialBranch)
```
kevgo: I think there are many legitimate reasons why a branch that gets synced doesn't have a proposal to update. How do you feel about simply not saying anything if there is no proposal?

internal/config/configdomain/proposal_body_update_with_stack_lineage_test.go
```diff
@@ -0,0 +1,188 @@
...
+				feat-a
+					feat-b
+			
+<!-- branch-stack-end -->`,
+		},
+		{
+			description: "Proposal Body with multiple line body",
+			currentBody: `
+Git-town is a town of Gitters.
+These Gitters are the stackers of tomorrow
```
kevgo, line 51: Haha, awesome! :1st_place_medal:

internal/config/configdomain/proposal_stack_lineage_builder.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/git-town/git-town/v21/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v21/internal/git/gitdomain"
+	. "github.com/git-town/git-town/v21/pkg/prelude"
+)
+
+func NewProposalStackLineageBuilder(args *ProposalStackLineageArgs) Option[ProposalStackLineageBuilder] {
+	if args == nil || args.Connector.IsNone() {
+		return None[ProposalStackLineageBuilder]()
+	}
```
kevgo: If you don't use a pointer here, you don't need to defensively check against missing arguments. Would this code work?

```suggestion
func NewProposalStackLineageBuilder(args ProposalStackLineageArgs) Option[ProposalStackLineageBuilder] {
	if args.Connector.IsNone() {
		return None[ProposalStackLineageBuilder]()
	}
```

internal/config/configdomain/proposal_stack_lineage_args.go
```diff
@@ -0,0 +1,39 @@
...
+	// IndentMarker
+	// Controls the marker following an indent.
+	IndentMarker string
+	// Lineage
+	// The lineage data for the current repository
+	Lineage Lineage
+	// MainAndPerennialBranches
+	// These branches will not be searched for having a proposal attached to them
+	// when building the stack lineage.
+	MainAndPerennialBranches Option[gitdomain.LocalBranchNames]
```
kevgo: What is the motivation for making this an option? It's not an optional value IMO. If there was neither a main nor perennial branches, this could be an empty list. There is always a main branch. Removing the optionality here also simplifies your code that uses this field. 

```suggestion
	MainAndPerennialBranches gitdomain.LocalBranchNames
```

internal/config/configdomain/proposal_stack_lineage_builder.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/git-town/git-town/v21/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v21/internal/git/gitdomain"
+	. "github.com/git-town/git-town/v21/pkg/prelude"
+)
+
+func NewProposalStackLineageBuilder(args *ProposalStackLineageArgs) Option[ProposalStackLineageBuilder] {
+	if args == nil || args.Connector.IsNone() {
```
kevgo: Rather than checking `IsNone` here, and then accessing this field using `GetOrPanic` on line 22, I suggest you properly unpack this optional value, return early if there is no connector, and then use the unpacked value on line 22.

internal/config/configdomain/proposal_stack_lineage_builder.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (self *ProposalStackLineageBuilder) Build(args *ProposalStackLineageArgs) Option[string] {
+	var builder strings.Builder
+	for _, text := range args.BeforeStackDisplay {
+		builder.WriteString(text)
+	}
+
+	length := len(self.orderedLineage)
+	for i := len(self.orderedLineage); i > 0; i-- {
```
kevgo:
```suggestion
	length := len(self.orderedLineage)
	for i := len(length); i > 0; i-- {
```

internal/config/configdomain/proposal_stack_lineage_builder.go
```diff
@@ -0,0 +1,131 @@
...
+	branch   gitdomain.LocalBranchName
+	proposal Option[forgedomain.ProposalData]
+}
+
+type ProposalStackLineageBuilder struct {
+	mainAndPerennialBranches gitdomain.LocalBranchNames
+	orderedLineage           []*proposalLineage
+}
+
+func (self *ProposalStackLineageBuilder) Build(args *ProposalStackLineageArgs) Option[string] {
```
kevgo: Return a non-optional string here. This function never returns the `None` variant.

```suggestion
func (self *ProposalStackLineageBuilder) Build(args *ProposalStackLineageArgs) string {
```

internal/config/configdomain/proposal_stack_lineage_builder.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (self *ProposalStackLineageBuilder) GetProposal(branch gitdomain.LocalBranchName) Option[forgedomain.ProposalData] {
+	response := None[forgedomain.ProposalData]()
+	for _, curr := range self.orderedLineage {
+		if curr.branch == branch {
+			response = curr.proposal
+		}
+	}
+	return response
```
kevgo: Can you return early here?

```suggestion
	for _, curr := range self.orderedLineage {
		if curr.branch == branch {
			return curr.proposal
		}
	}
	return None[forgedomain.ProposalData]()
```

internal/config/configdomain/proposal_stack_lineage_builder.go
```diff
@@ -0,0 +1,131 @@
...
+) (*ProposalStackLineageBuilder, error) {
+	if self.mainAndPerennialBranches.Contains(childBranch) || parentBranch.IsNone() {
+		self.orderedLineage = append(self.orderedLineage, &proposalLineage{
+			branch:   childBranch,
+			proposal: None[forgedomain.ProposalData](),
+		})
+		return self, nil
+	}
+
+	parent := parentBranch.GetOrPanic().BranchName().LocalName()
```
kevgo: Please try to avoid such panics at runtime. Git Town should never panic. Rather than exploding, what should Git Town do in case there is no parent branch here?

internal/config/configdomain/proposal_stack_lineage_builder.go
```diff
@@ -0,0 +1,131 @@
...
+	proposal, err := findProposalFn(childBranch, parent)
+	if err != nil {
+		return self, fmt.Errorf("failed to find proposal for branch %s: %w", childBranch, err)
+	}
+
+	proposalData, hasProposal := proposal.Get()
+	if !hasProposal {
+		return self, fmt.Errorf("no proposal found branch %q", childBranch)
```
kevgo: It's not an error here if a branch has no proposal. What should Git Town do when displaying the branch lineage and one of the branches has no proposal?

internal/config/configdomain/proposal_stack_lineage_builder_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/git-town/git-town/v21/internal/config/configdomain"
+	"github.com/git-town/git-town/v21/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v21/internal/git/gitdomain"
+	"github.com/git-town/git-town/v21/internal/subshell/subshelldomain"
+	. "github.com/git-town/git-town/v21/pkg/prelude"
+)
+
+type mockConnectorProposalStackLineageBuilder struct{}
+
+var _ forgedomain.Connector = (*mockConnectorProposalStackLineageBuilder)(nil)
```
kevgo: What does this unnamed variable do?

internal/config/configdomain/proposal_stack_lineage_builder_test.go
```diff
@@ -0,0 +1,173 @@
...
+func TestProposalStackLineageBuilder_CheckLineageAndProposals(t *testing.T) {
+	t.Parallel()
+	// arrange
+	mainBranch := gitdomain.NewLocalBranchName("main")
+	perennialBranch := gitdomain.NewLocalBranchName("development")
+	featureBranch := gitdomain.NewLocalBranchName("git-town/proposal-stack-lineage")
+	lineage := configdomain.NewLineage()
+	lineage.Root(mainBranch)
+	lineage.Set(perennialBranch, mainBranch)
+	lineage.Set(featureBranch, perennialBranch)
```
kevgo: Easier way to create a prepopulated lineage:

```suggestion
	lineage := configdomain.NewLineageWith(configdomain.LineageData{
		perennialBranch: mainBranch, 
		featureBranch: perennialBranch,
	})
```

Note that `perennialBranch: mainBranch` is wrong. Perennial branches have no parent branches.

internal/config/configdomain/proposal_stack_lineage_builder_test.go
```diff
@@ -0,0 +1,173 @@
...
+   - development
+     - PR https://www.github.com/git-town/pull/1 :point_left:
+`
+
+	// act
+	actual := configdomain.NewProposalStackLineageBuilder(&args)
+
+	// assert
+	must.True(t, actual.IsSome())
+	builder := actual.GetOrDefault()
```
kevgo: Here is how you destructure the optional value properly (this follows the way the Rust language does it):

```suggestion
	builder, hasBuilder := actual.Get()
	must.True(t, hasBuilder)
```

internal/forge/gitea/connector.go
```diff
@@ -206,7 +206,7 @@ func (self Connector) squashMergeProposal(number int, message gitdomain.CommitMe
 	return err
 }
 
-func (self Connector) updateProposalBodyFn(proposalData forgedomain.ProposalInterface, updatedBody string) error {
+func (self Connector) updateProposalBody(proposalData forgedomain.ProposalInterface, updatedBody string) error {
```
kevgo: Thanks for fixing this! :pray: This change is orthogonal to the other changes in this pull request, so I'm going to extract it into its own pull request.

internal/messages/en.go
kevgo: Thanks for tagging the new messages, that's super helpful!

internal/vm/opcodes/proposal_update_body.go
```diff
@@ -0,0 +1,28 @@
...
+	Proposal                forgedomain.ProposalInterface
```
kevgo: Please use `forgedomain.Proposal` here for proper serialization to and from JSON.

```suggestion
	Proposal                forgedomain.Proposal
```

internal/vm/opcodes/proposal_update_body.go
```diff
@@ -0,0 +1,28 @@
...
+	connector, hasConnector := args.Connector.Get()
+	if !hasConnector {
+		return forgedomain.UnsupportedServiceError()
+	}
+
+	updateProposalBodyFn, hasUpdateProposalBody := connector.UpdateProposalBodyFn().Get()
+	if !hasUpdateProposalBody {
+		return errors.New(messages.UpdateProposalBodyUnsupported)
+	}
+	return updateProposalBodyFn(self.Proposal, self.UpdatedBody)
```
kevgo:
```suggestion
  updateProposalBody, canUpdateProposalBody := connector.UpdateProposalBodyFn().Get()
	if !canUpdateProposalBody {
		return errors.New(messages.UpdateProposalBodyUnsupported)
	}
	return updateProposalBody(self.Proposal, self.UpdatedBody)
```

kevgo (commented):
Very good work on this! :rocket: I think this is the last review round. I mostly have some questions and smaller suggestions. Overall this is getting good enough to get merged in, and then we can continue polishing this inside the codebase.

internal/config/configdomain/proposal_stack_lineage_builder_test.go
yaadata: This is an interface guard (or as the Go docs call it `interface checks`) https://go.dev/doc/effective_go#blank_implements

I mainly did this out of habit and I can exclude it if you feel it is not appropriate for the code base. I was reluctant to even create this mock interface but i did not see an mocks for this Connector anywhere. Typically, I use https://github.com/vektra/mockery to generate interface mocks for my tests to evaluate more complex code integrations

internal/config/configdomain/proposal_stack_lineage_builder_test.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/81fdf11c19abcc427c9ab030f7d9628b16c999a4

internal/config/configdomain/proposal_stack_lineage_builder_test.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/aa09b6972caf3641efbce7654272488e54f6702a

internal/vm/opcodes/proposal_update_body.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/c446a8890f91fcdcce32a62ab5695b71b0b219dd

internal/cmd/sync/cmd.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/12f44aeb6cb9a2ddfffe5c2b766d3cb16ba416e8

internal/cmd/sync/cmd.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/a0a21fc983979ba05711701c3afbf72730eec1cc

internal/config/configdomain/proposal_stack_lineage_args.go
yaadata: You are right there is no reason for this to be an option

https://github.com/git-town/git-town/pull/4871/commits/aa09b6972caf3641efbce7654272488e54f6702a

internal/config/configdomain/proposal_stack_lineage_builder_test.go
kevgo: I see, it verifies that `mockConnectorProposalStackLineageBuilder` implements the `forgedomain.Connector` interface. 

In Go, checking whether a class implements an interface is typically done when you use the class in your actual code. This explicit check is redundant. I suggest removing it.

kevgo: There hasn't been a strong need for mocks in Git Town so far, since the codebase cleanly separates business logic from side effects. Once this PR lands, I'd like to refactor the logic here into the following steps:

1. determine the branch lineage to display (without proposal data for now)
2. load the proposals for all involved branches concurrently (so that it's fast, syncs should be snappy)
3. add the proposal data loaded in step 2 to the lineage data created in step 1
4. now we have lineage data with proposals, render that into HTML and embed it into the proposal body

Mocks don't add much value when testing this. Each step is testable by feeding in fake input and verifying the output. Step 2 could be tested with a mock to verify that it calls `connector.SearchManyProposals` with the right branch names, but that part of the logic is extremely simple and unlikely to have bugs, or change over time. The more valuable test for step 2 seems verifying that we extract the correct set of branch names from the lineage.

internal/vm/opcodes/proposal_update_body.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/46838e8c293c2621a39b21532cc22ec64dc866e1

internal/cmd/sync/cmd.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/dac757771bfc9b902006fe94d1f7f3ae2dab142f

internal/cmd/sync/cmd.go
yaadata: makes sense. I think this was a remnant of the `--proposal` implementation and printing to the terminal if no proposal was found for updating.

https://github.com/git-town/git-town/pull/4871/commits/da584500e5a32f19e493f8b220d4a06b7a5b28cd

internal/config/configdomain/proposal_stack_lineage_builder.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/68c4bff9eb04fa8582c4111ca308157d6b89b0b1

internal/config/configdomain/proposal_stack_lineage_builder.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/b2fc772deb8bd004a9847532e1917c405662d3e8

internal/config/configdomain/proposal_stack_lineage_builder.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/0742d557f95e8a6740806772798cc2fc1d8dad7b

internal/config/configdomain/proposal_stack_lineage_builder.go
yaadata: Currently, we stop the lineage stack generation on the first non-perennial branch that does not have a proposal (http://github.com/Ydot19/git-town/blob/0742d557f95e8a6740806772798cc2fc1d8dad7b/internal/config/configdomain/proposal_stack_lineage_builder.go#L69)

For now I removed this as an error https://github.com/git-town/git-town/pull/4871/commits/e0067cc888c327469ca7c3df1ec44cb04f89409f

internal/config/configdomain/proposal_stack_lineage_builder_test.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/c9ab92bbba825d3629365d3391b00af070e1bee6

internal/config/configdomain/proposal_stack_lineage_builder.go
yaadata: https://github.com/git-town/git-town/commit/e0067cc888c327469ca7c3df1ec44cb04f89409f

internal/config/configdomain/proposal_stack_lineage_builder.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/c9f60269a75d1ff828e038a911580651ba8a7006

internal/config/configdomain/proposal_stack_lineage_builder.go
yaadata: https://github.com/git-town/git-town/pull/4871/commits/e0067cc888c327469ca7c3df1ec44cb04f89409f

yaadata:
Added a commit for each review comment to make it easier to follow. Will do one more round of manual testing tomorrow. 


Also, I don't understand why the endtoendtest are failing.


If there are other smaller tweaks or adjustments in code style needed, you can push to this branch if it helps reduce the feedback cycle

kevgo:
Alright, merging this in. Thanks so much @Ydot19 for the hard work to get this pretty complex feature built and polished. You rock!

yaadata:
@kevgo thanks, I don't know what the expectations are moving forward or whether you want to drive the rest of the changes listed here https://github.com/git-town/git-town/pull/4871#issuecomment-3132610538 by you.

```
then the following Git Town CLI commands should update the lineage in the proposal:

propose
sync
swap
append
prepend
delete
ship
merge
swap
```

If you want to drive the rest, that sounds good to me (and maybe faster). If not, I can continue contributing to drive this to completion but would like to understand the sequence as to not duplicate work you are planning to execute

kevgo:
Thanks for coordinating on this. I’m currently focused on #5156 and a few other high-priority items, so I won’t be able to work on the feature you’ve built here for a while. I'd love for you to keep pushing it forward and get it to a releasable state.

Adding proposal lineage to other Git Town commands should be much smoother now that the heavy lifting is done. You just need to hook it into the other commands.

I may contribute a few tweaks to align the code more closely with the style and architecture of the rest of the codebase. Those changes shouldn’t interfere with your work, but if they do, I’m happy to prioritize your changes to avoid merge conflicts.

Does this work for you?

yaadata:
This works for me I'll continue onward!

PR #4875, opened 2025-05-08
Properly quote outputted backend commands

stephenwade:


internal/subshell/frontend_runner.go
```diff
@@ -37,17 +38,9 @@ func FormatCommand(currentBranch gitdomain.LocalBranchName, printBranch bool, ex
 	} else {
 		result = executable + " "
 	}
-	for index, part := range args {
-		if part == "" {
-			part = `""`
-		} else if strings.Contains(part, " ") {
```
kevgo, line 43: Thanks for modernizing this code as well!

kevgo (approved):
I really like this change. Thanks for cleaning this up!

PR #4879, opened 2025-05-08
Make Spanish E2E tests work in the dev container

stephenwade:
ref: https://github.com/git-town/git-town/pull/4863


kevgo:
Nice!

PR #4911, opened 2025-05-11
Set Git editor with GIT_EDITOR instead of core.editor

stephenwade:
When Git Town rebases a branch, it tries to prevent opening an editor for commit messages by running Git with the `core.editor` config option. However, the `GIT_EDITOR` environment variable overrides the `core.editor` config option (see&nbsp;[git-var in the Git docs](https://git-scm.com/docs/git-var#Documentation/git-var.txt-GITEDITOR)).

I run into this issue regularly when I run `make cuke` in the dev container. VS Code sets the GIT_EDITOR environment variable so that commit messages open in VS Code. Because this overrides the `code.editor` config option, any test that uses the rebase sync strategy opens a new tab in VS Code.

Here's a recording of before and after when working in the dev container:

https://github.com/user-attachments/assets/3fda3318-3bec-4064-b836-e10b2029c94c

### DIFF
--- a/internal/subshell/frontend_runner.go
+++ b/internal/subshell/frontend_runner.go
@@ -51,8 +51,18 @@ func PrintCommand(branch gitdomain.LocalBranchName, printBranch bool, cmd string
 	fmt.Println(colors.Bold().Styled(header))
 }
 
-// Run runs the given command in this ShellRunner's directory.
-func (self *FrontendRunner) Run(cmd string, args ...string) (err error) {
+func (self *FrontendRunner) Run(cmd string, args ...string) error {
+	err := self.execute([]string{}, cmd, args...)
+	return err
+}
+
+func (self *FrontendRunner) RunWithEnv(env []string, cmd string, args ...string) error {
+	err := self.execute(env, cmd, args...)
+	return err
+}
+
+// runs the given command in this ShellRunner's directory.
+func (self *FrontendRunner) execute(env []string, cmd string, args ...string) (err error) {
 	self.CommandsCounter.Value.Inc()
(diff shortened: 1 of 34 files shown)


internal/subshell/frontend_runner.go
```diff
@@ -31,28 +31,40 @@ type FrontendRunner struct {
...
+	header := FormatCommand(branch, printBranch, env, cmd, args...)
 	fmt.Println()
 	fmt.Println(colors.Bold().Styled(header))
 }
 
-// Run runs the given command in this ShellRunner's directory.
-func (self *FrontendRunner) Run(cmd string, args ...string) (err error) {
+func (self *FrontendRunner) Run(cmd string, args ...string) error {
+	err := self.execute([]string{}, cmd, args...)
+	return err
```
kevgo, line 58:
```suggestion
	return self.execute([]string{}, cmd, args...)
```

internal/subshell/frontend_runner.go
```diff
@@ -31,28 +31,40 @@ type FrontendRunner struct {
...
+}
+
+func (self *FrontendRunner) RunWithEnv(env []string, cmd string, args ...string) error {
+	err := self.execute(env, cmd, args...)
+	return err
```
kevgo, line 63:
```suggestion
	return self.execute(env, cmd, args...)
```

kevgo (approved):
Thanks for this nice improvement, very well implemented!

kevgo:
This is now available as part of [Git Town 20.2](https://github.com/git-town/git-town/releases/tag/v20.2.0).

PR #4915, opened 2025-05-12
sync: don't sync with perennial branch at root in detached mode

### LINKED ISSUE #4890: detached mode should not pull in updates from the perennial root branch
When people run `git town append new --detached`, it can run `git fetch`, and it should not pull in updates from the main branch, not even from the local one. "Detached" means the branch stays as it is, no new updates. Detached should still have meaning even if common IDE or Git plugins can keep pulling updates to the local main branch in the background.  

- [x] append --detached does not pull main
- [x] append --detached does pull tracking
- [x] hack --detached does not pull main
- [x] hack --detached pulls tracking (not applicable)
- [x] prepend --detached does not pull main
- [x] prepend --detached pulls tracking
- [x] propose always runs detached
- [x] sync --detached does not pull main
- [x] sync --detached pulls tracking

kevgo:
resolves #4890

also: 

- propose is now always detached


kevgo:
This is now available as part of [Git Town 20.2](https://github.com/git-town/git-town/releases/tag/v20.2.0).

PR #5015, opened 2025-06-04
Update documentation to reflect Arch Linux packages

alerque:
I moved git-town from the AUR to official Arch Linux repositories, so the install instructions should reflect that.


kevgo:
Thanks so much, this is awesome! Really honored be in the official Arch repos now. :pray:

alerque:
N.P.

Is anything needed to deploy this? I didn't work out how the website is released, but it might be nice to have this live since it went into effect with the last release and the currently linked AUR resources will no longer be available shortly.

kevgo:
Good call. I'll push these changes up to the website today.

kevgo:
It's live now: https://www.git-town.com/install.html#linux

PR #5053, opened 2025-06-20
Use `--merge-base` in diff-parent

stephenwade:
I have a parked branch that hasn't been updated for a while. I wanted to see what changes are in that branch. When I used the diff-parent command, it showed me all the differences between the up-to-date parent and the parked child, which is 99% unrelated to the changes I made in that branch.

This PR changes the diff-parent command to use the `--merge-base` flag when calling git-diff:

[`git diff --merge-base <parent> <child>`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-codegitdiffltoptionsgt--merge-baseltcommitgtltcommitgt--ltpathgtcode)

which is equivalent to

`git diff $(git merge-base <parent> <child>) <child>`.


kevgo:
That's a really cool idea, and closer to the intent of `git diff-parent`. Thanks for implementing this!

Issue #4942, opened 2025-05-16
fix flaky sync behavior, demonstrated by flaky end-to-end tests


kevgo:
Now that production code is doing more conditional rebasing, it's behavior becomes non-deterministic due to Git's coarse timestamps (with 1 second resolution).

The problem occurs when Git Town syncs a branch stack whose oldest branch was shipped using the "rebase" strategy. This now re-creates commits over several stages, including additional rebasing to remove the shipped commits from the branches.

If all these rebases happen in the same second, the rewritten commits end up with the same SHA. If different rebases happen in different seconds, the rewritten commits end up with different SHAs. As a result, Git Town sometimes executes additional rebase operations (because it perceives differing commits in ancestor branches) and sometimes it doesn't.

This flakiness happens especially under load, i.e. on CI, but also locally.

This is also not a nice user experience, because Git Town sometimes rebases parent branches and sometimes it doesn't do that, which will trip up users who pay attention to these things.

kevgo:
Example of the problem: when running `features/sync/stack/rebase_sync_strategy/branches_gone/ancestor_shipped_and_main_updated.feature`, it sometimes executes these additional commands.

```diff
  | BRANCH   | COMMAND                                                    |
  | branch-4 | git fetch --prune --tags                                   |
  |          | git checkout main                                          |
  | main     | git -c rebase.updateRefs=false rebase origin/main          |
  |          | git push                                                   |
  |          | git checkout branch-2                                      |
  | branch-2 | git -c rebase.updateRefs=false rebase --onto main branch-1 |
  |          | git checkout branch-3                                      |
  | branch-3 | git pull                                                   |
  |          | git -c rebase.updateRefs=false rebase --onto main branch-2 |
  |          | git push --force-with-lease                                |
  |          | git checkout branch-4                                      |
  | branch-4 | git pull                                                   |
  |          | git -c rebase.updateRefs=false rebase --onto main branch-2 |
  |          | git push --force-with-lease                                |
+ |          | git -c rebase.updateRefs=false rebase branch-3             |
+ |          | git push --force-with-lease --force-if-includes            |
  |          | git branch -D branch-1                                     |
  |          | git branch -D branch-2                                     |
```

kevgo:
A possible solution would be to always execute these additional commands, whether they are needed or not. It would make sense: it's intuitive that `branch-4` should always pull in updates from its parent `branch-3`.

kevgo:
[...]

PR #5099, opened 2025-06-30
Fix casing of expected Git error message

kevgo:


stephenwade:
FYI, the result of this PR was an empty commit that also failed on old Git versions. What happened here?

kevgo:
Thanks for catching this! :pray: This PR was extracted from another, bigger change. I might have shipped these two PRs in the wrong order, leading to this change ending up empty. No harm done in this case, since this was just a tiny cleanup.

The failure on the old Git version is due to https://github.com/git-town/git-town/issues/4942. I'm still working through the best way to handle that.

Issue #5201, opened 2025-07-19
setup assistant: allow using information from global Git metadata


kevgo:
Currently the setup assistant doesn't correctly distinguish between local and global Git metadata. Change it so that it if an entry for the option exists in global Git metadata, display an option to use that global entry and not store an entry into local Git metadata.

kevgo:
This is now available in [v21.4](https://github.com/git-town/git-town/releases/tag/v21.4.0).

PR #5230, opened 2025-07-22
setup assistant: correctly remove all values

kevgo:
The setup assistant didn't remove some values when the user chooses to use the global values.

part of #5201


kevgo:

Issue #5239, opened 2025-07-22
set-parent: allow choosing the new parent via a visual dialog


kevgo:
When running without arguments, instead of displaying an error message, let the user choose the new parent via the `switch` dialog.

kevgo:
This is now available in [v21.4](https://github.com/git-town/git-town/releases/tag/v21.4.0).

PR #5259, opened 2025-07-24
set-parent: show branch list with hierarchy

kevgo:
part of #5239


kevgo:

PR #5266, opened 2025-07-24
enter parent dialog: use hierarchical branch list

### LINKED ISSUE #5239: set-parent: allow choosing the new parent via a visual dialog

kevgo:
When Git Town asks the user for the parent branch of a branch, it now displays
the hierarchical list that "git town switch" also uses. This helps maintain deep
stacks.

resolves #5239


kevgo:

PR #5274, opened 2025-07-25
Always remove old branch type overrides

kevgo:
When pruning branches that were shipped at the remote, it didn't remove their
branch type overrides. This PR fixes that.


kevgo:

PR #5317, opened 2025-08-01
Option to disable auto-resolution of phantom merge conflicts

kevgo:
part of #5279


kevgo:

PR #5329, opened 2025-08-05
enhance(internal/forge): Github forge should find Proposals regardless of state but prioritize open PRs

Ydot19:
This matches the behavior of the gh cli being able to find even closed pull-requests. This is also needed if a stack of pull-requests are not shipped all at once and you still need to sync and keep track of the proposals in the stack in the lineage comment.

### DIFF
--- a/internal/forge/github/connector.go
+++ b/internal/forge/github/connector.go
@@ -130,7 +130,7 @@ func (self Connector) findProposalViaAPI(branch, target gitdomain.LocalBranchNam
 	pullRequests, _, err := self.client.PullRequests.List(context.Background(), self.Organization, self.Repository, &github.PullRequestListOptions{
 		Head:  self.Organization + ":" + branch.String(),
 		Base:  target.String(),
-		State: "open",
+		State: "all",
 	})
 	if err != nil {
 		self.log.Failed(err.Error())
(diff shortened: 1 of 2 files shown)


internal/forge/github/connector.go
```diff
@@ -140,10 +140,28 @@ func (self Connector) findProposalViaAPI(branch, target gitdomain.LocalBranchNam
 		self.log.Success("none")
 		return None[forgedomain.Proposal](), nil
 	}
+	var pullRequest *github.PullRequest
+	var proposal forgedomain.ProposalData
 	if len(pullRequests) > 1 {
-		return None[forgedomain.Proposal](), fmt.Errorf(messages.ProposalMultipleFromToFound, len(pullRequests), branch, target)
+		// Find an open pull request. If there is more than one, return multiple found error
```
kevgo, line 146: The way this works right now, if there is exactly one pull request, it uses that one, open or closed. If it finds more than one pull request, it uses the open one. If it cannot find an open pull request, it aborts with an error.

This behavior seems inconsistent.

yaadata: This comment is inconsistent with the implementation but the logic is doing what I am expecting.

The scenarios:

1. One pull request found, return only that one
2. If there are multiple but only one open PR in that multiple, return it

The more I think about it this doesn't make 100% sense for the problem I am trying to solve. I think it would be better to only look for all PRs and return if there is only one.


To your top comment, the problem I am solving is say for example you have this stack lineage

main
    Feat-1a
         Feat-1b


And the two feature branches have existing proposals. Say that Feat-1a is good to merge on its own but Feat-1b requires more review/changes. The user can merge Feat-1b but now on github it's marked as closed. If I execute a 'git sync' after Feat-1a is merged the proposal body will have an inconsistent view of the stack of proposals that lead to the complete feature.

In writing this, I realized because we never persist to disk the existing proposal lineage and compute it on the fly, after the merge of Feat-1a, the proposal lineage of feat-1b will no longer be accurate because it will be looking for a proposal targeting main. So now the proposal lineage on the PR body will be


main
    Feat-1b

And so this draft PR is no longer needed because it's not solving the problem I have in mind

yaadata:
I think we can close this due to this proposal not actually solving the problem I am trying to address


Please see my comment here: https://github.com/git-town/git-town/pull/5329#discussion_r2255623898

internal/forge/github/connector.go
kevgo: I also think it's better to now show already shipped branches in the stack visualization. Here is a use case why:

- I work on a team where the tech lead reviews my code
- the tech lead is slow, so I create branch `one` and send a code review
- while branch `one` is under review, I work on branch `two`. To build on top of my changes, I make it a child of branch `one`.
- I'm done with branch `two` and start working on branch `three`. It becomes a child of `two`.
- Finally branch `one` gets approved and shipped.
- I'm done with branch `three` and start working on branch `four`.
- branch `two` gets approved and shipped

This continues forever. After a few weeks there are hundreds of old branches. I don't want to see them in the lineage.

The value of seeing the lineage is mostly to get an overview of the upcoming changes to review.

kevgo:
Feel free to close this PR if you think it's no longer needed.

PR #5384, opened 2025-08-11
Move stack lineage logic to forge package

kevgo:
Currently the logic for calculating the stack lineage is in the configdomain
package. That package contains type definitions for configuration data. A better
package is `forge`, which contains logic around hosting sourcecode on forges.

This PR moves the definitions of data structures around stack lineage into the `forgedomain` package, and logic around this into the `forge` package. Separating these two packages is necessary because of the way Go resolves dependencies between packages.


kevgo:
@Ydot19 this PR moves your code into another package. It doesn't make any modifications to the actual code. Can you please give me the thumbs up that this PR won't interfere with your ongoing work? If it does, please let me know and I let you ship your changes first. I can easily re-create this PR in a few minutes. Thanks!

yaadata:
This shouldn't be a problem for my ongoing work!

Ydot19 (approved):

PR #5425, opened 2025-08-15
enhance: implement undo for proposal body update opcode

yaadata:
## Purpose:

If the proposal body update experiences an issue, enable undoing that update

-------------------------
 - main
   - https://github.com/git-town/git-town/pull/5425 :point_left:

### DIFF
--- a/internal/vm/opcodes/proposal_update_body.go
+++ b/internal/vm/opcodes/proposal_update_body.go
@@ -2,6 +2,7 @@ package opcodes
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/git-town/git-town/v21/internal/forge/forgedomain"
 	"github.com/git-town/git-town/v21/internal/messages"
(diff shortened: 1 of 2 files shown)


internal/vm/opcodes/proposal_update_body.go
```diff
@@ -26,3 +31,16 @@ func (self *ProposalUpdateBody) Run(args shared.RunArgs) error {
 	}
 	return updateProposalBodyFn(self.Proposal.Data, self.UpdatedBody)
 }
+
+func (self *ProposalUpdateBody) ShouldUndoOnError() bool {
+	return true
+}
```
kevgo, line 37: The `ShouldUndoOnError` setting is meant for commands where it makes no sense to ask the user to resolve the problem (like a merge conflict). An example for when it doesn't make sense to ask the user to resolve a merge conflict is the `ship` command. Only fully tested and reviewed changes should get shipped. If a ship fails with a merge conflict, we don't want the user to resolve this conflict on the main branch, because this can break the main branch. And that must never happen. Merge conflicts should always be resolved on feature branches.

For more information, these methods are documented [here](https://github.com/git-town/git-town/blob/main/internal/vm/shared/opcode.go). 

I don't think it makes sense to activate this setting for `ProposalUpdateBody`.

yaadata:
moving this back to draft given your review. I should've dug digger into the undo process

yaadata:
This pull-requests needs to be broken down into smaller pieces as there are a couple of steps needed to make this change work. Closing and will open follow-ups that are more single purpose

PR #5458, opened 2025-08-19
Add --no-auto-resolve flag option

kevgo:


kevgo:
This is now available as part of Git Town [21.5](https://github.com/git-town/git-town/releases/tag/v21.5.0).

PR #5512, opened 2025-09-08
[internal/cmd] enhance(set-parent): update proposal lineage when flag is set

yaadata:
## Summary

Updates proposal lineage automatically when reorganizing branch stacks with `set-parent`.

## Changes

- Added proposal lineage update logic to `set-parent` command when `proposals-show-lineage` is configured to `cli`
- Generates `ProposalUpdateLineage` opcodes for each branch in the stack after parent change

## Testing Plan

- [x] Verified that if a branch is moving to a new stack, all proposals in the new stack are updated.
- [x] Verified that if a branch is in an existing stack and moves to a new stack, all the proposals in the original stack are updated.
- [x] Verified PR descriptions update correctly after parent change
- [x] Confirmed no impact when `proposals-show-lineage` is disabled

-------------------------
 - main
   - https://github.com/git-town/git-town/pull/5512 :point_left:

### DIFF
--- a/internal/cmd/set_parent.go
+++ b/internal/cmd/set_parent.go
@@ -422,6 +422,26 @@ func setParentProgram(newParentOpt Option[gitdomain.LocalBranchName], data setPa
 					Branch: data.initialBranch,
 				},
 			)
+
+			if data.config.NormalConfig.ProposalsShowLineage == forgedomain.ProposalsShowLineageCLI && hasConnector {
+				tree, err := forge.NewProposalStackLineageTree(forge.ProposalStackLineageArgs{
+					Connector:                connector,
+					CurrentBranch:            initialBranchInfo.LocalBranchName(),
+					Lineage:                  data.config.NormalConfig.Lineage,
+					MainAndPerennialBranches: data.config.MainAndPerennials(),
+				})
+				if err != nil {
+					fmt.Printf("failed to update proposal stack lineage: %s\n", err.Error())
+				} else {
+					for branch, proposal := range tree.BranchToProposal {
+						prog.Add(&opcodes.ProposalUpdateLineage{
+							Current:         branch,
+							CurrentProposal: proposal,
+							LineageTree:     MutableSome(tree),
+						})
+					}
+				}
+			}
 		}
 	}
 	return optimizer.Optimize(prog), false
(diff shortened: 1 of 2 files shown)


yaadata:

internal/cmd/set_parent.go
```diff
@@ -424,5 +426,71 @@ func setParentProgram(newParentOpt Option[gitdomain.LocalBranchName], data setPa
 			)
 		}
 	}
+
+	// Update proposal lineage for both cases (removing parent or setting new parent)
+	updateProposalLineage(&prog, newParentOpt, data)
```
kevgo, line 431: Does this always add the opcodes to sync the proposal lineage? I thought it should only sync the lineage when the user has enabled that in the settings?

yaadata: The function exits early if the config is not set.

internal/cmd/set_parent.go
```diff
@@ -424,5 +426,71 @@ func setParentProgram(newParentOpt Option[gitdomain.LocalBranchName], data setPa
...
 	return optimizer.Optimize(prog), false
 }
+
+// updateProposalLineage updates the proposal stack lineage when changing the parent
+func updateProposalLineage(prog *program.Program, newParentOpt Option[gitdomain.LocalBranchName], data setParentData) {
+	connector, hasConnector := data.connector.Get()
+	if data.config.NormalConfig.ProposalsShowLineage != forgedomain.ProposalsShowLineageCLI || !hasConnector {
+		return
+	}
```
kevgo, line 443: This code is checking two independent conditions:
- whether we have a connector
- whether proposals are enabled

Checking them both together makes it too easy to miss one. Let's make this idiot-proof (for people like me) and separate these two checks visually.

It's more lines, but I think it's more clear and maintainable in the long term. What do you think?

```suggestion
	if data.config.NormalConfig.ProposalsShowLineage != forgedomain.ProposalsShowLineageCLI {
		return
	}
	connector, hasConnector := data.connector.Get()
	if !hasConnector {
		return
	}
```

yaadata: Addressed here https://github.com/git-town/git-town/pull/5512/commits/c9bdb75a1e8b3eb2796991b086f3115b874d6a45

PR #5522, opened 2025-09-10
More idiomatic interfaces for connectors

kevgo:
This PR splits the big `forge domain.Connector` interface into many specific interfaces that each describe one capability of a connector. While introducing more complexity, this has several advantages:

1. it gets rid of the ugly solution to return optional higher-level functions
2. it instantiates the third-party library for the forge only if really needed
3. it therefore allows to unit-test more elements of connectors
4. it is more idiomatic Go


internal/forge/proposal_stack_lineage_builder_test.go
```diff
@@ -10,71 +10,28 @@ import (
 	"github.com/git-town/git-town/v21/internal/forge"
 	"github.com/git-town/git-town/v21/internal/forge/forgedomain"
 	"github.com/git-town/git-town/v21/internal/git/gitdomain"
-	"github.com/git-town/git-town/v21/internal/subshell/subshelldomain"
 	. "github.com/git-town/git-town/v21/pkg/prelude"
 )
 
-type mockConnectorProposalStackLineageBuilder struct{}
+type mockProposalFinder struct{}
```
kevgo, line 16: @yaadata I'm updating the forge connector APIs to make them more idiomatic. That led to a few changes and simplifications in your code. No action required on your side, just a heads-up in case you hit merge conflicts.

PR #5587, opened 2025-09-23
fix up/down command doc example graphs

derekspelledcorrectly:
I believe the `up` and `down` command example branch graphs are fundamentally wrong in a very confusing way. They both show the branches being reordered in the stack, as with a `swap`, but that does not seem to match the description of what they do nor what they do in practice.

### DIFF
--- a/website/src/commands/down.md
+++ b/website/src/commands/down.md
@@ -29,9 +29,9 @@ stack:
 ```
 main
  \
-* branch-2
+* branch-1
    \
-    branch-1
+    branch-2
 ```
 
 ## Options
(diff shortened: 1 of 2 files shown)


kevgo:
You're 100% right, that was a copy-and-paste slip. Thanks for fixing it! 🙏

derekspelledcorrectly:
Happy to contribute in a small way. Thanks for the great tool! I discovered it a few days ago and am really enjoying it. It's a huge improvement over raw git for modern workflows, and obviates the need for the 30 complex git aliases I've been carrying around in a .gitconfig file for 15+ years.

kevgo:
That's exactly how Git Town started: as a shared collection of Git aliases and Bash scripts. It's still that, but re-implemented in a proper programming language for more correctness and more robust handling of edge cases.

PR #5513, opened 2025-09-08
[internal/cmd] enhance(prepend): enable updating proposal lineage when flag is set

yaadata:
## Summary

Enhancement: proposal lineage updates in the prepend command when proposals-show-lineage is set to CLI.

## Testing

- [x] Manually verified prepend operation works well with updating the proposal lineage

-------------------------
 - main
   - https://github.com/git-town/git-town/pull/5595
     - https://github.com/git-town/git-town/pull/5513 :point_left:

Stack generated by [Git Town](https://github.com/git-town/git-town)

PR #5595, opened 2025-09-30
[internal/forge] fix: proposal lineage should display properly in the event of a prepend operation

yaadata:
## Summary

Fix proposal stack lineage display when branches without proposals exist in the middle of the stack (occurs during prepend operations).

## Details

Say that you have the current stack

```zsh
$ git-town branch

main
	branch-a
		branch-b
	*		branch-d

```
Each of the non-perennial branches have proposals but when you run (for example)
               
```zsh
$ git-town prepend branch-c
$ git-town branch

main
	branch-a
		branch-b
	*		branch-c
				branch-d

```

Because branch-c is a new branch, it will not have a proposal but will be a target of proposal where the base=branch-d and the target=branch-c

Even though there is branch with a missing proposal in the middle of the stack, we should try to show all the available proposals leading up to the perennial branch.

### Before this change

Taking the example above, because all the original feature branches have proposals and you are on the first branch, this is the stack lineage that would be displayed after a prepend operation.

```zsh
main
	- proposal_a :point_left:  
		- proposal_b
```

### After this change

After the prepend operations, all the available proposals will be shown

```zsh
main
	- proposal_a :point_left:  
		- proposal_b
		      - proposal_d 
```

-------------------------
 - main
   - https://github.com/git-town/git-town/pull/5595 :point_left:
     - https://github.com/git-town/git-town/pull/5513


yaadata:

kevgo:
Nice improvement!

PR #5597, opened 2025-09-30
fix go.sum hash

Mause:
it's not clear why or how this entry is out-of-date, but we've had to add this as a patch upstream to be able to build it 

```
verifying github.com/charmbracelet/x/ansi@v0.9.2: checksum mismatch
        downloaded: h1:92AGsQmNTRMzuzHEYfCdjQeUzTrgE1vfO5/7fEVoXdY=
        go.sum:     h1:VxqhWbZIXFl3/ufdpzcwlaPZRPoaOj8r+KtyeRG3C0g=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```


kevgo:
Looks like others encounter this problem as well: https://github.com/charmbracelet/gum/issues/921. I was concerned about this being a supply-chain attack, but given the widespread exposure, and the many eyeballs on this, it seems unlikely this is the case here. 

Thanks for fixing this! :pray:

PR #5599, opened 2025-10-02
[internal/cmd] enhance(ship): update proposal stack lineage on ship command

yaadata:
## Testing

- [x] Validate that shipping from a child branch to a non-perennial branch appropriately updates the stack lineage
- [x] Validate that shipping a feature branch into a perennial branch appropriately updates the stack lineage

### Note:

There is a bug I found in the existing code where if you have the following stack

```
$ git-town branch
main
   * branch_a
         branch_c
```

And you ship `branch_a` into main (a perennial branch), the stack lineage becomes

```main
$ git-town branch
main
branch_c
```

and branch_c becomes a branch without a parent. What should happen is that `main` becomes the new parent of `branch_c`. Because this bug is with existing code, it will be handled in a separate pull-request (Fixed #5600 )

-------------------------
 - main
   - https://github.com/git-town/git-town/pull/5600
     - https://github.com/git-town/git-town/pull/5599 :point_left:

### DIFF
--- a/internal/cmd/ship/cmd.go
+++ b/internal/cmd/ship/cmd.go
@@ -12,6 +12,7 @@ import (
 	"github.com/git-town/git-town/v22/internal/config/cliconfig"
 	"github.com/git-town/git-town/v22/internal/config/configdomain"
 	"github.com/git-town/git-town/v22/internal/execute"
+	"github.com/git-town/git-town/v22/internal/forge"
 	"github.com/git-town/git-town/v22/internal/forge/forgedomain"
 	"github.com/git-town/git-town/v22/internal/git/gitdomain"
 	"github.com/git-town/git-town/v22/internal/gohacks"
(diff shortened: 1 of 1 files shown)


yaadata:
@kevgo this PR is ready for a review. Please also see the bug I found with the existing logic for ship. Details in the description (fixed in #5600)

internal/cmd/ship/cmd.go
```diff
@@ -169,6 +170,35 @@ Start:
...
+		if connector, hasConnector := sharedData.connector.Get(); hasConnector {
+			if proposalFinder, hasProposalFinder := connector.(forgedomain.ProposalFinder); hasProposalFinder {
+				tree, err := forge.NewProposalStackLineageTree(forge.ProposalStackLineageArgs{
+					Connector:                proposalFinder,
+					CurrentBranch:            sharedData.initialBranch,
+					Lineage:                  sharedData.config.NormalConfig.Lineage,
+					MainAndPerennialBranches: sharedData.config.MainAndPerennials(),
+				})
+				if err != nil {
+					fmt.Printf("failed to update proposal stack lineage: %s\n", err.Error())
```
kevgo, line 183: This error message seems wrong. forge.NewProposalStackLineageTree doesn't update the proposals.

kevgo (approved):
You could also extract this code into a method called `updateProposalLineage`. This would keep the `executeShip` function high-level.

internal/cmd/ship/cmd.go
yaadata: My train of thought was that it gives context to what the overall operation that failed PLUS wraps with the more specific error message. I do this in more than one place, so if we disagree with this error message I need to update the error messages in the other APIs that I have updated. I can do this in a separate pull-request to keep all these messages consistent

yaadata:
> You could also extract this code into a method called updateProposalLineage. This would keep the executeShip function high-level.

Let me extract this into a separate pull-request because `ship` is not the only command that has this duplicative logic

PR #5606, opened 2025-10-04
[internal/cmd] enhance(detach): update proposal targets on detach

yaadata:
### Background

On detach, update the proposal targets of the child branches of the detached
branch as well as the detached branch


internal/cmd/detach.go
```diff
@@ -161,22 +161,23 @@ Start:
...
-	initialBranch       gitdomain.LocalBranchName
-	inputs              dialogcomponents.Inputs
-	nonExistingBranches gitdomain.LocalBranchNames // branches that are listed in the lineage information, but don't exist in the repo, neither locally nor remotely
-	parentBranch        gitdomain.LocalBranchName
-	previousBranch      Option[gitdomain.LocalBranchName]
-	stashSize           gitdomain.StashSize
+	branchInfosLastRun     Option[gitdomain.BranchInfos]
+	branchToDetachInfo     gitdomain.BranchInfo
+	branchToDetachName     gitdomain.LocalBranchName
+	branchToDetachProposal Option[forgedomain.Proposal]
```
yaadata, line 170: New

yaadata:
@kevgo could I get a review of this pull-request. You already approved the sister pull-request https://github.com/git-town/git-town/pull/5607 related to proposal lineage. This pull-request updates the proposal targets of the children of detached branches

internal/cmd/detach.go
```diff
@@ -294,16 +295,26 @@ func determineDetachData(repo execute.OpenRepoResult) (data detachData, flow con
 	if branchHasMergeCommits {
 		return data, configdomain.ProgramFlowExit, fmt.Errorf(messages.BranchContainsMergeCommits, branchNameToDetach)
 	}
+
+	connectorProposalFinder := None[forgedomain.ProposalFinder]()
+	branchToDetachPropsal := None[forgedomain.Proposal]()
+	if connector, hasConnector := connector.Get(); hasConnector {
+		if proposalFinder, canFindProposals := connector.(forgedomain.ProposalFinder); canFindProposals {
+			connectorProposalFinder = Some(proposalFinder)
+			branchToDetachPropsal, err = proposalFinder.FindProposal(branchNameToDetach, parentBranch)
```
kevgo:
```suggestion
			branchToDetachProposal, err = proposalFinder.FindProposal(branchNameToDetach, parentBranch)
```

internal/cmd/detach.go
```diff
@@ -395,15 +407,52 @@ func detachProgram(repo execute.OpenRepoResult, data detachData, finalMessages s
 				Parent: data.config.ValidatedConfigData.MainBranch,
 			},
 		)
-		for _, child := range data.config.NormalConfig.Lineage.Children(data.branchToDetachName, data.config.NormalConfig.Order) {
+	}
+	if proposal, hasProposal := data.branchToDetachProposal.Get(); hasProposal {
+		prog.Value.Add(
+			&opcodes.ProposalUpdateTarget{
+				NewBranch: data.parentBranch,
+				OldBranch: data.branchToDetachName,
```
kevgo: This code changes the target branch of the proposal for the branch to detach.

Shouldn't it have these values? 

```suggestion
				NewBranch: data.config.ValidatedConfigData.MainBranch,
				OldBranch: data.parentBranch,
```

internal/cmd/detach.go
yaadata: https://github.com/git-town/git-town/pull/5606/commits/a61a2714722c644ae07e23a119e5d53fc743211d

internal/cmd/detach.go
yaadata: This is was a gap when I rebased the branch, I must have updated incorrectly https://github.com/git-town/git-town/pull/5606/commits/90a42753bbc063a612e72ae2cfad78a42b7c331d

I will do a round of manually testing before merging

yaadata: confirmed this works as expected again

internal/cmd/detach.go
```diff
@@ -294,16 +295,26 @@ func determineDetachData(repo execute.OpenRepoResult) (data detachData, flow con
...
-		if connector, hasConnector := connector.Get(); hasConnector {
-			if proposalFinder, canFindProposals := connector.(forgedomain.ProposalFinder); canFindProposals {
-				proposal, err = proposalFinder.FindProposal(childBranch, initialBranch)
-				if err != nil {
-					return data, configdomain.ProgramFlowExit, err
-				}
+		if proposalFinder, hasProposalFinder := connectorProposalFinder.Get(); hasProposalFinder {
+			proposal, err = proposalFinder.FindProposal(childBranch, initialBranch)
+			if err != nil {
+				return data, configdomain.ProgramFlowExit, err
```
kevgo: Now we have two separate blocks that use the `proposalFinder`. Both check whether we have a `proposalFinder`. Can you combine these two sections into one? This would allow us to get rid of the `connectorProposalFinder` variable.

1. Declare all the variables that need to be populated by the connector: `branchToDetachProposal`, `children`
2. Unpack the `proposalFinder`
3. If `canFindProposals`, then load the values of `branchToDetachProposal` and then run the loop that loads the values for `children`

internal/cmd/detach.go
```diff
@@ -395,15 +407,52 @@ func detachProgram(repo execute.OpenRepoResult, data detachData, finalMessages s
...
+					NewBranch: data.parentBranch,
+					OldBranch: data.branchToDetachName,
+					Proposal:  proposal,
 				},
 			)
 		}
 	}
+	if data.config.NormalConfig.ProposalsShowLineage == forgedomain.ProposalsShowLineageCLI {
+		_ = sync.AddStackLineageUpdateOpcodes(
+			sync.AddStackLineageUpdateOpcodesArgs{
```
kevgo: Nitpick: The rest of the codebase has the data type for the named arguments on the previous line. This makes it look like proper named arguments. Would that work for you here?

```suggestion
		_ = sync.AddStackLineageUpdateOpcodes(sync.AddStackLineageUpdateOpcodesArgs{
```

kevgo (approved):
Thanks for fixing this! It can be merged after the final commits are addressed.

internal/cmd/detach.go
yaadata: @kevgo this suggestion led to endtoend test errors because this suggestion

> If canFindProposals, then load the values of branchToDetachProposal and then run the loop that loads the values for children

assumes that the detachChild struct only is instantiated when a proposalFinder is defined. But detachChild struct can exist even if the forge does not have ProposalFinder defined. See: https://github.com/git-town/git-town/actions/runs/19173386524/job/54811664946?pr=5606

yaadata: https://github.com/git-town/git-town/pull/5606/commits/904d1e7773080e3830a1f643c3c0cf6948db73a8 related commit where I tried your suggestion

internal/cmd/detach.go
yaadata: https://github.com/git-town/git-town/pull/5606/commits/f7dbc98f64d2c2bd2c1c9495337bbe7c01a2c6a4

internal/cmd/detach.go
kevgo: Ah, right! Thanks for giving it a try! :slightly_smiling_face:

PR #5610, opened 2025-10-06
[internal/setup] feat: add proposal show lineage to config init

yaadata:
Add proposal show lineage configuration to setup

Adds support for configuring the proposals.showLineage setting during setup initialization. Users can now choose how Git Town shares stack lineage in proposals (none, cli, or ci) directly in the setup assistant, rather than requiring manual Git config changes.

**Changes:**
- Dialog component for proposals show lineage option selection
- Setup workflow now prompts for proposals show lineage setting when entering all configs
- Proper save and remove functions for persisting configuration to Git local scope
- Feature tests updated to reflect new config option

-------------------------
 - main
   - https://github.com/git-town/git-town/pull/5610 :point_left:

<sup>[Stack](https://www.git-town.com/how-to/github-actions-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>

### DIFF
--- a/internal/cli/dialog/proposals_show_lineage.go
+++ b/internal/cli/dialog/proposals_show_lineage.go
@@ -15,6 +15,8 @@ const (
 	proposalsShowLineageTitle = `Proposals Show Lineage`
 	proposalsShowLineageHelp  = `
 	How should Git Town share stack lineage in proposals?
+
+	See https://www.git-town.com/how-to/github-actions-breadcrumb.html for details.
 `
 )
 
@@ -30,15 +32,15 @@ func ProposalsShowLineage(args Args[forgedomain.ProposalsShowLineage]) (Option[f
 	entries = append(entries, list.Entries[Option[forgedomain.ProposalsShowLineage]]{
 		{
 			Data: Some(forgedomain.ProposalsShowLineageNone),
-			Text: "stack lineage is not displayed in proposals",
+			Text: "no stack lineage in proposals",
 		},
 		{
 			Data: Some(forgedomain.ProposalsShowLineageCLI),
-			Text: "the Git Town CLI updates the stack lineage in proposals",
+			Text: "Git Town CLI embeds the stack lineage into proposals",
 		},
 		{
 			Data: Some(forgedomain.ProposalsShowLineageCI),
-			Text: "stack lineage is displayed and updated through ci integrations, with tools such as https://github.com/git-town/action",
+			Text: "CI server embeds the stack lineage into proposals",
 		},
 	}...)
 	defaultPos := entries.IndexOf(args.Local)


internal/config/configfile/data.go
```diff
@@ -6,7 +6,7 @@ type Data struct {
 	Create                   *Create       `toml:"create"`
 	CreatePrototypeBranches  *bool         `toml:"create-prototype-branches"`
 	Hosting                  *Hosting      `toml:"hosting"`
-	Propose                  *Propose      `toml:"propose"`
+	Proposals                *Proposals    `toml:"proposals"`
```
yaadata: I updated this line because proposals do not show lineage only on a propose command but anytime a lineage changes or a sync command is executed

internal/config/configfile/data.go
```diff
@@ -55,8 +55,8 @@ func (self Hosting) IsEmpty() bool {
 	return self.ForgeType == nil && self.OriginHostname == nil && self.Platform == nil
 }
 
-type Propose struct {
-	Lineage *string `toml:"lineage"`
+type Proposals struct {
+	Lineage *string `toml:"show-lineage"`
```
yaadata: If we want to bikeshed the naming used in the config, this is the PR to do it before we make this public

kevgo:
Thanks for building this, it's an important PR! Since this change makes the stack lineage feature public to everybody, I'm going to wait with merging it until the stack lineage goes live officially. Which we are getting really close to! Does this work for you?

features/init/existing_configuration/git_global.feature
```diff
@@ -1,6 +1,5 @@
 @messyoutput
 Feature: setup a new repo when I have configured some things in global Git metadata
-
```
kevgo, line 3: Please add this empty line back :slightly_smiling_face:

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,56 @@
...
+	proposalsShowLineageHelp  = `
+How should Git Town share stack lineage in proposals?
+
+Possible Options:
+	- none: (default) Git Town should not share stack 
+					lineage in proposals
+	- cli: Git Town shares or updates stack lineage in proposals
+				anytime your stack lineage changes via the cli
+	- ci: Git Town manages and shows proposal lineage through ci 
+	      integrations, with tools such as https://github.com/git-town/action
```
kevgo: Can you move these explanations into the radio list through which the user chooses the option?

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,56 @@
...
+		entries = append(entries, list.Entry[Option[forgedomain.ProposalsShowLineage]]{
+			Data: None[forgedomain.ProposalsShowLineage](),
+			Text: fmt.Sprintf(messages.DialogUseGlobalValue, global),
+		})
+	}
+
+	entries = append(entries, list.Entries[Option[forgedomain.ProposalsShowLineage]]{
+		{
+			Data: Some(forgedomain.ProposalsShowLineageNone),
+			Text: "do not show stack lineage in proposals",
```
kevgo:
```suggestion
			Text: "do not display stack lineage in proposals",
```

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,56 @@
...
+		},
+		{
+			Data: Some(forgedomain.ProposalsShowLineageCLI),
+			Text: "cli command execution show the latest stack lineage in a proposal(s)",
```
kevgo:
```suggestion
			Text: "the Git Town CLI updates the stack lineage in proposals",
```

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,56 @@
...
+		},
+		{
+			Data: Some(forgedomain.ProposalsShowLineageCI),
+			Text: "ci process shows the latest  stack lineage in a proposal(s)",
```
kevgo:
```suggestion
			Text: "github.com/git-town/action updates the stack lineage in proposals",
```

internal/cmd/config/root.go
```diff
@@ -110,6 +110,9 @@ func printConfig(config config.UnvalidatedConfig) {
 	print.Entry("GitLab connector type", format.OptionalStringerSetting(config.NormalConfig.GitLabConnectorType))
 	print.Entry("GitLab token", format.OptionalStringerSetting(config.NormalConfig.GitLabToken))
 	fmt.Println()
+	print.Header("Proposals")
```
kevgo: Let's stick with the verb-based heading titles where possible.

```suggestion
	print.Header("Propose")
```

internal/config/configfile/data.go
kevgo: One thing to consider when we change the config file format: This would force all Git Town users (more than tens of thousands of people) to update the `git-branches.toml` files in their repos. Ideally we put our users through this only if really necessary.

I think the current verb form works fine for headings. The connection to proposals should be clear to users.

internal/config/configfile/data.go
kevgo: Unfortunately, the config file format is now written in stone because so many people use it now :slightly_smiling_face: All we can do is *add* new settings to the config file (including additional names for existing options), but we can never *change* existing options. That's a very good problem to have!

yaadata:
@kevgo agreed, I just wanted to put this on your radar

internal/config/configfile/load_test.go
```diff
@@ -44,8 +44,8 @@ share-new-branches = "push"
 forge-type = "github"
 origin-hostname = "github.com"
 
-[propose]
-lineage = "ci"
+[proposals]
+show-lineage = "cli"
```
kevgo:
```suggestion
lineage = "cli"
```

I like the old way better.

1. It's shorter.
2. If we say something specific like `show-lineage`, then there should also be another option with lineage, like `hide-lineage`, `change-lineage`, etc. We don't have that.

internal/messages/en.go
```diff
@@ -235,6 +235,7 @@ Please upgrade to the new format: create.new-branch-type = "prototype"`
 	ProposalNoNumberGiven                   = "no proposal number given"
 	ProposalNoParent                        = "branch %q has no parent and can therefore not be proposed"
 	ProposalSourceCannotUpdate              = "cannot update the proposal source branch on your forge"
+	ProposalsShowLineage                    = "Proposals Show Lineage: %s\n"
```
kevgo:
```suggestion
	ProposalsLineage                    = "Proposals display the stack lineage: %s\n"
```

internal/config/configfile/data.go
yaadata: I thought about this. We currently don't write this header in any toml files today. This is what made me comfortable making this change.

We can keep it but I think it will be confusing because users will associate this section with the `propose` command when its not the only command making this change. It was the confusing part for me when enabling this section to be written to our toml files.

You have final say but this was my rationale

kevgo: Thanks for the extra context! I had forgotten this isn't public yet. And good call on raising the naming question. This is one of those decisions we’ll be stuck with indefinitely, so it’s worth getting right.

### Background on the naming pattern

The config file uses section names that are mostly _verbs_:

- `create`
- `ship`
- `sync`

These don’t map 1:1 to Git Town commands (there is no `create` command), but they represent **what Git Town is doing** when those settings apply. Multiple commands trigger "create", "sync", or "ship" behavior, and all of them rely on the options in those sections.

Examples:

- `hack`, `append`, and `prepend` all _create_ branches
- `hack`, `append`, `prepend`, `propose`, and `sync` all sync branches

We also have two _noun_ sections where no good verb exists:

- `branches`: Git Town manages branches, it doesn't "branch" itself
- `hosting`: Git Town integrates with hosting providers, it doesn’t "host" itself

### Applying this to `propose` vs `proposals`

So for the new section, the question becomes:

> Are we describing how Git Town _proposes branches_, or how Git Town _manages proposals_?

Both interpretations are reasonable. Here's the scorecard so far:

Argument | Favors
-- | --
`propose` could be confused with the `git town propose` command | `proposals`
But we already accept similar ambiguity with `sync` | `propose`
Git Town proposes branches as part of the create-sync-propose-ship workflow | `propose`
Other commands don't "propose" but might update proposals | `proposals`
One could argue that keeping proposals up to date is part of proposing | `propose`
Git Town will update proposals that were created externally | `proposals`

So we're at a 3–3 tie (depending on how one frames it).

### Considering upcoming changes

#5695 plans a few additions. Here's what the section will likely contain soon:


```toml
[propose]
show-lineage = true
title = "first-commit-title"
body-template = "This proposal is for {{branch}}"
body-template-file = "pull-request.liquid"
```

vs:

```toml
[proposals]
show-lineage = true
title = "first-commit-title"
body-template = "This proposal is for {{branch}}"
body-template-file = "pull-request.liquid"
```

The `title`, `body-template`, and `body-template-file` settings only apply for creating proposals, they don't apply when updating proposals. This is an argument for naming the section `propose`.

yaadata: @kevgo should `proposals-show-lineage` be moved to the sync section? Showing the state of proposal in stack is really a sync operation whenever the lineage changes. Maybe with a rename

```toml
[sync]
proposal-lineage = "cli"
```

I think keeping it called `propose` makes sense for `title`, `body-template`, and `body-template-file`

kevgo: The name `sync.proposal-lineage` feels misleading to me: it implies that `sync` will synchronize lineage data across machines or with a remote service. I'd rather avoid that association, and also keep the `sync.*lineage*` namespace free in case we ever implement cross-machine sync in the future. 🙂

Another consideration: #5678 changes `git town propose` so it embeds the branch lineage when creating proposals. This effectively changes the entire feature: lineage is now written by `propose`, not only maintained by other Git Town commands afterward. You were right that the `propose` section originally felt off, but only because the feature hadn't fully landed yet.

Once #5678 is merged, the feature becomes: `git town propose` embeds lineage into the proposal, and the other Git Town commands just keep that embedded lineage current. So the core behavior is owned by `propose`.

Given that, I think the original config format is the most accurate:

```toml
[propose]
lineage = "cli"
```

I'd vote we stick with this. How does that sound?

yaadata: Sounds good

internal/messages/en.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/71b82014378a76f4e89867a35fe1a1f8e38e7ee6

internal/config/configfile/load_test.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/79423b27f74651b1dec7468ca59113d5ea95ffd0

internal/cmd/config/root.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/79423b27f74651b1dec7468ca59113d5ea95ffd0

features/init/existing_configuration/git_global.feature
yaadata: https://github.com/git-town/git-town/pull/5610/commits/42fb6c735bd9a001899c33a182cdd634306acaa1

internal/cli/dialog/proposals_show_lineage.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/d44e0d78caf0d32a7ee6667ecd82dacbc5e1fa12

internal/cli/dialog/proposals_show_lineage.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/048a7241e743f86df018f048db8b8e2e193af83b

internal/cli/dialog/proposals_show_lineage.go
yaadata: I have updated this suggestion https://github.com/git-town/git-town/pull/5610/commits/953dbc8bd7d07d57b78fb864a78ee28869e13bef but I don't agree with it. This assumes that this option is exclusive to the github forge because these actions are exclusive to github

internal/cli/dialog/proposals_show_lineage.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/f2233f78b17c38df00172096607bce73d5a77dc1

internal/cli/dialog/proposals_show_lineage.go
yaadata: I actually updated per your prior suggestion here https://github.com/git-town/git-town/pull/5610#discussion_r2474463875

yaadata:
@kevgo this is ready for another look

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,48 @@
...
+		entries = append(entries, list.Entry[Option[forgedomain.ProposalsShowLineage]]{
+			Data: None[forgedomain.ProposalsShowLineage](),
+			Text: fmt.Sprintf(messages.DialogUseGlobalValue, global),
+		})
+	}
+
+	entries = append(entries, list.Entries[Option[forgedomain.ProposalsShowLineage]]{
+		{
+			Data: Some(forgedomain.ProposalsShowLineageNone),
+			Text: "stack lineage is not displayed in proposals",
```
kevgo: Let's use active voice here.



```suggestion
			Text: "no stack lineage in proposals",
```

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		{
+			Data: Some(forgedomain.ProposalsShowLineageCLI),
+			Text: "the Git Town CLI updates the stack lineage in proposals",
```
kevgo:
```suggestion
			Text: "Git Town CLI embeds the stack lineage into proposals",
```

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		{
+			Data: Some(forgedomain.ProposalsShowLineageCI),
+			Text: "stack lineage is displayed and updated through ci integrations, with tools such as https://github.com/git-town/action",
```
kevgo:
```suggestion
			Text: "CI server embeds the stack lineage into proposals",
```

internal/cli/dialog/proposals_show_lineage.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/git-town/git-town/v22/internal/cli/dialog/dialogdomain"
+	"github.com/git-town/git-town/v22/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v22/internal/messages"
+	. "github.com/git-town/git-town/v22/pkg/prelude"
+)
+
+const (
+	proposalsShowLineageTitle = `Proposals Show Lineage`
+	proposalsShowLineageHelp  = `
+	How should Git Town share stack lineage in proposals?
```
kevgo, line 17: Since there are so many options here, what do you think about linking to the web page for this (assuming we add more details to that page)?

```suggestion
	How should Git Town share stack lineage in proposals?
	
	See https://www.git-town.com/how-to/github-actions-breadcrumb.html for details.
```

yaadata: Can I do this in a follow-up PR. I want to update the page name to be `https://www.git-town.com/how-to/display-proposal-stack-lineage.html` with the page contents updated to include both both the cli and github action. Once that page is updated I will loop back and update this here.

yaadata: actually nevermind, I applied this suggestion and I will follow through on the improvements in a separate PR https://github.com/git-town/git-town/pull/5610/commits/defa031457336146902772c2e0439cd21d2a33a9

internal/cli/dialog/proposals_show_lineage.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/defa031457336146902772c2e0439cd21d2a33a9

internal/cli/dialog/proposals_show_lineage.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/defa031457336146902772c2e0439cd21d2a33a9

internal/cli/dialog/proposals_show_lineage.go
yaadata: https://github.com/git-town/git-town/pull/5610/commits/defa031457336146902772c2e0439cd21d2a33a9

yaadata:
@kevgo 

There are no blockers I am aware of. 2 Follow-ups on lineage though

- I would like to see https://github.com/git-town/git-town/pull/5678 land even though its awaiting some suggested improvements. We can create a github issue for your comment here https://github.com/git-town/git-town/pull/5678#discussion_r2464157039 so we don't lose track of suggested improvements. 

- I think we need to improve our caching story. Sometimes we pull the same proposal details more than once in a single git-town command. I don't know if you have something in mind but would like to create an discussion topic on this so we have an idea of what we want so we can get this landed. I think it would delight users to see these commands get snappier.

kevgo:
@yaadata I'm with you on this. I'll look into landing #5678. I also agree that we should improve how we cache knowledge about proposals. I have a few ideas in mind for that. I've been holding off on changes here until you wrap up your work on it, since it's going to touch quite a bit of the codebase.

yaadata:
@kevgo Sounds good. I will leave #5678 in your court and focus on the documentation page for lineage.

This wraps up my work on this. If you need support or issues that come up related to my changes I will be around. I will be on the lookout for`help wanted` tags  if you have other features/capabilities you want tackled but don't have the bandwidth.

kevgo:
@yaadata if you have availability, and since you're super familiar with this part of the codebase (and built the entire feature) feel free to take a stab at #5678 if you're interested! I'm planning to wrap up the things I currently have in flight over the next couple of days.

If you'd rather pick up something else, feel free to work on any of the unassigned issue. They all need help :slightly_smiling_face: I'll start assigning issues to myself when I pick them up.

yaadata:
Yep I can shepherd that pull-request

kevgo:
Thanks!

PR #5678, opened 2025-10-25
ProposalCreate: inject the branch lineage when creating a proposal

j2fw:
I quite like the lineage string in the PR description, as it adds useful context for reviewers! However, if I use `git town propose`, it doesn't add the lineage. This change prepends the `ProposalUpdateLineage` opcode after the proposal is created, after which point we have the full lineage including the current proposal.

Since this change puts it into the `ProposalCreate` opcode, I believe it should take effect in these commands:

- append
- prepend
- propose

Perhaps ironically, this PR doesn't have the lineage. As per [the docs](https://www.git-town.com/how-to/forked-repo.html), I have an `upstream` remote set up, but when I did `git town propose`, it created the PR within my fork. Thus, I've made this proposal manually.

### Tests

These tests passed:

- `make unit`
- `make unit-all`
- `make unit-race`

I also tried to run `make cuke`. It failed, but it also seems to fail on `main` for me.

This is my first contribution, so I'm very open to notes and corrections.

### DIFF
--- a/internal/vm/opcodes/proposal_create.go
+++ b/internal/vm/opcodes/proposal_create.go
@@ -51,6 +51,7 @@ func (self *ProposalCreate) Run(args shared.RunArgs) error {
 				Order:                    args.Config.Value.NormalConfig.Order,
 			})
 			if err != nil {
+				// TODO: make sure error message return from failing to construct lineage is consistent across all invocations
 				fmt.Printf("failed to construct proposal stack lineage: %s\n", err.Error())
 			}
 			proposalOpt, err := proposalFinder.FindProposal(self.Branch, parentBranch)


j2fw:
cc @yaadata: you seem to have done a lot with the proposal lineage feature

yaadata:
@j2fw I thought about adding this to ProposalCreate but I hesitated due to the fact that from some forge types we allow you to complete the proposal creation within the browser, an example of this is with the github cli. This is now an async process and proposal number (pr url) is generated after the cli command has completed.

So my question is: does this actually work the way that you expect with manual testing of your branch? When I tried it did not that is why I left this out

yaadata:
This limitations is why I was thinking about adding docs to recommend `git-town sync` after a proposal has been created, because the sync command will update an existing proposal with the stack lineage

j2fw:
> So my question is: does this actually work the way that you expect with manual testing of your branch? When I tried it did not that is why I left this out

Right, that crossed my mind too. In the proposed code, we check `proposalOpt.IsSome()`. This should be None in the case when we shuttle off to the browser, and I expect it'll complete without having added the lineage. The user can then `sync` manually if they want.

I suppose it'd be possible to sit around and poll for the PR, but I thought that could be confusing for the user.

My workflow these days (I've only used the github forge so far) is to make the proposal as draft, which I expect should work with this.

> This limitations is why I was thinking about adding docs to recommend `git-town sync` after a proposal has been created, because the sync command will update an existing proposal with the stack lineage

Right, this has been my workflow so far! However I feel it's a little nicer to have it bundled up here. This is just IMHO though.

yaadata:
Oh I see,

From my standpoint I quite like this. We make sure that the cli does as much as it can.

internal/vm/opcodes/proposal_create.go
```diff
@@ -27,12 +28,43 @@ func (self *ProposalCreate) Run(args shared.RunArgs) error {
...
+		if proposalFinder, canFindProposals := connector.(forgedomain.ProposalFinder); canFindProposals {
+			lineageTree, err := forge.NewProposalStackLineageTree(forge.ProposalStackLineageArgs{
+				Connector:                proposalFinder,
+				CurrentBranch:            self.Branch,
+				Lineage:                  args.Config.Value.NormalConfig.Lineage,
+				MainAndPerennialBranches: args.Config.Value.MainAndPerennials(),
+				Order:                    args.Config.Value.NormalConfig.Order,
+			})
+			if err != nil {
+				fmt.Printf("failed to construct proposal stack lineage: %s\n", err.Error())
```
yaadata, line 55: nit: 

For myself, could you add a todo to `TODO: make sure error message return from failing to construct lineage is consistent across all invocations`

yaadata (approved):
LGTM, @kevgo for any additional thoughts

internal/vm/opcodes/proposal_create.go
j2fw: > nit:
> 
> For myself, could you add a todo to `TODO: make sure error message return from failing to construct lineage is consistent across all invocations`

done!

j2fw:
Updated with the TODO comment.

I also noticed a bug when testing more where the `gh` connector's CreateProposal tries to open the PR before it exists when using the "Create in browser" flow of the `gh` cli. I created #5681 about it.

I think it's a bit beyond the scope of this change and can be addressed in another proposal.

internal/vm/opcodes/proposal_create.go
```diff
@@ -27,12 +28,44 @@ func (self *ProposalCreate) Run(args shared.RunArgs) error {
...
 		return forgedomain.UnsupportedServiceError()
 	}
-	return connector.CreateProposal(forgedomain.CreateProposalArgs{
+
+	err := connector.CreateProposal(forgedomain.CreateProposalArgs{
 		Branch:         self.Branch,
 		FrontendRunner: args.Frontend,
 		MainBranch:     self.MainBranch,
 		ParentBranch:   parentBranch,
 		ProposalBody:   self.ProposalBody,
```
kevgo, line 37: Here we create the proposal at the forge using the provided body. At this point, we already have all info needed to render the branch lineage. Could we just include the fully rendered branch lineage in the proposal body here? That way, we wouldn’t need to add it in a separate step later.

Benefits:

* The proposal includes the correct branch lineage from the start
* Works consistently whether the proposal is created via `gh` or manually in the browser

j2fw: Ah, one quick clarification: we're operating on the *proposal* lineage, rather than the branch lineage! This provides clickable links to other proposals for reviewers into the proposal body.

Thus, if we want to include the proposal we're creating in this opcode, AFAICT we have to wait until it exists before inserting. I couldn't find a way to get a github pull request link by branch rather than by number unfortunately.

kevgo: Good point. Could we format the proposal lineage so that the proposal for the current branch isn't rendered as a link? That link would just point to the same page the user is already on, so it doesn’t really add value anyways.

kevgo: > I couldn't find a way to get a github pull request link by branch rather than by number unfortunately.

The [GitHub API](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests) provides a way to list pull requests and filter by base branch name. That should let us retrieve all pull requests associated with a given branch. Is this what you were looking for?

j2fw: > Good point. Could we format the proposal lineage so that the proposal for the current branch isn't rendered as a link? That link would just point to the same page the user is already on, so it doesn’t really add value anyways.

Wow true, this could work and prevent a round-trip of churn because we know the branch already!

I'll take a shot at this when I have a sec

yaadata:
@j2fw would you like assistance on the recommended changes suggested by Kevin

yaadata:
For transparency, to move this PR further along I'll attempt the suggested improvements in a separate PR. All credit to @j2fw for this PR, this change will be properly attributed in my upcoming pr cc: @kevgo

kevgo:
@yaadata good choice! I really like that you give @j2fw the credit here. When you are ready, go ahead and merge this PR, and feel free to tackle the reviewer feedback and missing features in a follow-up. Thanks for helping move this over the finish line! :raised_hands:

j2fw:
Thanks for taking over the PR! Life got in the way of following up here.

Issue #5677, opened 2025-10-24
Option to prepend username to branch


james-harlyy:
In our company, we have branch naming conventions where we prepend the branch name with our username. Such as `james-harlyy/feature-a`. It would be nice to have an option in the Git Town Config to automatically do this when creating branches via `hack`, `append`, `prepend`, `rename` etc.

I'm opening a conversation to start this, I can potentially contribute this myself.

kevgo:
Thanks for reaching out about this, I like the idea! Can you please share some thoughts how this should work in your opinion? Here are some ideas:

- set a (hypothetical) Git Town setting `branch-prefix` to `james-harlyy/`
- `git town hack feature-a` ==> Git Town creates branch `james-harlyy/feature-a`
- `git town hack james-harlyy/feature-a` ==> Git Town recognizes that the prefix already exists and does not prepend another one
- `git town rename feature-b` ==> Git Town renames the current branch to `james-harlyy/feature-b`

If I want to create a branch without the configured prefix, I would override the configured prefix using a CLI flag:

```sh
git town hack --branch-prefix="" release/2025_10
```

Or set a different prefix for that call:

```
git town hack --branch-prefix="release/" 2025_10
```

Or do the same via an environment variable:

```
GIT_TOWN_BRANCH_PREFIX=release/ git town hack 2025_10
```

kevgo:
If you’d like to take a stab at implementing this, that would be awesome! Feel free to take on as much or as little of it as you like.

Here's a minimal starting point: loading the setting from Git metadata and using it in the `append` command.

#### Set up the configuration setting:

* Create a new data type for this setting, similar to [GitHubToken](https://github.com/git-town/git-town/issues/internal/forge/forgedomain/github_token.go).
* Add a field for it to [PartialConfig](https://github.com/git-town/git-town/blob/main/internal/config/configdomain/partial_config.go) and [NormalConfig](https://github.com/git-town/git-town/blob/main/internal/config/normal_config.go).
* Extract its value from the bulk-loaded Git metadata in [NewPartialConfigFromSnapshot](https://github.com/git-town/git-town/blob/main/internal/config/parse_snapshot.go).

#### Use the configuration setting:

* Apply the configured prefix to `targetBranch` in [append](https://github.com/git-town/git-town/blob/main/internal/cmd/append.go).

#### Development tips:

* `make install` runs the type checker and installs the binary into `~/go/bin` for local testing.
* `make lint` runs the linters.
[...]

PR #5682, opened 2025-10-27
Added the "GitHub" username configuration

james-harlyy:
- This PR adds the "GithubUsername" configuration option, although it currently does nothing.
- The child PR uses this option in the `append`, `prepend` and `hack` commands. https://github.com/james-harlyy/git-town/pull/2

### DIFF
--- a/internal/cmd/config/root.go
+++ b/internal/cmd/config/root.go
@@ -107,6 +107,7 @@ func printConfig(config config.UnvalidatedConfig) {
 	print.Entry("Gitea token", format.OptionalStringerSetting(config.NormalConfig.GiteaToken))
 	print.Entry("GitHub connector type", format.OptionalStringerSetting(config.NormalConfig.GitHubConnectorType))
 	print.Entry("GitHub token", format.OptionalStringerSetting(config.NormalConfig.GitHubToken))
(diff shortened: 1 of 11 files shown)


james-harlyy:
@kevgo , I have added a "GithubUsername". Was debating whether to do this + a GitLab username, or to just use a generic branch-prefix configuration instead?

stephenwade:
In this comment, I explain why I think there should be a generic branch prefix config: https://github.com/git-town/git-town/issues/5677#issuecomment-3452118589

kevgo:
> I have added a "GithubUsername". Was debating whether to do this + a GitLab username, or to just use a generic branch-prefix configuration instead?

Something to keep in mind is that Git Town is used by a wide variety of teams in very different environments. We can't assume everyone follows the `<GitHub-username>/<branch-name>` convention. For example, on my projects I use `<initials>-<branch-name>` (e.g. `kg-feature`).

Because of that, I think a generic `BranchPrefix` config would be a better fit. That would let me set it to `kg-`, you set it to `james-harlyy/`, and someone working in a monorepo with lots of projects could set it to `<project>/<user>/` or whatever convention they use.

If some teams later find manual configuration too cumbersome, and it's cumbersome enough that enough people complain about it to us, we could explore ways to populate this setting automatically in a follow-up improvement.

Does that direction make sense to you?

PR #5705, opened 2025-11-04
Write stack footer in smaller font

kevgo:
Here is a preview of how this would look like (not the correct stack for this branch, just some random branches). I think it's better if this footer is a bit smaller. What do you think @yaadata?

 - main
   - https://www.github.com/git-town/pull/97 :point_left:
     - https://www.github.com/git-town/pull/98
       - https://www.github.com/git-town/pull/100

<sup>Stack generated by [Git Town](https://github.com/git-town/git-town)</sup>


internal/forge/proposal_stack_lineage_builder.go
```diff
@@ -59,7 +59,7 @@ func (self *ProposalStackLineageBuilder) Build(args ProposalStackLineageArgs) st
 	var builder strings.Builder
 	builder.WriteString("\n-------------------------\n")
 	builder.WriteString(self.build(self.tree.Node, args, 0))
-	builder.WriteString("\nStack generated by [Git Town](https://github.com/git-town/git-town)\n")
+	builder.WriteString("\n<sup>Stack generated by [Git Town](https://github.com/git-town/git-town)</sup>\n")
```
yaadata, line 62: I really like this.

internal/forge/proposal_stack_lineage_builder.go
```diff
@@ -59,7 +59,7 @@ func (self *ProposalStackLineageBuilder) Build(args ProposalStackLineageArgs) st
 	var builder strings.Builder
 	builder.WriteString("\n-------------------------\n")
```
yaadata, line 60: Unrelated, I know that we decided not to add any descriptive info before displaying the stack.

I now have real world data that when I first showed this in pull requests with coworkers, some did not know how to interpret or were new to stacked pull requests.

Do you still think that having descriptive text before the stack display is not warranted? See example below:

"Current Dependencies on/for this Pull-Request

main
   - pull_request_a 👈 
      - pull_request_b

Stack generated by Git Town"

kevgo: Thanks for sharing the real-world feedback, that's super helpful! I agree with the need to explain what this stack is, why it is there, and how to interpret it. What I'd like to avoid is adding such boilerplate to each and every pull request. It is only useful the first time somebody sees it but will always be there and add noise. 

Could we add a link to the footer that points to an explanation what this stack is? Here is how my example stack would look like with such a link:

 - main
   - https://www.github.com/git-town/pull/97 :point_left:
     - https://www.github.com/git-town/pull/98
       - https://www.github.com/git-town/pull/100

<sup>[Stack](https://www.git-town.com/stacked-changes.html#embed-the-stack-lineage-into-pull-requests) generated by [Git Town](https://github.com/git-town/git-town)</sup>

The word "Stack" is now a link that points to a page on [git-town.com](https://www.git-town.com) that explains in all detail what this stack display is. 

The source code for the website lives at https://github.com/git-town/git-town/tree/main/website. Feel free to make changes there as well.

Would this address your concern?

kevgo:
@yaadata if you agree with the changes made in this pull request, could you please approve it?

internal/forge/proposal_stack_lineage_builder.go
yaadata: This would address my concern! Could we add it in this PR? thanks @kevgo

yaadata:
@kevgo I approved the footer change.

internal/forge/proposal_stack_lineage_builder.go
kevgo: This change is outside the scope of the current branch, so I'll handle it in a separate one to keep the release history clean. I'll open that branch shortly.

kevgo:
Thanks for the review and suggestions for improvement! :+1:

PR #5742, opened 2025-11-06
docs: complete command summaries with missing flags

erik-rw:
Updates the command summaries in documentation to include all available command-line flags that were previously omitted:

- branch: Added -d|--display-types and -o|--order flags
- hack: Added -b|--beam, --stash|--no-stash, --sync, and --auto-resolve flags
- propose: Replaced -d|--detached with -s|--stack and added --auto-resolve flag
- switch: Added -o|--order flag

This ensures the command summaries accurately reflect all available options for these commands and helps users discover all functionality.

✅  Checked those changes locally by running `cd website && make serve`

### DIFF
--- a/website/src/commands/branch.md
+++ b/website/src/commands/branch.md
@@ -1,7 +1,7 @@
 # git town branch
 
 ```command-summary
-git town branch [-v | --verbose]
+git town branch [-d | --display-types] [-o | --order] [-v | --verbose]
 ```
 
 The _branch_ command is Git Town's equivalent of the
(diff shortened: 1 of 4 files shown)


kevgo (approved):
Thanks for jumping on this fix! 🙌 It’s always amazing how quickly docs can drift from reality, even when we *think* we're keeping them current. I just went through a similar cleanup yesterday and updated Go comments across 120 files in [this PR](https://github.com/git-town/git-town/pull/5725), so I definitely feel your pain. Nice work getting this back in sync!

PR #5757, opened 2025-11-09
Stack description should point to page defining stacks not the github action

yaadata:
The breadcumb `stack` link should point to a page that describes what a stack diff is. It should not point to the github action how-to page which is just one way to display that information on a pull-request


yaadata:
@kevgo could I get an approval here

kevgo:
Thanks for helping improve this! Here’s the idea behind the current setup — curious what you think:

* [Stacked Changes](https://www.git-town.com/stacked-changes.html) explains what stacked changes are and how to create them.
* [Breadcrumbs](https://www.git-town.com/how-to/github-actions-breadcrumb.html) explains how to embed stack lineage into proposals. Right now, it only mentions the GitHub Action. Once the CLI feature you built (adding lineage directly) goes public, we'll update this page to include that too.

When someone sees stack lineage in a proposal and wants to learn more, they can follow the link to the Breadcrumbs page, which should explain that this is stack lineage, and also link back to the Stacked Changes page to help people who don't even know what stacked changes are to learn about them.

So both pages should link to each other. Right now they don’t.

Does that make sense?

yaadata:
Makes sense. It was not intuitive to me why we are leading people who may not understand stack diffs to a How-To page about making breadcrumbs. 

Seems you have a plan for this page that I wasn't aware of going forward.

kevgo:
I agree that the term "breadcrumbs" is misleading. This page is about how to embed stack lineage into proposals. It needs to be updated.

PR #5779, opened 2025-11-17
Verify correctness of command-summary

### LINKED ISSUE #5743: Verify the commands on the website using Text-Runner
https://github.com/git-town/git-town/pull/5742 just fixed a substantial amount of missing documentation on the website. Verifying this could be automated using Text-Runner.

kevgo:
resolves #5743

Uses Text-Runner to verify that the command summaries always contain the correct
and complete set of CLI flags. Adds missing flags.

### DIFF
--- a/tools/ensure_no_files_with_dashes.sh
+++ b/tools/ensure_no_files_with_dashes.sh
@@ -14,6 +14,7 @@ files_with_dashes=$(find . -name '*-*' |
 	grep -v './dprint-changelog.json' |
 	grep -v './cucumber-sort.json' |
 	grep -v './.gherkin-*' |
+	grep -v './text-runner' |
 	grep -v 'tools(/|\\)rta')
 if [ -n "$files_with_dashes" ]; then
 	count=$(echo "$files_with_dashes" | wc -l)


kevgo:
@stephenwade FYI

website/src/commands/compress.md
```diff
@@ -1,9 +1,13 @@
 # git town compress
 
+<a type="command-summary">
+
 ```command-summary
-git town compress [-m <text> | --message <text>] [-s | --stack] [--dry-run] [-v | --verbose]
```
stephenwade, line 4: issue(minor): This change is not a big problem, but it feels wrong to me. It now seems to say that you can either run
`git town compress -m` or
`git town compress --message <text>`, but not
`git town compress -m <text>`.

I'd prefer either
`[-m <text> | --message <text>]` as originally written, or
`[(-m | --message) <text>]`.

website/src/commands/config.md
```diff
@@ -1,12 +1,13 @@
 # git town config
 
+<a type="command-summary">
+
 ```command-summary
-git town config [-v | --verbose]
-git town config get-parent [<branch-name>] [-v | --verbose]
-git town config remove [-v | --verbose]
```
stephenwade, line 6: issue: This PR removes the subcommands from the command summary.

website/src/commands/continue.md
```diff
@@ -1,9 +1,13 @@
 # git town continue
 
+<a type="command-summary">
+
 ```command-summary
-git town continue [-v | --verbose]
+git town continue [-h | --help] [-v | --verbose]
```
stephenwade, line 6: issue(minor): In this file, `[-h | --help]` was added before `[-v | --verbose]`. In most other files, it was added after `[-v | --verbose]`.

website/src/commands/down.md
```diff
@@ -1,9 +1,13 @@
 # git town down
 
+<a type="command-summary">
+
 ```command-summary
-git town down [-m | --merge] [-v | --verbose]
+git town down [-m | --merge] [-d | --display-types <type>] [-o | --order <asc|desc>] [-v | --verbose] [-h | --help]
```
stephenwade, line 6: issue(minor): Another issue where the added flags don't match the order of the doc. Ideally, the flags list should match the order of the doc, with the implied `[-h | --help]` at the end of the list.

ref: #5781

website/src/commands/hack.md
```diff
@@ -1,9 +1,13 @@
 # git town hack
 
+<a type="command-summary">
+
 ```command-summary
-git town hack [<branch-name>...] [-b | --beam] [-p | --prototype] [-d | --detached] [-c | --commit] [-m | --message <message>] [--propose] [--stash | --no-stash] [--sync] [--auto-resolve] [--dry-run] [-v | --verbose]
```
stephenwade, line 4: issue: This PR removes `--no-stash` from the command summary.

website/src/commands/help.md
```diff
@@ -1,7 +1,11 @@
 # git town help
 
+<a type="command-summary">
+
 ```command-summary
-git town help
+git town help [-h | --help]
```
stephenwade, line 6: question: Does `git town help` really have a `-h | --help` flag? Should it?

website/src/commands/status.md
```diff
@@ -1,10 +1,13 @@
 # git town status
 
+<a type="command-summary">
+
 ```command-summary
-git town status [-p | --pending] [-v | --verbose]
-git town status reset [-v | --verbose]
```
stephenwade, line 5: issue: This PR removed the subcommand from the command summary.

website/src/commands/compress.md
kevgo: I was worried about this as well. Went with the less verbose solution hoping it would be intuitive enough. Since you also see it as a concern, let's go with the more correct version. I have implemented this in https://github.com/git-town/git-town/pull/5811.

website/src/commands/config.md
kevgo: I don't see this page as a summary for command "git town config" and all its subcommands. To me, the summary is in the sidebar and on https://www.git-town.com/configuration-commands.html. This page describes the `git town config` command only. Subcommands like `git town config get-parent` have their own page.

website/src/commands/continue.md
kevgo: I ended up sorting all flags alphabetically. The Git documentation does it that way.  Alphabetic sorting avoids many bikeshed arguments and makes it easier to look up flags at a glance. Does this work for you?

website/src/commands/down.md
kevgo: Now that some commands have accumulated a pretty long list of flags, I’m starting to wonder if we should just sort them alphabetically. A few reasons:

* Folks who aren't familiar with all the flags need to scan the entire list anyway, so the specific ordering doesn’t help much.
* If you *are* looking for a particular flag, alphabetical ordering makes it easier and faster to find.
* We avoid spending time debating the "right" ordering. Any manual ordering we come up with will be subjective and probably contentious.
* Alphabetical ordering is trivial to enforce programmatically.

WDYT?

website/src/commands/hack.md
kevgo: `git town help hack` doesn’t list it either. The CLI framework the codebase uses doesn't support that. Text-Runner verifies that the command summary stays in sync with what the Git Town CLI reports. That way, when we add a new flag, it forces us to update the command summary on the website as well.

website/src/commands/help.md
kevgo: Yes:

```
> git town help hack

Flags:
      --auto-resolve     auto-resolve phantom merge conflicts
  -b, --beam             beam some commits from this branch to the new branch
  -c, --commit           commit the stashed changes into the new branch
  -d, --detached         don't update the perennial root branch
      --dry-run          print but do not run the Git commands
  -h, --help             help for hack
  -m, --message string   the commit message
      --propose          propose the new branch
  -p, --prototype        create a prototype branch
      --stash            stash uncommitted changes when creating branches
      --sync             sync branches (default true)
  -v, --verbose          display all Git commands run under the hood
```

This is built-in and I'm not aware of a way to change it. 

> Should it?

Why not? There is a man page for the `man` command as well :slightly_smiling_face:  The `help` command is a normal command with a certain amount of functionality. It seems reasonable to give users a way to learn how to use the help efficiently.

website/src/commands/compress.md
stephenwade: Looks good to me!

website/src/commands/continue.md
stephenwade: I'm okay with alphabetic sorting. I liked the idea of keeping the unique flags (mostly everything except `-h` and `-v`) at the beginning, but I can see how that leaves open a lot of room for bikeshedding.

website/src/commands/down.md
stephenwade: LGTM (https://github.com/git-town/git-town/pull/5779#discussion_r2562750000)

website/src/commands/config.md
stephenwade: Because they're subcommands, I feel they deserve a place in the command summary. I don't feel strongly about this, however, so I won't push to keep them.

website/src/commands/hack.md
stephenwade: Well, I think the CLI help should show it too, but that seems to be a separate issue. I'll file a new issue to track this, and I'm fine with the website matching the CLI output to allow this lint to exist.

stephenwade: #5814

website/src/commands/config.md
kevgo: The [page for the config command](https://www.git-town.com/commands/config.html) currently has a separate section `Subcommands` that lists the subcommands.

Listing them again in the summary for the `git town config` command is redundant. Given a choice, which one would you get rid of?

image

stephenwade: I'd keep the subcommands links.

kevgo: Thanks, sounds like this is resolved then.

PR #5808, opened 2025-11-24
Remove the ErrorCollector

kevgo:
The ErrorCollector is somewhat elegant, but an unnecessarily custom idiom that
contributors to this codebase aren't familiar with. Another thing they need to
learn in order to understand the codebase and contribute to it.

Go now provides a mechanism out of the box that achieves the same thing. It
requires a bit more boilerplate, but I think being idiomatic wins over
conciseness here. The code is still pretty readable.


kevgo:
This is too dangerous: an error can be silently assigned multiple times, there is no way to catch this. The existing linters aren't smart enough to catch this when it happens inside if clauses.

Issue #3629, opened 2024-06-11
.git/index.lock: File exists


rnystrom:
### Steps to reproduce

Unfortunately I have no idea how to reproduce this issue, or if I'm the only person experiencing it.

### Command output

```shell
(verbose) git version
git version 2.39.3 (Apple Git-146)

(verbose) git config -lz --global
filter.lfs.clean
git-lfs clean -- %ffilter.lfs.smudge
git-lfs smudge -- %ffilter.lfs.process
git-lfs filter-processfilter.lfs.required
trueuser.name
Ryan Nystromuser.email
rnystrom@whoisryannystrom.comalias.b
branchalias.co
checkoutalias.ci
commitalias.st
statusalias.lg
log --color --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commitalias.p
!git checkout main && git pull && git fetch --prune && git removed-branches --prune --force && git branchalias.append
town appendalias.hack
town hackalias.kill
town killalias.prepend
town prependalias.prune-branches
town prune-branchesalias.rename-branch
town rename-branchalias.sync
town syncalias.pr
town proposealias.sw
town switchalias.undo
town undoalias.cont
town continuealias.wip
!git add . && git commit --all --message wipcore.editor
/usr/local/bin/codeiniti.defaultbranch
main
(verbose) git config -lz --local
core.repositoryformatversion
0core.filemode
truecore.bare
falsecore.logallrefupdates
truecore.ignorecase
truecore.precomposeunicode
trueremote.origin.url
git@github.com:campsite/campsite.gitremote.origin.fetch
+refs/heads/*:refs/remotes/origin/*branch.main.remote
originbranch.main.merge
refs/heads/mainbranch.main.github-pr-owner-number
campsite#campsite#6388branch.main.github-pr-base-branch
campsite#campsite#ryan/cam-7134-note-comment-number-does-not-increment-as-new-comments-comegit-town.main-branch
maingit-town.perennial-branches
lfs.repositoryformatversion
0pull.rebase
truebranch.rn/member-n-1.github-pr-owner-number
campsite#campsite#4926git-town-branch.rn/single-upload-pipeline.parent
mainbranch.rn/single-upload-pipeline.remote
originbranch.rn/single-upload-pipeline.merge
refs/heads/rn/single-upload-pipelinegit-town-branch.rn/html-parser.parent
[...]

PR #5816, opened 2025-11-27
Fix retry bug in backend runner

meowsus:
I was reviewing https://github.com/git-town/git-town/issues/3629 because I've been experiencing this bug, still, in the most recent version of git-town. I'm new to git-town 👋 and am a graphite ex-pat. I just started a new job in a supermassive monorepo and wanted to start stacking my branches to help others review my code without having to stare at a 100+ file wall of change.

Anyway, enough about me. I was surprised to notice that the bug described in https://github.com/git-town/git-town/issues/3629 still persisted for me. I chalked it up to the repo being gargantuan and Cursor (vscode clone) having to do a lot more work to get its internal git state synced up behind the scenes.

When Cursor was closed, or when its git integration was disabled, I experienced no issues. When Cursor was open with its git integration enabled, when I was working on a stack of a significant size (3+ branches), running `git sync` would fail fairly consistently. Reviewing the issue, I noticed https://github.com/git-town/git-town/pull/3888 was proposed to try to solve this. When looking at the logs after a failure, it really seemed like the the retry functionality would bail after one retry attempt.

I am pretty sure I proved this with https://github.com/git-town/git-town/pull/5816/commits/be89f96633c0b9a88e40ab8882dbd58519cb1697 and fixed the issue with https://github.com/git-town/git-town/pull/5816/commits/84b7f75b54f4a13837d1972f07727b7aa95076d2

Let me know how you feel about it!

### DIFF
--- a/internal/subshell/retry_test.go
+++ b/internal/subshell/retry_test.go
@@ -19,7 +19,6 @@ func TestBackendRunner_RetryOnIndexLock(t *testing.T) {
 	t.Run("detects index.lock error pattern correctly", func(t *testing.T) {
 		t.Parallel()
 		tmpDir := t.TempDir()
-		runner := subshell.BackendRunner{Dir: Some(tmpDir), Verbose: false, CommandsCounter: NewMutable(new(gohacks.Counter))}
 
 		testCases := []struct {
 			name        string
@@ -61,9 +60,11 @@ func TestBackendRunner_RetryOnIndexLock(t *testing.T) {
 		for _, tc := range testCases {
 			t.Run(tc.name, func(t *testing.T) {
 				t.Parallel()
+				// Create a separate runner for each subtest to avoid data races
+				runner := subshell.BackendRunner{Dir: Some(tmpDir), Verbose: false, CommandsCounter: NewMutable(new(gohacks.Counter))}
 				scriptPath := filepath.Join(tmpDir, fmt.Sprintf("test-%s.sh", tc.name))
 				scriptContent := fmt.Sprintf(`#!/bin/bash
->&2 echo "%s"
+>&2 echo %q
 exit 1
 `, tc.errorMsg)
 				must.NoError(t, os.WriteFile(scriptPath, []byte(scriptContent), 0o600))
@@ -114,9 +115,17 @@ exit 1
 		tmpDir := t.TempDir()
 		runner := subshell.BackendRunner{Dir: Some(tmpDir), Verbose: false, CommandsCounter: NewMutable(new(gohacks.Counter))}
 
-		// Create a script that always fails with lock error
+		// Create a script that counts attempts and always fails with lock error
+		counterFile := filepath.Join(tmpDir, "attempt-counter")
 		scriptPath := filepath.Join(tmpDir, "always-fails.sh")
 		scriptContent := `#!/bin/bash
+COUNTER_FILE="` + counterFile + `"
+if [ ! -f "$COUNTER_FILE" ]; then
+    echo "1" > "$COUNTER_FILE"
+else
+    COUNT=$(cat "$COUNTER_FILE")
+    echo $((COUNT + 1)) > "$COUNTER_FILE"
+fi
 >&2 echo "fatal: Unable to create '.git/index.lock': File exists."
 exit 1
 `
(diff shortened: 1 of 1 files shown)


internal/subshell/backend_runner.go
```diff
@@ -52,16 +52,16 @@ func (self BackendRunner) execute(env []string, executable string, args ...strin
...
 	concurrentGitRetriesLeft := concurrentGitRetries
 	var outputText string
 	var outputBytes []byte
 	var err error
 	for {
+		subProcess := exec.CommandContext(context.Background(), executable, args...) // #nosec
+		subProcess.Env = append(subProcess.Environ(), env...)
+		if dir, has := self.Dir.Get(); has {
+			subProcess.Dir = dir
+		}
```
meowsus, line 64: This was the bug. The subprocess operation was outside of the loop. The [frontend runner](https://github.com/meowsus/git-town/blob/84b7f75b54f4a13837d1972f07727b7aa95076d2/internal/subshell/frontend_runner.go) got it right.

internal/subshell/retry_test.go
```diff
@@ -0,0 +1,343 @@
...
+	})
+
+	t.Run("retries and succeeds on transient lock error", func(t *testing.T) {
+		t.Parallel()
+		tmpDir := t.TempDir()
+		runner := subshell.BackendRunner{Dir: Some(tmpDir), Verbose: false, CommandsCounter: NewMutable(new(gohacks.Counter))}
+
+		// Create a script that fails twice with lock error, then succeeds
+		scriptPath := filepath.Join(tmpDir, "retry-script.sh")
+		scriptContent := `#!/bin/bash
```
kevgo, line 156: Pretty brilliant idea to use a Bash script that returns the error!

internal/subshell/retry_test.go
```diff
@@ -0,0 +1,343 @@
...
+		_, err := runner.Query("bash", scriptPath)
+		duration := time.Since(start)
+
+		// Should fail after exhausting retries
+		must.Error(t, err)
+		// Should take at least 4 seconds (5 attempts with 4 delays between them)
+		must.GreaterEq(t, 4*time.Second, duration)
+	})
+
+	t.Run("retries correct number of times", func(t *testing.T) {
```
kevgo: Would it make sense to combine this test with `exhausts retries and fails after max attempts` above? They both seem to fail until Git Town gives up.

internal/subshell/retry_test.go
```diff
@@ -0,0 +1,343 @@
...
+				errorMsg:    "fatal: not a git repository",
+				shouldRetry: false,
+			},
+		}
+
+		for _, tc := range testCases {
+			t.Run(tc.name, func(t *testing.T) {
+				scriptPath := filepath.Join(tmpDir, fmt.Sprintf("test-%s.sh", tc.name))
+				scriptContent := fmt.Sprintf(`#!/bin/bash
+>&2 echo "%s"
```
kevgo: You could let Go quote the string for you.

```suggestion
>&2 echo %q
```

internal/subshell/retry_test.go
```diff
@@ -0,0 +1,343 @@
...
+	})
+
+}
+
+func TestFrontendRunner_RetryOnIndexLock(t *testing.T) {
+	t.Parallel()
+
+	// NOTE: FrontendRunner DOES properly retry because the subprocess is created
+	// INSIDE the retry loop (unlike BackendRunner which has a bug).
+
```
kevgo: I think this comment may be obsolete now. Your PR makes `BackendRunner` and `FrontendRunner` behave the same. :slightly_smiling_face: 

```suggestion
```

internal/subshell/backend_runner.go
kevgo: Excellent catch! That definitely slipped through. I also use Cursor with Git integration, but somehow don't hit this concurrency issue. So buggy retry for `BackendRunner` went under the radar.

kevgo:
The new unit tests are tripping the race detector. You should be able to reproduce this locally by running `make unit-race`.

meowsus:
> Excellent catch! That definitely slipped through. I also use Cursor with Git integration, but somehow don't hit this concurrency issue. So buggy retry for BackendRunner went under the radar.

Thanks! I assume it's due to the size of the monorepo I'm working on. I was trying to learn the ergonomics of this project (which is so great, by the way) on a simple Next.js personal project I was working on and had no issues either. I'm just assuming that this bug is really evident in larger repos with tons of frequent code churn.

internal/subshell/retry_test.go
meowsus: I can't take credit for this approach but I thought it was pretty clever too! I'm very new to writing in Golang, so I had Claude help with most of the test work.

kevgo: Yeah, Claude comes up with surprisingly good solutions sometimes!

PR #5850, opened 2025-12-17
Redact user email

kevgo:
Email addresses are confidential too, so they get redacted as well.


kevgo:
This is now available as part of Git Town [22.4](https://github.com/git-town/git-town/releases/tag/v22.4.0).

PR #5869, opened 2025-12-22
propose: ability to skip a failing branch

kevgo:


kevgo:
This is now available as part of Git Town [22.4](https://github.com/git-town/git-town/releases/tag/v22.4.0).

Issue #3003, opened 2024-01-23
Create & sync stack lineage on proposals


tranhl:
## Problem
My team have bought into Git Town as our stacked workflow tool of choice. We're loving the workflow so far and the productivity gains from not being blocked. It's also drastically reduced PR size as contributors no longer feel the need to submit work in large chunks out of fear of being blocked. Huge wins overall 🎉 

However our one and only pain point is that we now have 30+ proposals open at any given time 😅. This makes it pretty hard to track which proposals belong to which stack. We like to review from the base up for any given stack, so this is a big point of friction for us during code review.

Our current method is to figure out what the base PR to any stack is (it'll be the one targeting the main branch), review it, take note of it's branch name, then find the PR that targets that branch. Rinse repeat until we're done. Refined Github makes this a little easier, but it's still not great:

![image](https://github.com/git-town/git-town/assets/24240419/c40ab69b-67ab-49cc-8dc0-448bde98ae27)

## Solution?

[graphite.dev](https://graphite.dev/) has a neat feature where it maintains a stack lineage for each proposal it creates in a stack.

![image](https://github.com/git-town/git-town/assets/24240419/d60e66e6-dc58-434e-a994-8b5e1a71d4be)

This allows your team to easily visualise & navigate through your stack on Github.

It would be super helpful if Git Town could provide a similar feature, perhaps built into `git-town sync` and `git-town propose`?

tranhl:
Thanks for getting back! Happy to answer your questions.

Let's use the stack from the example image above as reference:

> Do you review all PRs in a stack together, or only the top-level PR in a stack?

We start from the base (`[Gameboard] Initializing an empty gameboard` in this case). We'd then work our way up the stack. So next would be `[Gameboard] Add moving logic to gameboard`, and so on.

Ideally we review as often as PRs are being created, so that stacks don't blow up in size and cause merge conflict grief for everyone.

> Do you ship all PRs in a stack together, or do you always only the top-level PR?

The author would then ship the entire stack one PR at a time, starting from the base. This means `[Gameboard] Initializing an empty gameboard` would be shipped first.

> The stack you share in your screenshot looks like it's an ongoing "workstream" with more PRs to come soon. Because all you have right now is an empty gameboard with a controller and some styling. I imaging subsequent PRs will add additional game elements like rules, maps, leaderboards, etc.

[...]

PR #5913, opened 2025-12-26
Remove the rebuild code

kevgo:
part of #3003

Now that connectors are caching proposal data, we don't need to reuse existing
lineage anymore for their internal cache. Now we can re-create a new tree for
each branch.

### DIFF
--- a/internal/cmd/sync/sync_proposals.go
+++ b/internal/cmd/sync/sync_proposals.go
@@ -1,8 +1,6 @@
 package sync
 
 import (
-	"fmt"
-
 	"github.com/git-town/git-town/v22/internal/config/configdomain"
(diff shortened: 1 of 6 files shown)


internal/vm/opcodes/proposal_update_lineage.go
```diff
@@ -30,9 +30,6 @@ func (self *ProposalUpdateLineage) Run(args shared.RunArgs) error {
 	if !hasBuilder {
 		return nil
 	}
-	if err := builder.UpdateStack(lineageArgs); err != nil {
```
kevgo, line 33: Calling `proposallineage.NewBuilder` on line 29 should be enough to build the lineage here. Calling `builder.UpdateStack` on the same builder without having made any changes to the lineage does nothing.

PR #5924, opened 2025-12-30
Provide test dir

kevgo:
part of #3003

This PR adds configuration data used only in test mode to the regular configuration data infrastructure. The thinking behind this is to consolidate all configuration data into one location, which simplifies configuration data management.

In the new setup, the code documents that `TEST_HOME` is loaded only from environment variables.

-------------------------
- main
  - https://github.com/git-town/git-town/pull/5924 :point_left:
    - https://github.com/git-town/git-town/pull/5925

<sup>[Stack](https://www.git-town.com/how-to/github-actions-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>

### DIFF
--- a/internal/cli/dialog/switch_branch_test.go
+++ b/internal/cli/dialog/switch_branch_test.go
@@ -32,9 +32,9 @@ func TestNewSwitchBranch(t *testing.T) {
 					beta:  main,
 				})
 				branchInfos := gitdomain.BranchInfos{
-					gitdomain.BranchInfo{LocalName: Some(alpha), SyncStatus: gitdomain.SyncStatusLocalOnly},
-					gitdomain.BranchInfo{LocalName: Some(beta), SyncStatus: gitdomain.SyncStatusLocalOnly},
-					gitdomain.BranchInfo{LocalName: Some(main), SyncStatus: gitdomain.SyncStatusLocalOnly},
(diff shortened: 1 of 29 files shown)


kevgo:
This PR makes a questionable design decision. `NormalConfig` and `PartialConfig` only contain things that can be loaded from the various config sources (CLI, config file, git metadata, env). `TestDir` is not a proper configuration value. It is only loaded from an environment variable. It is only used in test mode. Wouldn't it make more sense to load and provide `TestDir` separately from the proper config data? Or is consistent handling of all config data, whether for testing or production, more important here?

kevgo:
> This PR makes a questionable design decision. `NormalConfig` and `PartialConfig` only contain things that can be loaded from the various config sources (CLI, config file, git metadata, env). `TestDir` is not a proper configuration value. It is only loaded from an environment variable. It is only used in test mode. Wouldn't it make more sense to load and provide `TestDir` separately from the proper config data? Or is consistent handling of all config data, whether for testing or production, more important here?

That's correct, and intentional. The goal is to handle all configuration data, for production and testing, the same way. So that there are fewer exceptions. Updated the PR description to clarify this.

kevgo:
No need for a TEST_HOME. The E2E tests already inject a custom HOME env var. Just use that and put the new connector data file next to the runstate and runlog. Keep injecting GIT_TOWN_TEST as a bool as it was done before.

PR #5959, opened 2026-01-11
fix: standard version shorthand is always a lowercase -v not -V

yaadata:
I believe this to be a mistake. I don't know of any cli that uses capital `-V` as the shorthand for version, rather the standard is lowercase `-v`

### DIFF
--- a/internal/cli/flags/version.go
+++ b/internal/cli/flags/version.go
@@ -8,7 +8,7 @@ const versionLong = "version"
 
 func Version() (AddFunc, ReadBoolFlagFunc) {
 	addFlag := func(cmd *cobra.Command) {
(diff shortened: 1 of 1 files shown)


kevgo:
Thanks for reaching out about this! I agree that `-v` used to be an industry standard. However, many modern CLI tools like `ripgrep`, `uv`, `dprint`, or `rustc` now use `-V` as shorthand for `--version` and reserve `-v` for `--verbose`. The rationale is that `--verbose` is typically used far more often than `--version`, so it gets the more ergonomic, lowercase flag. I'd like to follow this trend on Git Town. `git town --verbose` is quite useful while investigating issues.

PR #5960, opened 2026-01-12
chore: local claude settings should not be checked in

yaadata:
Per the official documentation, this file should not be checked in. I ran into some troubles with claude running stop hooks unexpectedly and I was wondering why. Individuals should use their own `.claude/settings.local.json` for experimentation and custom setup that doesn't need to be shared.

Screenshot 2026-01-11 at 6 08 03 PM
Screenshot 2026-01-11 at 6 11 00 PM


kevgo:
Thanks for fixing this! :problem

Issue #5919, opened 2025-12-28
verifying API calls in E2E tests


kevgo:
Git Town has a famously comprehensive end-to-end (E2E) test suite. The last notable gap is coverage for forge API calls, which currently require manual testing. That's not a sustainable approach long-term. Here is a proposal for adding E2E coverage for API interactions.

When Git Town runs in E2E test mode (i.e. `GIT_TOWN_TEST` is set), it would instantiate a forge-specific `MockConnector` (`MockGitHubConnector`, `MockGitLabConnector`) instead of the real forge connector. This mock wraps the real connector and overrides specific behaviors:

- It intercepts outbound API requests. Instead of sending them to the cloud, it records these requests to a JSON file and responds with pre-recorded data loaded from that same file. 
- It executes Git operations that simulate activities the forge does, like merging or deleting branches in the `origin` clone.

The mock connector would fail the test if either of the following occurs:

* Git Town makes an API call that is not documented in the E2E test
* Git Town makes a different API call than what is documented in the E2E test
* Git Town does not execute all the API calls documented in the E2E test

The JSON file would live alongside the existing E2E test artifacts, in the per-test directory that already contains the `development`, `origin`, and `upstream` clones.

kevgo:
Idea for how to define existing proposals. This data gets written to the JSON file and provides info for the replies to calls that search, lookup, merge, or update proposals.

```cucumber
Given these proposals exist:
  | ID | SOURCE    | TARGET    | TITLE     |
  | 1  | feature-1 | main      | Feature 1 | 
  | 2  | feature-2 | feature-1 | Feature 2 |
```

Connectors log activities they perform to the CLI. This allows to verify API calls made in the `Then Git Town runs the commands` step.

PR #5970, opened 2026-01-13
Initial implementation of E2E tests for proposal changes

kevgo:
part of #5919

Since this is a pretty large feature that touches many parts of the codebase, here comes the first super minimal implementation. It changes a single E2E test. I'd like to get this shipped before more iteration leads to even more complexity, which will be hard to review. Additional E2E tests will be changed in subsequent pull requests.


kevgo:
The tests are failing on CI because CI sets `XDG_CONFIG_HOME=/home/runner/.config`. The E2E setup writes `proposals.json` to `$HOME/.config/git-town`, but the Git Town executable reads it from `$XDG_CONFIG_HOME/git-town`.

Issue #5865, opened 2025-12-22
git town propose --stack fails consistently


meowsus:
### Steps to reproduce

I wanted to document this for posterity. I might be able to look into the issue myself, but I'm a little swamped at the moment.

Given the following stack:

```
  main
    branch-1
*     branch-2
```

Running `git town propose --stack` on the selected (*) branch locks me into a weird state with the following error message:

```
[branch-2] gh pr create --head=branch-1 --base=main
a pull request for branch "branch-1" into branch "main" already exists:
https://github.com/[REDACTED]/[REDACTED]/pull/4453


Error: exit status 1

To continue after having resolved conflicts, run "git town continue".
To go back to where you started, run "git town undo".
```

The state I'm left in is:

1. There is an open PR created for `branch-1`
2. This new PR wasn't opened in the browser
3. I'm back to being checked out on `branch-2` when the error occurs
4. `git town continue` reproduces the error above (each time its called)
5. `git town undo` closes the PR that was just silently created (expected behavior)

My workaround has been to create each PR manually in the stack, which isn't a pain, but isn't ideal behavior for this feature.

### Command output

```shell
$ git town propose --stack --verbose

(verbose) git version
git version 2.52.0

(verbose) git rev-parse --show-toplevel
/Users/[REDACTED]/Code/[REDACTED]

(verbose) git config -lz --global
user.name
[REDACTED]

user.email
[REDACTED]

core.editor
vim

alias.f
fetch

alias.co
checkout

alias.s
status -s

alias.cam
commit -am

alias.commend
commit --amend

alias.d
diff

alias.fuck
reset --hard HEAD

alias.pushup
push origin HEAD

alias.repull
pull --rebase

alias.ri
rebase -i

alias.cob
checkout -b

alias.cm
commit -m

alias.softfuck
reset --soft HEAD~1

alias.cp
cherry-pick

alias.append
town append

alias.compress
town compress

alias.contribute
town contribute

alias.diff-parent
town diff-parent

alias.down
town down

alias.hack
town hack

alias.delete
town delete

alias.observe
town observe

alias.park
town park

alias.prepend
town prepend

alias.propose
town propose

alias.rename
town rename

alias.repo
town repo

alias.set-parent
town set-parent

alias.ship
town ship

alias.sync
town sync

alias.up
town up

alias.tb
town branch

rerere.enabled
true


(verbose) git config -lz --local
core.repositoryformatversion
0

core.filemode
true

core.bare
false

core.logallrefupdates
[...]

PR #5973, opened 2026-01-14
Fix branch syncing bug around `git town propose --stack` command

meowsus:
Hey @kevgo! I finally got a second to attempt to fix the issue I reported here: #5865 (and that you helped fix [here](https://github.com/git-town/git-town/pull/5878)). Turns out it was just an OOO issue, but I added a test case to prove to myself that it was actually happening. Give it a whirl and let me know what you think :)

### DIFF
--- a/internal/cmd/propose.go
+++ b/internal/cmd/propose.go
@@ -384,6 +384,7 @@ func proposeProgram(repo execute.OpenRepoResult, data proposeData) program.Progr
 		prog.Value.Add(&opcodes.BranchTrackingCreateIfNeeded{
 			CurrentBranch: branchToPropose.name,
 		})
+		prog.Value.Add(&opcodes.CheckoutIfNeeded{Branch: branchToPropose.name})
 		prog.Value.Add(&opcodes.ProposalCreate{
 			Branch:        branchToPropose.name,
 			MainBranch:    data.config.ValidatedConfigData.MainBranch,
--- a/features/propose/stack/branch_checkout.feature
+++ b/features/propose/stack/branch_checkout.feature
@@ -1,4 +1,4 @@
-Feature: propose an entire stack
+Feature: propose stack checks out each branch before creating proposal
 
   Background:
     Given a Git repo with origin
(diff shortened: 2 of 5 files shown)


internal/cmd/propose.go
```diff
@@ -384,6 +384,7 @@ func proposeProgram(repo execute.OpenRepoResult, data proposeData) program.Progr
 		prog.Value.Add(&opcodes.BranchTrackingCreateIfNeeded{
 			CurrentBranch: branchToPropose.name,
 		})
+		prog.Value.Add(&opcodes.CheckoutIfNeeded{Branch: branchToPropose.name})
```
kevgo, line 387: Nice use of the `CheckoutIfNeeded` opcode here! This avoids unnecessary checkouts when the branch is already active

kevgo:
Thanks for fixing this, @meowsus! :pray: Yep, the fix itself is pretty straightforward: when running `git town propose --stack`, Git Town should first check out the branch being proposed, then create the proposal.

I was planning to land the new proposal-creation test framework before addressing this bug, but you beat me to it, which is excellent news 🙂

The test failures happen because a few other end-to-end tests now also check out the branch before proposing it, so they'll need to be updated to match the new behavior. You can run `make cuke` to see the test failures on your machine.

kevgo (approved):

meowsus:
Awesome that this bug is valid. Yesterday I used stack proposals twice and the bug didn't manifest itself, which caused me to second guess this bugfix. I thought that maybe I hadn't been upgraded to the latest version (v22.4.0) before when I last saw it, but I remembered that I started seeing the prompt to run `git town skip`, which was the prompt added in that version.

I'm updating the tests now :)

kevgo:
Thanks so much for fixing this! :+1:

meowsus:
Thanks so much for making the only viable, free, and open-source alternative to Graphite, in my opinion!

kevgo:
This is available in [v22.5](https://github.com/git-town/git-town/releases/tag/v22.5.0).

PR #5988, opened 2026-01-16
detach: correctly detach branch with multiple children

kevgo:


kevgo:
This is available in [v22.5](https://github.com/git-town/git-town/releases/tag/v22.5.0).

PR #6000, opened 2026-01-23
add the dry run before the sync to repro the bug

### LINKED ISSUE #5944: Sync command with amended commits in parent is failing in 22.4.0
### Steps to reproduce

Syncing amended parent was fixed as part of  https://github.com/git-town/git-town/issues/4586

When I tried to `dry-run` on 22.4.0 I get the following o/p.

Image

Please note the second dry run loses the paren't commit hash and executes the command `[child] git -c rebase.updateRefs=false rebase parent` 

On running the sync, I get the o/p with merge conflict as it was in https://github.com/git-town/git-town/issues/4586

Image

**Trying with v21.0.0**

I tried the same with 21 and I didn't see this issue. 

Image

I see the parent sha on every dry run.  

### Command output

```shell
gt sync -s --dry-run 

[child] git fetch --prune --no-tags
(dry run)

[child] git checkout parent
(dry run)

[parent] git push --force-with-lease --force-if-includes
(dry run)

[parent] git push --force-with-lease --force-if-includes
(dry run)

[parent] git push --force-with-lease --force-if-includes
(dry run)

[parent] git checkout child
(dry run)

[child] git -c rebase.updateRefs=false rebase --onto parent 441dd581f49e27ad892dbe2addaf659111bddea0
(dry run)


The Git Town configuration file contains the deprecated setting "push-new-branches".
Please upgrade to the new format: share-new-branches = "push"

 ~/personal/code/git-town-rebase-issue-demo > gt sync -s --dry-run

[child] git fetch --prune --no-tags
(dry run)

[child] git checkout parent
(dry run)

[parent] git push --force-with-lease --force-if-includes
(dry run)

[parent] git push --force-with-lease --force-if-includes
(dry run)

[parent] git push --force-with-lease --force-if-includes
(dry run)

[parent] git checkout child
(dry run)

[child] git -c rebase.updateRefs=false rebase parent
(dry run)
```

### Additional information

EDIT: My colleague mentioned he experiences problem with v21.5.0

pradeepmurugesan:
resolves #5944 

-------------------------
- fix-independent-amend-sync :point_left:

<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>

### DIFF
--- a/features/sync/features/dry_run.feature
+++ b/features/sync/features/dry_run.feature
@@ -27,9 +27,5 @@ Feature: dry run
       |         | git push                                          |
     And the initial branches and lineage exist now
     And the initial commits exist now
-
-  Scenario: undo
-    When I run "git-town undo"
-    Then Git Town runs no commands
-    And the initial branches and lineage exist now
-    And the initial commits exist now
+  #
+  # Cannot test undo because dry-run now doesn't create a runstate.


pradeepmurugesan:
@kevgo  I am able to reproduce the bug with an integration test. But couldn't really find why we are not using the `--onto` option here / where things go wrong. 

Can you kindly help me with a pointer to fix this issue 🙏

features/sync/stack/rebase_sync_strategy/commit_amended/independent_changes.feature
```diff
@@ -35,9 +35,22 @@ Feature: syncing a branch with independent changes where a commit was amended
       |          |               | branch-2 commit B | file      | line 0\n\nline 1: branch-1 content A\n\nline 2: branch-2 content B |
       |          | origin        | branch-2 commit A | file      | line 0\n\nline 1: branch-1 content A\n\nline 2: branch-2 content A |
     And the current branch is "branch-2"
-    When I run "git-town sync"
 
   Scenario: result
+    When I run "git-town sync --dry-run"
```
kevgo: Nice work reproducing this issue in an end-to-end test! This is a different test than the file you are modifying, so please move your change to a new `.feature` file:

- create a copy of this file and name it something like `features/sync/stack/rebase_sync_strategy/commit_amended/independent_changes_after_dry_run.feature`
- move your changes from the current file to the new file
- move the calls to `git-town sync --dry-run` and then `git-town sync` to the `Background` block

So the new file should have these steps (roughly):

```cucumber
Background:
  ...
  And the current branch is "branch-2"
  When I run "git-town sync --dry-run"
  Then Git Town runs the commands
  When I run "git-town sync"

Scenario: result
  Then Git Town runs the commands

kevgo:
I moved the test that reproduces the incorrect behavior after running `git sync --dry-run` into its own file and added comments that document the expected vs. actual behavior. This should make the bug easier to reason about and fix.

kevgo:
The fix here is to avoid persisting the runstate when `--dry-run` is enabled. With this change, dry-running a Git Town command becomes a true no-op: we only print what *would* happen, without making any changes to the hard drive.

This feels more consistent with the semantics of `--dry-run`. 

One implication is that dry-run commands can no longer be undone. Since they don't actually change anything, an undo is also always no-op, so this isn't a problem as well.

PR #6024, opened 2026-01-28
propose: add breadcrumb direction and style config options

luan:
## Summary

Adds two new configuration options for proposal breadcrumbs:

- `proposal-breadcrumb-direction`: `top-down` (default) or `bottom-up`
- `proposal-breadcrumb-style`: `tree` (default) or `auto` (flatten when linear)

I could see this being too much configuration though, happy to not have this if you prefer to have the more opinionated approach.

**Bottom-up** renders the stack with the current branch at top and root at bottom:
```
    - feature-2 👈
  - feature-1
- main
```

**Auto style** flattens to a flat list when the lineage has no branching:
```
- feature-2 👈
- feature-1
- main
```

## Configuration

TOML:
```toml
[propose]
direction = "up"
style = "auto"
```

Git config:
```bash
git config git-town.proposal-breadcrumb-direction up
git config git-town.proposal-breadcrumb-style auto
```

## Test plan

- [x] Unit tests pass
- [x] Lint passes
- [x] E2E tests for new options

### DIFF
--- a/internal/forge/forgedomain/proposal_breadcrumb_style.go
+++ b/internal/forge/forgedomain/proposal_breadcrumb_style.go
@@ -11,8 +11,8 @@ import (
 type ProposalBreadcrumbStyle string
 
 const (
-	ProposalBreadcrumbStyleTree ProposalBreadcrumbStyle = "tree"
-	ProposalBreadcrumbStyleAuto ProposalBreadcrumbStyle = "auto"
+	ProposalBreadcrumbStyleTree ProposalBreadcrumbStyle = "tree" // always render the breadcrumb as a tree, even if linear
+	ProposalBreadcrumbStyleAuto ProposalBreadcrumbStyle = "auto" // render the breadcrumb flat if linear, otherwise as a tree
 )
 
 func (self ProposalBreadcrumbStyle) String() string {


kevgo:
Hey @luan thanks so much for implementing these two settings, they are super useful! 🙏

kevgo:
It looks like when the new setting `proposal-breadcrumb-direction` is set to `bottom-up`, it should probably force `proposal-breadcrumb-style` to `auto`. The reason is that rendering an inverted indented list doesn't work on GitHub.

For example, this Markdown:

```md
      - three
   - two
- one
```

gets rendered like this:

      - three
   - two
- one


So `bottom-up` + `tree` is broken in GitHub-flavored Markdown.

What do you think?

kevgo:
Thoughts about simplifying the values `top-down` and `bottom-up` to just `down` and `up`?

When set to `down` (down from the root):

```
- main
  - feature-1
    - feature-2 👈
```

When set to `up` (up from the root):

```
- feature-2 👈
- feature-1
- main
```

kevgo:
@luan I assume you want to use the `bottom-up` direction, otherwise you wouldn't have made this PR. 🤔 

If the `bottom-up` direction forces the linear style (as proposed above), do we need setting `proposal-breadcrumb-style` at all? It seems like it would only be relevant for users who want `top-down` with the linear style.

I'm inclined to leave `proposal-breadcrumb-style` out for now and only introduce it if there is actual user demand with a good justification for the additional complexity. Every user-visible config option adds to Git Town's cognitive overhead and learning curve, which we want to keep as low as possible.

internal/forge/forgedomain/proposal_breadcrumb_direction_test.go
```diff
@@ -0,0 +1,28 @@
...
+func TestParseProposalBreadcrumbDirection(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		give string
+		want Option[forgedomain.ProposalBreadcrumbDirection]
+		err  bool
+	}{
+		{give: "", want: None[forgedomain.ProposalBreadcrumbDirection](), err: false},
+		{give: "top-down", want: Some(forgedomain.ProposalBreadcrumbDirectionTopDown), err: false},
+		{give: "bottom-up", want: Some(forgedomain.ProposalBreadcrumbDirectionBottomUp), err: false},
```
kevgo: Please add uppercase test cases

internal/setup/enter.go
```diff
@@ -279,53 +279,55 @@ EnterForgeData:
...
+		MainBranch:                  mainBranchResult.UserChoice,
+		NewBranchType:               newBranchType,
+		ObservedRegex:               observedRegex,
+		Offline:                     None[configdomain.Offline](), // the setup assistant doesn't ask for this
+		Order:                       order,
+		PerennialBranches:           perennialBranches,
+		PerennialRegex:              perennialRegex,
+		ProposalBreadcrumb:          proposalBreadcrumb,
+		ProposalBreadcrumbDirection: None[forgedomain.ProposalBreadcrumbDirection](), // the setup assistant doesn't ask for this
+		ProposalBreadcrumbStyle:     None[forgedomain.ProposalBreadcrumbStyle](),     // the setup assistant doesn't ask for this
```
kevgo: Please add a TODO so that I don't forget to add this to the setup assistant later.

```suggestion
		ProposalBreadcrumbDirection: None[forgedomain.ProposalBreadcrumbDirection](), // TODO: add this to the setup assistant
		ProposalBreadcrumbStyle:     None[forgedomain.ProposalBreadcrumbStyle](),     // TODO: add this to the setup assistant
```

luan:
Ahh good point on the markdown render. I hadn't noticed that.

Think I've addressed all of your feedback, thanks for getting back to me so quickly!

internal/config/envconfig/load.go
```diff
@@ -133,53 +133,55 @@ func Load(env EnvVars) (configdomain.PartialConfig, error) {
...
+		MainBranch:                  gitdomain.NewLocalBranchNameOption(env.Get(mainBranch)),
+		NewBranchType:               configdomain.NewBranchTypeOpt(newBranchType),
+		ObservedRegex:               observedRegex,
+		Offline:                     offline,
+		Order:                       order,
+		PerennialBranches:           gitdomain.ParseLocalBranchNames(env.Get(perennialBranches)),
+		PerennialRegex:              perennialRegex,
+		ProposalBreadcrumb:          proposalBreadcrumb,
+		ProposalBreadcrumbDirection: None[forgedomain.ProposalBreadcrumbDirection](), // not loaded from env vars
+		ProposalBreadcrumbStyle:     None[forgedomain.ProposalBreadcrumbStyle](),     // not loaded from env vars
```
kevgo: Please also add these todos.

```suggestion
		ProposalBreadcrumbDirection: None[forgedomain.ProposalBreadcrumbDirection](), // TODO: load this from the env vars
		ProposalBreadcrumbStyle:     None[forgedomain.ProposalBreadcrumbStyle](),     // TODO: load this from the env vars
```

internal/forge/forgedomain/proposal_breadcrumb_style.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/git-town/git-town/v22/internal/messages"
+	. "github.com/git-town/git-town/v22/pkg/prelude"
+)
+
+type ProposalBreadcrumbStyle string
+
+const (
+	ProposalBreadcrumbStyleTree ProposalBreadcrumbStyle = "tree"
+	ProposalBreadcrumbStyleAuto ProposalBreadcrumbStyle = "auto"
```
kevgo: Let's edit some basic documentation for these settings as the values aren't self-explanatory to the uninitiated.

```suggestion
	ProposalBreadcrumbStyleTree ProposalBreadcrumbStyle = "tree" // always render the breadcrumb as a tree, even if linear
	ProposalBreadcrumbStyleAuto ProposalBreadcrumbStyle = "auto" // render the breadcrumb flat if linear, otherwise as a tree
```

internal/proposallineage/render_tree.go
```diff
@@ -3,22 +3,37 @@ package proposallineage
...
-	renderNode(&builder, tree, currentBranch, 0, false)
+
+	flat := style == forgedomain.ProposalBreadcrumbStyleAuto && tree.IsLinear()
+
+	switch direction {
+	case forgedomain.ProposalBreadcrumbDirectionDown:
+		renderNodeDown(&builder, tree, currentBranch, 0, false, flat)
+	case forgedomain.ProposalBreadcrumbDirectionUp:
+		maxDepth := tree.MaxDepth()
+		renderNodeUp(&builder, tree, currentBranch, maxDepth, flat)
```
kevgo, line 23: Since the `up` format is always flat, we shouldn't need the `flat` parameter here. Can we remove it?


```suggestion
		renderNodeUp(&builder, tree, currentBranch, maxDepth)
```

kevgo:
Rather than going back and forth on minutiae, I'm going to merge this in and address the remaining nitpicks as part of adding end-to-end tests myself.

kevgo:
This is now available as part of [v22.6](https://github.com/git-town/git-town/releases/tag/v22.6.0). Enjoy!

Issue #5995, opened 2026-01-18
allow non-TTY usage (AI / automation)


kevgo:
Git Town currently requests missing information (e.g. parent branches) via interactive dialogs. This works well for humans, but makes Git Town unusable in non-interactive environments such as:

- execution from an AI assistant in a subshell
- CI jobs
- other automation where no TTY is available

To support this use case better, Git Town should:

- [ ] exit with a descriptive error message if it needs information but cannot open a TTY
- [ ] allow to provide all needed information (parent branches, branch type) via CLI arguments

related: 

- https://github.com/git-town/git-town/discussions/5928
- https://github.com/git-town/git-town/issues/5967

xmarkclx:
+1. 

I want to propose this as well.
Some of the commands work though, so we may need a list of commands which doesn't.

It's been some time but I think I had this issue with
`propose`

kevgo:
These dialogs can show up for any Git Town command if data is missing:

- when there is an unfinished runstate: `internal/cli/dialog/unfinished_run_state.go` 
- when the branch type isn't known: `internal/cli/dialog/unknown_branch_type.go`
- when the parent branch isn't known: `internal/cli/dialog/lineage.go`
- when the main branch isn't set: `internal/setup/enter.go`

kevgo:
A first step is displaying a more meaningful error message when these functions return an error. The error message should mention a way to provide the needed information programmatically before retrying the command:

- provide another Git Town command call that sets the needed information
- provide the CLI flag to call the failed Git Town command with that provides the needed information

bendrucker:
Working on a change to catch `ErrNoTTY` in the two remaining paths and provide descriptive messages:

- **Lineage** (`cli/dialog/lineage.go`): include the branch name and suggest `git town set-parent` / `git config git-town-branch.<branch>.parent`
- **Unfinished state** (`validate/handle_unfinished_state.go`): suggest `git town continue`, `skip`, `undo`, or `status reset`

This would complete the error message coverage, with additional flags to be discussed/added separately.

pcfreak30:
I want to see this as well. I semi have my ai use git already but cant automate the propose command and I prefer using the `gh` tool backend. Having it auto take the detected defaults, and support things like discard the last state as a flag would be awesome.

kevgo:
@pcfreak30 can you please explain in more detail what you mean with this statement:

> Having it auto take the detected defaults, and support things like discard the last state as a flag would be awesome

[...]

PR #6051, opened 2026-02-08
Detect missing TTY before launching interactive dialogs

bendrucker:
When Git Town commands need interactive input but no TTY is available (CI pipelines, piped input, AI agent environments), BubbleTea crashes with `open /dev/tty: device not configured`. This adds TTY detection at the dialog component level so users get a clear, actionable error message instead.

## Changes

- Add a `RequireTTY()` guard to each dialog primitive (`RadioList`, `CheckList`, `TextField`, `TextDisplay`, `SwitchBranch`) that checks for an interactive terminal before launching a BubbleTea program
- TTY detection uses [`go-isatty`](https://github.com/mattn/go-isatty) (already a transitive dependency via termenv) for the primary stdin check, including Cygwin/MSYS2 terminal support on Windows, with a `/dev/tty` (Unix) or `CONIN$` (Windows) fallback for piped-stdin-with-controlling-terminal scenarios
- When the main branch is not configured and no TTY is available, return an error guiding users to either `git town init` or `git config git-town.main-branch <branch>`
- Guard `fmt.Printf` calls in `main_branch.go` and `unfinished_run_state.go` that previously ran unconditionally after dialog calls, printing spurious output on error

## Context

Git Town works fully non-interactively when configured — dialogs only trigger as configuration fallbacks when the main branch or parent lineage is missing. All preferences support `GIT_TOWN_*` environment variables and `.git-town.toml` config files, making Git Town usable in CI and by AI agents when properly set up. This change focuses on the most common first-contact failure: the setup wizard crash when the main branch isn't configured.

### DIFF
--- a/internal/validate/config.go
+++ b/internal/validate/config.go
@@ -40,7 +40,7 @@ func Config(args ConfigArgs) (config.ValidatedConfig, dialogdomain.Exit, error) 
 			return config.EmptyValidatedConfig(), exit, err
 		}
 		if exit {
-			return config.EmptyValidatedConfig(), exit, err
+			return config.EmptyValidatedConfig(), exit, nil
 		}
 		err = setup.Save(userInput, args.Unvalidated.Immutable(), setupData, enterAll, args.Frontend)
 		if err != nil {


internal/validate/config.go
```diff
@@ -30,7 +33,13 @@ func Config(args ConfigArgs) (config.ValidatedConfig, dialogdomain.Exit, error)
...
 		userInput, exit, enterAll, err := setup.Enter(setupData, args.ConfigDir)
-		if err != nil || exit {
+		if err != nil {
+			if errors.Is(err, dialogcomponents.ErrNoTTY) {
+				return config.EmptyValidatedConfig(), false, errors.New(messages.NoTTYMainBranchMissing)
+			}
+			return config.EmptyValidatedConfig(), exit, err
+		}
+		if exit {
 			return config.EmptyValidatedConfig(), exit, err
```
kevgo: Nitpick: At this point the error is now already handled.

```suggestion
			return config.EmptyValidatedConfig(), exit, nil
```

kevgo (approved):
Thanks for this excellent contribution! 🙌 This is the missing piece needed to ship #5995. I put together some end-to-end tests for this yesterday, so the timing couldn't have been better.

Really nice work building this so clean and idiomatically!

kevgo:
@bendrucker welcome to Git Town! I've sent you an invite to our GitHub org so you can work directly in the main repo instead of using a fork.

kevgo:
This is now available as part of [v22.6](https://github.com/git-town/git-town/releases/tag/v22.6.0). Enjoy!

Issue #1201, opened 2019-05-01
Please respect TERM=dumb


zenspider:
I haven't looked into the code too much yet (not a go coder (yet?)) but the menuing on `git town config setup` (or when auto-invoked) expects full VT-xx/xterm support, which is not available when TERM=dumb or TERM=emacs (for example). Either `os.GetEnv("TERM")` should be queried and decisions made based on that or actual term capability should be queried.

If you point me in the right direction, I can do a (possibly fake) PR with the idea behind it.

kevgo:
Hey @zenspider! There are two elements of Git Town output that could cause problems here:

1. formatting branches in bold text: I think we enforce color at all times because our end-to-end tests need to see them. We never had issues with this setup, but I can see an adjustment if needed, i.e. add configuration to omit colors.
2. complex UI when configuring Git Town: we use https://github.com/AlecAivazis/survey in this case. The only alternative I can think of is providing command-line switches for the config variables. We had that in the past but decided that a nice terminal UI is nicer. 

Can you get the Git Town configuration done in a "real" terminal outside of Emacs? If that's a problem, please share a screenshot of the broken output.

zenspider:
The bold codes are stripped by emacs (in my config, not by default).

The main problem is #2, that menu picker thing. I would think defaulting to plain string reads would be fine, no? (easy for the branching question, not as easy but still doable for the perennials question).

zenspider:
I'm also happy to file the same ticket on the survey project and link back to here if you'd rather delegate it off instead of work around.

kevgo:
Yeah, I think this is an upstream issue. Git Town is just a user of survey. I didn't find any configuration options around fallback to simpler input on dumber terminals in their documentation. So this seems a survey issue.

zenspider:
It _can be_ an upstream issue. It can also be applied here by detecting TERM and using plain IO reads. Since upstream hasn't even responded in 2 months, I'm guessing any traction will have to be here.

zenspider:
I walked away when the maintainers punted... Nice idea, bad UX.

kevgo:
Apologies for a misconfigured stale-bot closing this when it totally shouldn't have.

PR #6052, opened 2026-02-08
Descriptive no-TTY errors for lineage and unfinished state dialogs

bendrucker:
Extends the `ErrNoTTY` handling from #6051 to the two remaining dialog points, completing descriptive error message coverage for non-interactive terminals.

## Issue

Related to #5995 (task 1 — descriptive error messages).

When Git Town needs information but cannot open a TTY, the lineage and unfinished state dialogs propagated the raw `ErrNoTTY` without actionable guidance. #6051 solved this for the main branch case; this PR covers the remaining paths.

## Changes

- **`cli/dialog/lineage.go`**: Catch `ErrNoTTY` from `SwitchBranch()` and return a message including the branch name and remediation commands (`git town set-parent`, `git config git-town-branch.<branch>.parent`)
- **`validate/handle_unfinished_state.go`**: Catch `ErrNoTTY` from `AskHowToHandleUnfinishedRunState()` and return a message with the interrupted command name and resolution options (`continue`, `skip`, `undo`, `status reset`)
- **`messages/en.go`**: Add `NoTTYParentBranchMissing` and `NoTTYUnfinishedState` message constants

## Testing

Two new Cucumber scenarios verify the error messages:
- `no_tty_unknown_parent.feature` — syncing a branch with no configured parent
- `no_tty_unfinished_state.feature` — syncing when a prior command was interrupted by a merge conflict

To make these tests portable across environments, a `GIT_TOWN_NO_TTY` env var was added to `HasTTY()`. On Linux CI, `/dev/tty` is accessible even without a real terminal (the CI runner uses `script` to create a PTY), so `HasTTY()` returns `true` and dialog subprocesses hang waiting for input. The alternative — using `Setsid` to detach from the controlling terminal — was not viable because `RequireTTY()` is checked before dialog test inputs are consumed, so detaching would break all existing dialog tests. The test runner now sets `GIT_TOWN_NO_TTY=1` for commands that have no `GITTOWN_DIALOG_INPUT` env vars, ensuring the no-TTY path is reliably triggered without affecting dialog tests.

### DIFF
--- a/features/sync/features/no_tty_unfinished_state.feature
+++ b/features/sync/features/no_tty_unfinished_state.feature
@@ -1,0 +1,24 @@
+Feature: descriptive error when there is unfinished state and no TTY is available
+
+  Background:
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the commits
+      | BRANCH  | LOCATION | MESSAGE                   | FILE NAME        | FILE CONTENT   |
+      | feature | local    | conflicting local commit  | conflicting_file | local content  |
+      |         | origin   | conflicting origin commit | conflicting_file | origin content |
+    And the current branch is "feature"
+    When I run "git-town sync"
+
+  Scenario: sync with unfinished state and no TTY
+    When I run "git-town sync"
+    Then Git Town prints the error:
+      """
+      there is an unfinished "sync" command and no interactive terminal available to resolve it.
+      To continue the command, run: git town continue
+      To skip the current branch, run: git town skip
+      To undo the command, run: git town undo
+      To discard the unfinished state, run: git town status reset
+      """
--- a/features/sync/features/no_tty_unknown_parent.feature
+++ b/features/sync/features/no_tty_unknown_parent.feature
@@ -1,0 +1,15 @@
+Feature: descriptive error when parent is unknown and no TTY is available
+
+  Scenario: sync a branch with unknown parent and no TTY
+    Given a Git repo with origin
+    And the branches
+      | NAME  | TYPE   | LOCATIONS     |
+      | alpha | (none) | local, origin |
+    And the current branch is "alpha"
+    When I run "git-town sync"
+    Then Git Town prints the error:
+      """
+      no parent configured for branch "alpha" and no interactive terminal available.
+      To set the parent, run: git town set-parent <parent>
+      To configure manually, run: git config git-town-branch.alpha.parent <parent>
+      """
--- a/internal/cli/dialog/dialogcomponents/tty.go
+++ b/internal/cli/dialog/dialogcomponents/tty.go
@@ -10,8 +10,14 @@ import (
 // ErrNoTTY indicates that an interactive terminal is required but not available.
 var ErrNoTTY = errors.New("no interactive terminal available")
 
+// NoTTYEnvVar allows overriding TTY detection via an environment variable.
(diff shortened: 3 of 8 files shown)


bendrucker:
Fix~ing~ed the E2E issue, they passed locally but fail in CI because `HasTTY` ends up returning true. Adding an environment variable to opt into no-TTY behavior even when a TTY is available.

bendrucker:
@kevgo Do you want to review this? I realize I have merge access now but I figured you'd want to look.

kevgo:
Apologies for the delay! I'm currently exploring an alternative approach where the end-to-end tests invoke the Git Town executable through a subshell without TTY. The main benefit is that we can test the no-TTY behavior naturally, without having to simulate it via the `GIT_TOWN_NO_TTY` env variable.

I,d like to see whether this approach pans out before merging this PR. Please keep it open for now, we may still end up going with the solution proposed here if my alternative doesn't work out.

Feel free to merge smaller changes (typos, minor refactors) on your own as long as the tests pass. For larger changes with architectural impact, like this PR, I'd still prefer a review.

Thanks!

bendrucker:
> I'm currently exploring an alternative approach where the end-to-end tests invoke the Git Town executable through a subshell without TTY. 

Nice, I hit a wall with doing this surgically (Setsid), but if you can make a broader change to the test runner to enable it without that hack, that is definitely better.

To the extent there's any deserving env var, I usually like when tools have `NO_INTERACTIVE` or similar. Sometimes you have a TTY, but you don't want interactive prompts. And if you're having to using `NO_TTY=1` to force that, that's a forcing the user into a hack.

If you don't end up changing the test runner, I could make that change here so this is more of an explicit self-documenting feature and less of a leaking test concern.

kevgo:
Claude Code found a way to make `Setsid` work on Linux and macOS. That's enough to verify that the Git Town executable prints the correct error messages when there is no TTY available.

See https://github.com/git-town/git-town/pull/6068 for my implementation.

The bulk of the effort was adding comprehensive end-to-end tests to document how each Git Town command behaves in a non-TTY environment.

kevgo:
I really like the idea of supporting an env var like `NO_INTERACTIVE`.

It would be great if we could lean on an existing convention rather than inventing something new. Is `NO_INTERACTIVE` widely used? A quick Google search didn't reveal much.

#1201 proposes to rely on `TERM=dumb`, which seems to be a fairly well-established signal that the environment doesn't support interactive/TTY features. If set, Git Town wouldn't prompt via dialogs, but print what is missing and how to provide it via CLI flags.

Would using `TERM=dumb` cover your use case? If so, I'd prefer aligning with that instead of introducing something proprietary.

bendrucker:
`TERM=dumb` sounds fine, feels familiar but didn't come to mind. No particular attachment to any key/value. Automatically handling no-TTY is much more important, the env var is just a fallback for exceptional cases.

kevgo:
I have implemented support for `TERM=dumb` in https://github.com/git-town/git-town/pull/6088. Can this PR be closed?

kevgo:
Thanks again for this valuable contribution. It ended up inspiring a whole set of improvements, all of which are now implemented:

* Git Town now automatically disables interactive features when running in a non-TTY environment.
* E2E tests now create a no-TTY subshell directly, without the need for an env var
* Interactive features can now be disabled manually via `TERM=dumb`

Given that, I'm going to close this PR since there aren't any remaining actionable items left. I consider this a successful contribution since all its ideas made it into the product. :pray:

PR #6053, opened 2026-02-08
Redirect golangci-lint cache to sandbox-writable path

bendrucker:
The Claude Code sandbox blocks writes to `~/Library/Caches/golangci-lint`, causing `golangci-lint cache clean` (run by `make lint`) to fail. This created an infinite loop with the Stop hook: lint fails, failure reported to Claude, Claude responds, Stop hook fires again.

Redirects the `golangci-lint` cache to `/tmp/claude/golangci-lint` via the `GOLANGCI_LINT_CACHE` env var. This path is already in the sandbox write allowlist and is portable across platforms.

https://golangci-lint.run/docs/configuration/cli/#cache

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,7 +1,8 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-settings.json",
   "env": {
-    "DISABLE_NON_ESSENTIAL_MODEL_CALLS": "1"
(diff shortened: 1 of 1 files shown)


kevgo:
Thanks! Being able to run in the Claude Code sandbox is important.

PR #6090, opened 2026-03-08
Update golangci-lint

kevgo:


kevgo:
Thanks for getting this over the finish line! 🙏

PR #6106, opened 2026-03-23
Add --headless flag to propose command

alexgeek:
## Summary

- Add `--headless` flag and `propose.headless` config setting that skips opening a browser when creating proposals
- Useful for CI/CD pipelines and headless environments
- Configurable via CLI flag, config file (`[propose] headless = true`), git metadata, or env var (`GIT_TOWN_PROPOSE_HEADLESS`)
- CLI connectors (gh/glab) create the proposal without opening the browser
- Web-only connectors return an error since they require a browser to create proposals


kevgo:
> Web-only connectors return an error since they require a browser to create proposals

Instead of erroring out, we could fall back to printing the URL that would create the proposal and then exit. This would let the user complete the flow manually.

This would match the [existing behavior](https://github.com/git-town/git-town/blob/main/features/propose/browser/none.feature) when Git Town can't find a browser executable.

kevgo:
To fix the failing tests, please run `git town sync` on this branch. I recently updated some of the test infrastructure.

internal/cli/flags/propose_headless.go
```diff
@@ -0,0 +1,23 @@
...
+	. "github.com/git-town/git-town/v22/pkg/prelude"
+	"github.com/spf13/cobra"
+)
+
+const proposeHeadlessLong = "headless"
+
+// type-safe access to the CLI arguments of type configdomain.ProposeHeadless
+func ProposeHeadless() (AddFunc, ReadProposeHeadlessFlagFunc) {
+	addFlag := func(cmd *cobra.Command) {
+		cmd.Flags().BoolP(proposeHeadlessLong, "", false, "create the proposal without opening a browser")
```
kevgo: Since this flag has no shorthand, you can simplify its definition:

```suggestion
		cmd.Flags().Bool(proposeHeadlessLong, false, "create the proposal without opening a browser")
```

internal/config/configdomain/propose_headless.go
```diff
@@ -0,0 +1,4 @@
+package configdomain
+
+// ProposeHeadless indicates whether the propose command should skip opening a browser.
+type ProposeHeadless bool
```
kevgo, line 4: What do you think about renaming this flag to just `Headless`?

My thinking is that the flag's purpose is to indicate headless mode (i.e. no browser launch). Right now it's only implemented for the `propose` command, but it feels like something we'll also want to make available in the other commands that open browsers, like [git town repo](https://www.git-town.com/commands/repo.html).

If we end up adding this flag to the `git town repo` command, then it no longer makes sense to define the flag in the `propose` config section. A more appropriate place could be  the `hosting` section of the [configuration file](https://www.git-town.com/configuration-file.html).

```toml
[hosting]
headless = true
```

```sh
git config git-town.hosting-headless true
```

Thoughts?

internal/forge/azuredevops/web_connector.go
```diff
@@ -28,7 +29,12 @@ func (self WebConnector) BrowseRepository(runner subshelldomain.Runner) error {
...
 func (self WebConnector) CreateProposal(data forgedomain.CreateProposalArgs) error {
-	browser.Open(self.NewProposalURL(data), data.FrontendRunner, self.browser)
+	proposalURL := self.NewProposalURL(data)
+	if data.Headless {
+		fmt.Printf(messages.BrowserOpen, proposalURL)
+		return nil
+	}
+	browser.Open(proposalURL, data.FrontendRunner, self.browser)
 	return nil
 }
```
kevgo, line 39: Nitpick: very tiny possibility to make the intent a bit more clear here

```suggestion
	if data.Headless {
		fmt.Printf(messages.BrowserOpen, proposalURL)
	} else {
		browser.Open(proposalURL, data.FrontendRunner, self.browser)
	}
	return nil
}
```

internal/forge/gh/connector.go
```diff
@@ -55,6 +55,9 @@ func (self Connector) CreateProposal(data forgedomain.CreateProposalArgs) error
 	if err := self.Frontend.Run("gh", args...); err != nil {
 		return err
 	}
+	if data.Headless {
+		return nil
```
kevgo, line 59: Rather than showing nothing here, you could run `gh pr view` (without `--web` at the end). This displays proposal information in the CLI.

Example output for one of my pull requests:

```
> gh pr view
Give functions with excessive argument counts named arguments git-town/git-town#5308
Merged • kevgo wants to merge 7 commits into main from kg-named-args • about 8 months ago
+121 -47 • ✓ Checks passing


  No description provided


View this pull request on GitHub: https://github.com/git-town/git-town/pull/5308
```

internal/forge/glab/connector.go
```diff
@@ -53,7 +53,11 @@ func (self Connector) CreateProposal(data forgedomain.CreateProposalArgs) error
 	if !hasTitle || !hasBody {
 		args = append(args, "--fill")
 	}
-	args = append(args, "--web")
+	if data.Headless {
+		args = append(args, "--yes")
```
kevgo, line 57: Great choice to use `--yes` here!

internal/setup/enter.go
```diff
@@ -321,6 +321,7 @@ EnterForgeData:
 		PerennialRegex:              perennialRegex,
 		ProposalBreadcrumb:          proposalBreadcrumb,
 		ProposalBreadcrumbDirection: proposalBreadcrumbDirection,
+		ProposeHeadless:             None[configdomain.ProposeHeadless](), // the setup assistant doesn't ask for this
```
kevgo, line 324: Let's add this to the setup assistant, but okay to do separately from this change.

```suggestion
		ProposeHeadless:             None[configdomain.ProposeHeadless](), // TODO: implement this
```

internal/vm/opcodes/proposal_create.go
```diff
@@ -47,6 +51,7 @@ createProposal:
 	err := connector.CreateProposal(forgedomain.CreateProposalArgs{
 		Branch:         self.Branch,
 		FrontendRunner: args.Frontend,
+		Headless:       bool(args.Config.Value.NormalConfig.ProposeHeadless),
```
kevgo, line 54: `forgedomain.CreateProposalArgs` is under our control, so we should be able to use the correct data type here instead of a generic `bool`:

```suggestion
		Headless:       args.Config.Value.NormalConfig.ProposeHeadless,
```

kevgo (approved):
Thanks for tackling this meaningful quality-of-life improvement for Git Town. We've seen a number of user reports around issues with automatically opening browsers, it's great to finally have a solid solution for this class of problems.

Also, great work on the implementation. The code is genuinely excellent. Reviewing it gave me a chance to think a bit about how we might generalize this feature across Git Town in the future. I left some fairly detailed comments along those lines. Nothing blocking for this PR, just brainstorming potential future refinements.

Apologies for the delayed review as well, I wanted to take a bit of time to think through naming here.

One concern with `headless` is that it already encodes a negation ("no head"), which can lead to awkward double negatives like:

```
headless = false
```

I'm wondering if we could make this more direct by naming the setting something like `browser`. For example, disabling all browser interactions would look like this as a CLI argument:

```sh
$ git town propose --no-browser
```

In the config file:

```toml
[forge]
browser = false
```

As Git metadata:

```sh
$ git config git-town.browser false
```

Curious what you think about that direction.

kevgo:
I’m going to go ahead and merge this. We can iterate naming and do other refactoring in follow-up PRs.

Big thanks @alexgeek, really appreciate the contribution 🙌

I've sent you an (optional) invite to the `git-town` GitHub org. If you accept, you'll be able to open PRs directly against this repo (no fork needed).

kevgo:
Hey @alexgeek could you just set the existing [browser](https://www.git-town.com/preferences/browser.html) preference to "(none)"? That should already prevent any browser windows from opening. Not sure what value the new `headless` setting is adding beyond that.

alexgeek:
> Hey @alexgeek could you just set the existing [browser](https://www.git-town.com/preferences/browser.html) preference to "(none)"? That should already prevent any browser windows from opening. Not sure what value the new `headless` setting is adding beyond that.

Hi, 

So I had been using git town with `glab` and `gh` and had not been using web connectors.
Because I was looking at that path I did not see the `browser` preference and both `gh` and `glab` had hard coded in `--web` with no way to turn it off. So it made sense to add a flag to conditionally swap out the `--web` for `--yes` with `glab` and just remove the `--web` for `gh`. 
I also found that I would want to switch between using `--headless` and not depending on if I was on a machine with a browser or not, using an agent, or if I simply wanted to create the proposal with more control, I think its a bit less ergonomic as persistent configuration personally.  

Using `--headless` going forward makes sense to me to drive intent / make an easier decision. If we get to a point where we would usually open a browser but `--headless` is enabled (whether for propose or other features like potentially `repo`) then I think at that point we simply check if we have enough supplied data and a way to use that data to make the action (API call, CLI tool call) - if not an error can be returned (missing data, no API available, not yet implemented etc.). 

In my experience across pipelining a lot of tools being able to explicitly state I want something to be headless is a good thing. I.e. say I have a pipeline running commands with `--headless` - I'd much rather it fail fast so I can fix the pipeline rather than potentially hanging or perhaps opening google chrome on every pipeline run. 

My preference would be to keep `--headless` and if it is set, make it rule that you can never wait for input from stdin/open a browser/on a UI. 

On the other hand I could perhaps see an argument to have "headless" as a browser option instead of its own specific flag, and perhaps could also use `--browser=headless` or `--browser=none` instead but that is bit looser.

internal/config/configdomain/propose_headless.go
alexgeek: Yes agree (expanded answer in comment at bottom)

kevgo:
Yeah, my thinking is that if the user runs `git town propose --no-browser` (or if the `browser` config is set to `(none)` or an empty string), then Git Town should never attempt to open a browser. In the future that would also mean omitting the `--web` flag when invoking `gh`.

That feels sufficient to fully suppress browser launches.

I like your idea of broadening the meaning of the "headless" mode to disable *all* interactivity: not just opening browser windows, but also any interactive prompts for user input. Right now the only way to get that behavior is by setting `TERM=dumb`, which is pretty implicit and not very discoverable.

A dedicated configuration option for this would be much cleaner, and it aligns well with increasing agentic/automation use cases.

kevgo:
I have created https://github.com/git-town/git-town/issues/6123 to track this idea.

kevgo:
After playing with the new `headless` flag in this PR, I realized it's really bundling two separate concerns:

* disabling interactive prompts/dialogs
* preventing browser windows from opening

I ended up modeling these independently via the new [interactive](https://www.git-town.com/preferences/interactive.html) mode and [browser](https://www.git-town.com/preferences/browser.html) setting. On top of that, Git Town now automatically disables both when running in a non-TTY environment.

With those pieces in place, a dedicated `headless` setting didn't seem to add much value anymore. It would mostly just be a convenience wrapper around two other settings. I don't think that tradeoff is worth the additional surface area and cognitive overhead for users, so I left it out of this release.

In general, I'd like Git Town settings to stay as orthogonal and composable as possible.

That said, thanks a lot for the inspiration here. The new `interactive` setting came directly out of exploring your `headless` proposal, and I hope it is as useful to you as the `headless` setting was.

alexgeek:
> After playing with the new `headless` flag in this PR, I realized it's really bundling two separate concerns:
> 
> * disabling interactive prompts/dialogs
> * preventing browser windows from opening
> 
> I ended up modeling these independently via the new [interactive](https://www.git-town.com/preferences/interactive.html) mode and [browser](https://www.git-town.com/preferences/browser.html) setting. On top of that, Git Town now automatically disables both when running in a non-TTY environment.
> 
> With those pieces in place, a dedicated `headless` setting didn't seem to add much value anymore. It would mostly just be a convenience wrapper around two other settings. I don't think that tradeoff is worth the additional surface area and cognitive overhead for users, so I left it out of this release.
> 
> In general, I'd like Git Town settings to stay as orthogonal and composable as possible.
> 
> That said, thanks a lot for the inspiration here. The new `interactive` setting came directly out of exploring your `headless` proposal, and I hope it is as useful to you as the `headless` setting was.

Makes sense to me! Will migrate over to your new modes. Cheers

Issue #6060, opened 2026-02-14
breadcrumb: the current proposal should no longer be a link


kevgo:
See https://github.com/git-town/action/issues/72.

- The current proposal is no longer a link, but the proposal title as text plus the pointing hand.
- Possibly make this text bold.

yaadata:
@kevgo 

I believe this change caused unintended consequences. Now the breadcrumb is replacing the current proposal with the branch name and not the proposal title. 

For the related issue, we never saw a visualization render issue with squiggly red underline and we still don't for cli breadcrumbs. I unfortunately had to re-pin back to 22.6.0 because exposing the branch name in the breadcrumb is rather awkward

yaadata:
Fix in #6107 

Could we get a 22.7.1 release for this change as well

kevgo:
I also don't see the red squiggly line, maybe this is a small roll-out of a new feature to a smaller subset of users.

kevgo:
> Could we get a 22.7.1 release for this change as well

@yaadata as you wish [v22.7.1](https://github.com/git-town/git-town/releases/tag/v22.7.1). Hope that unblocks you.

yaadata:
@kevgo thank you

PR #6107, opened 2026-03-24
fix: use proposal title for current breadcrumb branch

yaadata:
Addresses miss.

Mentioned here https://github.com/git-town/git-town/issues/6060#issuecomment-4119753751


internal/proposallineage/render_tree.go
```diff
@@ -32,12 +32,14 @@ func renderNodeDown(builder *strings.Builder, node TreeNodeWithProposal, current
...
-			builder.WriteString("**")
 		case hasProposal:
-			builder.WriteString(proposal.Data.Data().URL)
+			if isCurrentBranch {
+				builder.WriteString("**")
+				builder.WriteString(proposal.Data.Data().Title.String())
+				builder.WriteString("**")
+			} else {
+				builder.WriteString(proposal.Data.Data().URL)
+			}
```
yaadata: this is the change.

Some of the cucumber tests don't scaffold with a proposal hence why they go to the default block. It shouldn't be possible for a current branch of a proposal going to the default block

kevgo:
@yaadata great catch, this new version is much better, thanks so much!

kevgo:
This is now available as part of [v22.7.1](https://github.com/git-town/git-town/releases/tag/v22.7.1). Enjoy!

PR #6114, opened 2026-03-29
Move the existing configdomain.Browser setting to forgedomain

kevgo:


kevgo:
Let's move this to `browserdomain` instead.

PR #6133, opened 2026-04-09
chore: make sure mise.toml version consistent with go.mod

yaadata:
This file was introduced in
https://github.com/git-town/git-town/pull/6106/changes

Mise is a phenomenal tool management software. I don't know if it was
intentional to have it checked into main but if it was, we should keep the
versions consistent with go.mod


yaadata:
@kevgo I am going to merge this after checks pass because it breaks local development for people that use `mise` (such as myself).

If checking in this file was not intentional, we can remove it in a follow up PR

kevgo:
@yaadata thanks for cleaning this up and aligning the Go versions. Also 👍 on handling this via async review (merge first, review later). Feel free to do this more often for small and uncritical changes.

I was aware that this file was added as part of a contribution recently, and I'm generally fine with including config files for external tools it that makes things smoother for contributors.

Mise is an awesome tool. I didn’t want to make it a requirement for contributors, which is why I put together an alternative with [run-that-app](https://github.com/kevgo/run-that-app) that is completely non-intrusive. That said, I'm totally on board with supporting mise as an option for those who prefer it.

PR #6146, opened 2026-04-12
docs: DEVELOPMENT.md should mention go v1.26 not v1.23

yaadata:


yaadata:
Merging...

We should also limit CI end-to-end checks to only code changes and exclude doc ONLY changes

kevgo:
Thanks for fixing this! :pray: 

> We should also limit CI end-to-end checks to only code changes and exclude doc ONLY changes

Can you please give an example for a test that you would like to disable on CI, and provide a bit more background around why you want to disable it?

yaadata:
This is to remove unnecessary CI steps for a PR.

If I'm only updating documentation (for example), do the end to end tests really need to run? My firm stance is no.

GitHub Actions has path/extension filters to prevent certain ci pipelines from executing https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-excluding-paths

kevgo:
I agree that there is potential for optimizations.

Historically, when the end-to-end tests started slowing down CI, our approach was to improve their performance rather than selectively skipping them for certain types of changes. This benefits all PRs, not just doc-only changes. Doc-only PRs are pretty rare for Git Town, so making E2E tests run faster had more and better impact.

Even doc-only changes need to get validated on CI. Right now, those checks run as part of the [linter job](https://github.com/git-town/git-town/blob/2c3b1d55146588c07cd22289c6a13979770a0b70/.github/workflows/lint.yml#L19), which takes about 3 minutes. Previously, our E2E tests ran faster than that, so skipping them didn't really accelerate doc-only PRs.

Recently we added a ton of new E2E tests, and the E2E CI job now takes around 5 minutes. That's still quite reasonable, but I agree we're approaching a point where E2E tests are consistently the slowest part of CI.

PR #6148, opened 2026-04-14
fix: use fallback breadcrumbs for gitea and forgejo

yaadata:
## Summary
Use the fallback breadcrumb render strategy for Gitea and Forgejo so proposal breadcrumbs show explicit markdown links with titles.

## Why
Gitea and Forgejo linked-reference docs describe `#1234` and `!1234` as linked references, but they do not document GitHub/GitLab-style inline title rendering in proposal descriptions. 

From manual testing, the title rendering does not show from reference links like on Github/Gitlab.

Using the fallback renderer to address this gap.

## References
- https://docs.gitea.com/1.24/usage/automatically-linked-references
- https://forgejo.org/docs/v15.0/user/linked-references/


yaadata:
@kevgo this change is a correction to my prior PR after I went, signed up on Gitea/Forgejo and actually tested this.



Because checks pass and this is 6 lines of user facing code being changed, going to merge this as is

kevgo:
Thanks for going the extra mile and verifying and fixing this! :raised_hands:

PR #6149, opened 2026-04-16
fix(sync): use actual stash size instead of hardcoded 0 in BeginStashSize

endigma:
## Summary
- The sync command hardcoded `BeginStashSize: 0` in the RunState, while every other command correctly used `data.stashSize`. This caused `git-town undo` after a failed sync to incorrectly pop pre-existing stash entries that Git Town did not create.
- Fixed by using `data.stashSize` like all other commands.
- Added an end-to-end test covering the scenario: pre-existing stash, no open changes, sync conflict, then undo — verifying the stash is left untouched.
- New test `conflict_with_pre_existing_stash.feature` covers result, undo, and continue scenarios
- Existing `add_stash.feature` still passes

### DIFF
--- a/cucumber-sort.json
+++ b/cucumber-sort.json
@@ -166,5 +166,8 @@
     "^the initial tags exist now$",
     "^the initial proposals exist now$",
     "^the proposals are now$"
+  ],
+  "unknown-steps": [
+    "^I add an unrelated stash entry with file \".*\"$"
   ]
 }
--- a/features/sync/current_branch/feature_branch/merge_sync_strategy/tracking_branch/with/conflict_with_pre_existing_stash.feature
+++ b/features/sync/current_branch/feature_branch/merge_sync_strategy/tracking_branch/with/conflict_with_pre_existing_stash.feature
@@ -1,0 +1,47 @@
+Feature: undo sync conflict does not pop pre-existing stash
+
+  Background:
+    Given a Git repo with origin
+    And the branches
+      | NAME    | TYPE    | PARENT | LOCATIONS     |
+      | feature | feature | main   | local, origin |
+    And the commits
+      | BRANCH  | LOCATION | MESSAGE                   | FILE NAME        | FILE CONTENT   |
+      | feature | local    | conflicting local commit  | conflicting_file | local content  |
+      |         | origin   | conflicting origin commit | conflicting_file | origin content |
+    And the current branch is "feature"
+    And I add an unrelated stash entry with file "stashed_file"
+    When I run "git-town sync"
+
+  Scenario: result
+    Then Git Town runs the commands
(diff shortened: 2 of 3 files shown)


endigma:
Also @kevgo, please consider not putting broad automatic execution AI configuration in your repository. It may work locally, but stuff like hooks and huge allowlists in repo-tracked `.claude/settings.json` and other files can break all kinds of environments. 

For example, both of the hooks you have fail when $TERM is unset, e.g. when running Claude Code via the Agent SDK.

I would suggest making this `settings.local.example.json` or etc, so contributors wishing to use your settings can do so but are not forced to without dirty git state or local overrides.

kevgo:
Thanks for fixing this bug! :pray:

kevgo:
> Also @kevgo, please consider not putting broad automatic execution AI configuration in your repository. It may work locally, but stuff like hooks and huge allowlists in repo-tracked `.claude/settings.json` and other files can break all kinds of environments.
> 
> For example, both of the hooks you have fail when $TERM is unset, e.g. when running Claude Code via the Agent SDK.
> 
> I would suggest making this `settings.local.example.json` or etc, so contributors wishing to use your settings can do so but are not forced to without dirty git state or local overrides.

Thanks for letting me know about this. Happy to adjust the agent configuration, it's completely experimental at this point. Following your suggestion, https://github.com/git-town/git-town/pull/6151 renames `.claude/settings.json` to `.claude/settings.local.example.json`.

I'm a bit surprised that the make tasks don't run inside agentic environments. They just run various CLI apps. Thanks for notifying me about that as well.

kevgo:
I've sent you an invite to the `git-town` GitHub org. It's totally optional. If you accept, you'll be able to make further contributions without the need for a fork, run CI without needing approval, and you'll get a Git Town badge on your GitHub profile.

endigma:
Hi @kevgo , do you know approximately when this might be cut in a release? Looking forward to not losing my stashes :)

kevgo:
Normally we'd ship this immediately, but the next release is a major version bump. That means we're batching all breaking changes together before cutting the release. It's a rare timing window, but you happened to hit it. Expect roughly another 1–2 weeks.

If you want to use your fix sooner, you can build Git Town locally. In the repo folder:

```bash
make install
```

The Git Town binary will be available at:

```
~/go/bin/git-town
```

Does this work for you?

endigma:
Curious what makes the next release a major version bump, already something breaking on main?

kevgo:
Nothing is really breaking (Git Town takes stability serious) but there are a few adjustments worth pointing attention to via a major version number bump. In Git Town 23, the `git town up` and `git town down` commands get swapped, the `git town walk` command will stop if a command produces uncommitted changes, and Git Town will no longer open a browser when running in a non-interactive shell.

PR #6170, opened 2026-04-26
[PART 1] enhancement: breadcrumb exclusion logic when rendering

yaadata:
## Summary

This teaches proposal breadcrumb rendering how to remove configured branch types from the lineage before writing the breadcrumb section.

The rendering path now:

- calculates the normal branch lineage tree with branch-type data attached to each `TreeNode`
- filters out branches whose types are configured in `proposal-breadcrumb-exclude-branches`
- preserves visible descendants when an excluded branch sits in the middle of the lineage
- renders the resulting `TreeNodes` into the proposal body from both `git town propose` and the proposal breadcrumb update opcode

### Why `TreeNodes`

Filtering a tree can produce more than one visible root. For example, when the configured exclusions remove the root branch, its visible children still need to be rendered instead of discarded. `TreeNodes` follows the codebase's collection-type naming convention while still representing that multi-root result.

### Implementation

- Adds `TreeNodes` and `TreeNodesWithProposal` as first-class proposal-lineage collection types.
- Adds `BranchType` to `TreeNode` so filtering has all required data in the tree itself.
- Adds `FilterTree` to remove excluded branch types while promoting kept descendants.
- Moves proposal enrichment and rendering onto the collection types: `TreeNodes.AddProposals` and `TreeNodesWithProposal.Render`.
- Extends `RenderSectionArgs` with branch type data and excluded branch types.
- Passes configured exclusions through the propose command and proposal breadcrumb update opcode.

<details>
<summary>Sequence diagram</summary>

```mermaid
sequenceDiagram
    participant Command as propose/update opcode
    participant RenderSection
    participant Tree as CalculateTree
    participant Filter as FilterTree
    participant Proposals as TreeNodes.AddProposals
    participant Render as TreeNodesWithProposal.Render

    Command->>RenderSection: config, lineage, branch types, excluded types
    RenderSection->>Tree: CalculateTree(current branch)
    Tree-->>RenderSection: TreeNode with branch types
    RenderSection->>Filter: FilterTree(tree, excluded)
    Filter-->>RenderSection: TreeNodes
    RenderSection->>Proposals: AddProposals(connector)
    Proposals-->>RenderSection: TreeNodesWithProposal
    RenderSection->>Render: Render(current branch, direction, connector)
    Render-->>Command: Markdown breadcrumb section
```

</details>

-------------------------
- main
  - **[PART 1] enhancement: breadcrumb exclusion logic when rendering** :point_left:
    - #6173
      - #6192

<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>


internal/cmd/propose.go
```diff
@@ -400,7 +400,16 @@ func proposeProgram(repo execute.OpenRepoResult, data proposeData) program.Progr
...
+			lineageSection := proposallineage.RenderSection(proposallineage.RenderSectionArgs{
+				BranchTypes:   make(configdomain.BranchesAndTypes),
+				Breadcrumb:    data.config.NormalConfig.ProposalBreadcrumb,
+				Connector:     data.connector,
+				CurrentBranch: branchToPropose.name,
+				Direction:     data.config.NormalConfig.ProposalBreadcrumbDirection,
+				Excluded:      set.New[configdomain.BranchType](),
+				Lineage:       data.config.NormalConfig.Lineage,
+				Order:         data.config.NormalConfig.Order,
+			})
```
kevgo: Thanks for doing this long-overdue refactor, this function had indeed way too many arguments!

Since you already have a branch stack in place, one thing you might consider is splitting the positional -> named argument migration into a separate prepended branch. That change is independent from the rest of the work here, so it could get merged much faster on its own. This is actually a good example for a small and non-debatable cleanup that you can easily merge on your own and then I'll review it asynchronously.

Pulling this refactor out also keeps the remaining changes in this PR smaller and more focused, which makes it easier to review and reason about them.

internal/proposallineage/forest.go
```diff
@@ -0,0 +1,11 @@
+package proposallineage
+
+type Forest []TreeNode
```
kevgo: The naming convention in this codebase for such collection types is:

```suggestion
type TreeNodes []TreeNode
```

internal/proposallineage/render_test.go
```diff
@@ -109,13 +148,171 @@ func TestRender(t *testing.T) {
...
+		want := `
+
+-------------------------
+- **feature-a proposal title** :point_left:
+  - [prototype-a proposal title](https://www.github.com/git-town/git-town/pull/2)
+    - [feature-b proposal title](https://www.github.com/git-town/git-town/pull/3)
+
+<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>
+`[1:]
+		must.Eq(t, want, have)
```
kevgo: You compare two strings here, this can be done a bit more efficiently.

```suggestion
		must.EqOp(t, want, have)
```

internal/proposallineage/tree.go
```diff
@@ -3,11 +3,13 @@ package proposallineage
...
 	"github.com/git-town/git-town/v23/internal/git/gitdomain"
+	. "github.com/git-town/git-town/v23/pkg/prelude"
 )
 
 type TreeNode struct {
-	Branch   gitdomain.LocalBranchName
-	Children []TreeNode
+	Branch        gitdomain.LocalBranchName
+	Children      Forest
+	LineageParent Option[gitdomain.LocalBranchName]
```
kevgo, line 13: If you add the branch type here, then the tree contains all the information needed to filter it. This makes it easier to test and debug the logic that performs the filtering. 

```suggestion
	LineageParent Option[gitdomain.LocalBranchName]
	BranchType    configdomain.BranchType
```

internal/proposallineage/tree_with_proposals.go
```diff
@@ -7,9 +7,10 @@ import (
 )
 
 type TreeNodeWithProposal struct {
-	Branch   gitdomain.LocalBranchName
-	Children []TreeNodeWithProposal
-	Proposal Option[forgedomain.Proposal]
+	Branch        gitdomain.LocalBranchName
+	Children      ForestWithProposals
```
kevgo: Before, the field `Children` had the type `[]TreeNodeWithProposal`. Now it has the type `ForestWithProposals`, which is a newtype around `[]TreeNodeWithProposal` with currently no methods. In the current state, the newtype doesn't add enough value, and using the old slice type (`[]TreeNodeWithProposal`) would be simpler and more explicit.

If you end up adding methods to the new type, then there would be a reason to keep it.

internal/proposallineage/forest_with_proposals.go
```diff
@@ -0,0 +1,17 @@
+package proposallineage
+
+import (
+	"github.com/git-town/git-town/v23/internal/forge/forgedomain"
+	. "github.com/git-town/git-town/v23/pkg/prelude"
+)
+
+type ForestWithProposals = []TreeNodeWithProposal
```
kevgo:
```suggestion
type TreeNodesWithProposal = []TreeNodeWithProposal
```

internal/proposallineage/forest_with_proposals.go
```diff
@@ -0,0 +1,17 @@
...
+
+func AddProposalsToForest(forest Forest, connector Option[forgedomain.Connector]) ForestWithProposals {
```
kevgo: Make this a method of `Forest` (or once renamed `TreeNodes`)?

internal/proposallineage/render_forest.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"strings"
+
+	"github.com/git-town/git-town/v23/internal/config/configdomain"
+	"github.com/git-town/git-town/v23/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v23/internal/git/gitdomain"
+	. "github.com/git-town/git-town/v23/pkg/prelude"
+)
+
+func RenderForest(forest ForestWithProposals, currentBranch gitdomain.LocalBranchName, direction configdomain.ProposalBreadcrumbDirection, connector Option[forgedomain.Connector]) string {
```
kevgo: Make this a method on `ForestWithProposals` (or once renamed `TreeNodesWithProposals`)? That seems to be the main data structure for this function, i.e. the thing that gets rendered. The other arguments just fine-tune how the forest gets rendered.

internal/proposallineage/render_section.go
```diff
@@ -5,20 +5,43 @@ import (
...
+}
+
 // RenderSection provides the branch lineage for the given branch in Markdown format, ready to be embedded into a proposal body.
-func RenderSection(lineage configdomain.Lineage, currentBranch gitdomain.LocalBranchName, order configdomain.Order, breadcrumb configdomain.ProposalBreadcrumb, direction configdomain.ProposalBreadcrumbDirection, connector Option[forgedomain.Connector]) string {
+func RenderSection(args RenderSectionArgs) string {
 	// step 1: calculate the lineage tree for the given branch
-	tree := CalculateTree(currentBranch, lineage, order)
+	tree := CalculateTree(args.CurrentBranch, args.Lineage, args.Order)
+
+	forest := FilterTree(
```
kevgo: This introduces a new step in the workflow, so let’s assign it a step number to be consistent with the other steps here. Can you please also update the numbering below accordingly?

```suggestion
  // step 2: filter out the excluded branch types 
	forest := FilterTree(
```

kevgo (commented):
Thanks for implementing this! I really like the overall direction: enrich the tree with more data and then filter out some of the nodes. That's the right approach.

The existing implementation is already in pretty good shape. I have nothing to add, only a things to take away :slightly_smiling_face:  Below are a few suggestions mostly for simplifying it and make this a simple solutions for a complex problem (which is one of the hardest parts of programming). Curious what you think!

yaadata:
@kevgo haha I was wondering if you were going to be okay with the Forest naming. Went more classical data structure naming but yes I will update

internal/cmd/propose.go
yaadata: https://github.com/git-town/git-town/pull/6210

yaadata:
Assuming all is 🟢 I believe this PR is ready to land

internal/cmd/propose.go
```diff
@@ -401,10 +401,12 @@ func proposeProgram(repo execute.OpenRepoResult, data proposeData) program.Progr
 		proposalBody := data.proposalBody
 		if updateBreadcrumb {
 			lineageSection := proposallineage.RenderSection(proposallineage.RenderSectionArgs{
+				BranchTypes:   make(configdomain.BranchesAndTypes),
 				Breadcrumb:    data.config.NormalConfig.ProposalBreadcrumb,
 				Connector:     data.connector,
 				CurrentBranch: branchToPropose.name,
 				Direction:     data.config.NormalConfig.ProposalBreadcrumbDirection,
+				Excluded:      set.New[configdomain.BranchType](),
```
kevgo, line 409: When do `BranchTypes` and `Excluded` get populated?

internal/proposallineage/render_section.go
```diff
@@ -5,29 +5,43 @@ import (
...
 	}
 
-	// step 2: add proposals to the tree
-	treeWithProposals := AddProposalsToTree(tree, args.Connector)
+	// step 3: add proposals to the tree
+	treesWithProposals := treeNodes.AddProposals(args.Connector)
 
-	// step 3: render the tree into Markdown format
-	return RenderTree(treeWithProposals, args.CurrentBranch, args.Direction, args.Connector)
+	// step 4: render the tree[s] into Markdown format
```
kevgo, line 45: There is always going to be only one tree. What is different is that it is filtered now.

```suggestion
	// step 4: render the tree into Markdown format
```

internal/proposallineage/tree.go
```diff
@@ -20,25 +23,35 @@ func (self TreeNode) BranchCount() int {
...
-			childNode := buildTree(child, lineage, includeBranches, order)
+			childNode := buildTree(child, Some(branch), lineage, includeBranches, order, branchTypes)
 			children = append(children, childNode)
 		}
 	}
 	return TreeNode{
-		Branch:   branch,
-		Children: children,
+		Branch:        branch,
+		BranchType:    branchTypeFor(branch, branchTypes),
```
kevgo, line 45: You could use the existing `BranchesAndTypes.Get()` method here. The suggested change has the same behavior as your `branchTypeFor` function. But it makes it explicit that in this case you default to an empty string if the branch type isn't known. If you go this way, then you can remove the `branchTypeFor` function.

```suggestion
		BranchType:    branchTypes.Get(branch).GetOr(""),
```

internal/proposallineage/tree_filter.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"github.com/git-town/git-town/v23/internal/config/configdomain"
+	"github.com/git-town/git-town/v23/pkg/set"
+)
+
+// FilterTree removes branches whose type is excluded.
+// When it removes a branch in the middle of the tree,
+// it keeps the visible descendants and moves them up.
+// This can turn one tree into multiple visible roots,
+// so the result is a forest data structure.
```
kevgo, line 12: I see now where the forest comes from. Thanks for the explanation!

internal/proposallineage/tree_filter.go
```diff
@@ -0,0 +1,39 @@
...
+func FilterTree(
+	tree TreeNode,
+	excluded set.Set[configdomain.BranchType],
+) TreeNodes {
+	return filterTreeNode(tree, excluded)
+}
```
kevgo, line 18: This function FilterTree doesn't do anything, it just calls filterTreeNode with the exact same arguments and returns its value. You could delete FilterTree and then rename filterTreeNode to FilterTree.

What do you think about making this function a method on `TreeNode`?

internal/proposallineage/tree_filter_test.go
```diff
@@ -0,0 +1,241 @@
...
+					proposallineage.TreeNode{
+						Branch:        "feature",
+						LineageParent: Some(gitdomain.LocalBranchName("main")),
+						Children:      proposallineage.TreeNodes{},
+					},
+				},
+			},
+		}
+
+		must.Eq(t, treeNodesWithBranchTypes(want, branchTypes), have)
```
kevgo, line 60: I would be okay with verifying the actual `have` variable here, not a version augmented with `treeNodesWithBranchTypes`. The purpose of these tests is also to document how a typical result of `proposallineage.FilterTree` looks like.

```suggestion
		must.Eq(t, want, have)
```

internal/proposallineage/tree_filter_test.go
```diff
@@ -0,0 +1,241 @@
...
+)
+
+func TestFilterTree(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Exclude leaf node", func(t *testing.T) {
+		t.Parallel()
+
+		tree := proposallineage.TreeNode{
+			Branch:        "main",
```
kevgo, line 21: I think this test should also define the `BranchType` for each `TreeNode`.


```suggestion
			Branch:        "main",
			BranchType:    configdomain.BranchTypeMainBranch,
```

internal/proposallineage/tree_filter_test.go
```diff
@@ -0,0 +1,241 @@
...
+			},
+		}
+		branchTypes := configdomain.BranchesAndTypes{
+			"main":      configdomain.BranchTypeMainBranch,
+			"feature":   configdomain.BranchTypeFeatureBranch,
+			"prototype": configdomain.BranchTypePrototypeBranch,
+		}
+		excluded := set.New(configdomain.BranchTypePrototypeBranch)
+
+		have := proposallineage.FilterTree(treeWithBranchTypes(tree, branchTypes), excluded)
```
kevgo, line 44: Could we provide the input data verbatim here instead of using a helper like `treeWithBranchTypes`? Please add the branch types directly to the declaration of the `tree` variable above.

I realize this is somewhat redundant, but in tests, explicit is better than implicit.

internal/proposallineage/tree_nodes.go
```diff
@@ -0,0 +1,25 @@
...
+	result := make(TreeNodesWithProposal, 0, len(self))
+
+	for _, tree := range self {
+		result = append(result, AddProposalsToTree(tree, connector))
+	}
+	return result
+}
+
+func (self TreeNodes) BranchCount() int {
+	var count int
```
kevgo, line 20: It's better to make the initial value explicit, even though the zero value would also be correct here.

```suggestion
	count := 0
```

internal/vm/opcodes/proposal_update_breadcrumb.go
```diff
@@ -36,10 +38,12 @@ func (self *ProposalUpdateBreadcrumb) Run(args shared.RunArgs) error {
 	for _, proposal := range proposals {
 		oldProposalBody := proposal.Data.Data().Body.GetOrZero()
 		lineageSection := proposallineage.RenderSection(proposallineage.RenderSectionArgs{
+			BranchTypes:   make(configdomain.BranchesAndTypes),
 			Breadcrumb:    args.Config.Value.NormalConfig.ProposalBreadcrumb,
 			Connector:     args.Connector,
 			CurrentBranch: self.Branch,
 			Direction:     args.Config.Value.NormalConfig.ProposalBreadcrumbDirection,
+			Excluded:      set.New[configdomain.BranchType](),
```
kevgo, line 46: When do the values `BranchTypes` and `Excluded` get populated? I think we should either populate the correct values here, or remove this arguments because they don't carry any information.

kevgo (commented):
Nice implementation! A few ideas to tie down a few loose ends and then this is production-grade. Thanks for building this!

internal/cmd/propose.go
yaadata: In the next PR in the stack

internal/proposallineage/render_section.go
yaadata: Could you take a peak at the PR description and there is related test showcasing why we now call these trees

internal/vm/opcodes/proposal_update_breadcrumb.go
yaadata: Next PR in the stack

internal/proposallineage/tree_nodes.go
yaadata: Should we have a lint for this?

internal/proposallineage/render_section.go
kevgo: Right, I saw that later too. Forgot to remove this comment :smile:

internal/proposallineage/tree_nodes.go
kevgo: Yes, good idea. I have created https://github.com/git-town/git-town/issues/6215 for this.

internal/proposallineage/tree.go
kevgo: Unaddressed comment

internal/proposallineage/tree_filter.go
kevgo: Unaddressed comment

internal/proposallineage/tree_filter_test.go
kevgo: unaddressed comment

internal/proposallineage/tree_filter_test.go
kevgo: unaddressed comment

internal/proposallineage/tree_filter_test.go
kevgo: unaddressed comment

internal/proposallineage/tree_nodes.go
kevgo: I've explored building a linter for this, but there are too many exceptions to handle reliably.

For now, please apply the suggested change here. Thanks!

kevgo (commented):
Hey @yaadata, just checking in on a few comments and suggestions that are still unresolved in this PR. No rush if you're busy. The PR is looking great, and I'd like to merge it to ship this feature.

There are a few merge conflicts now, mostly due to an additional parameter being passed into the rendering logic. They should be fairly straightforward to resolve.

Please let me know how you want to move forward with this. Thanks!

yaadata:
@kevgo there is a full stack of diff(3 different PRs that all need reviews) for this change but I currently don't have bandwidth to land this over the next few weeks, probably until July.

I can't commit to definite timeline so for now I will close these changes and re-open at a later date when I am ready for reviews. If you'd like, you can copy these PRs if the idea is to land them sooner.

PR #6189, opened 2026-05-05
Stop annoying assistant as soon as possible. Allow of user to exit with minimal setup.

txgruppi:
Add a "Complete Minimal Setup" dialog step so users can end setup as soon as the required configuration is entered. Propagate minimal-mode through enter/save flows to persist only essential settings and stop the annoying assistant as early as possible.

### DIFF
--- a/internal/cli/dialog/complete_minimal.go
+++ b/internal/cli/dialog/complete_minimal.go
@@ -1,0 +1,34 @@
+package dialog
+
+import (
+	"github.com/git-town/git-town/v22/internal/cli/dialog/dialogcomponents"
+	"github.com/git-town/git-town/v22/internal/cli/dialog/dialogcomponents/list"
+	"github.com/git-town/git-town/v22/internal/cli/dialog/dialogdomain"
+	"github.com/git-town/git-town/v22/internal/config/configdomain"
+)
+
+const (
+	completeMinimalTitle = `Complete Minimal Setup`
+	completeMinimalHelp  = `
+Complete the minimal setup of Git Town,
+and could end the assistant here
+or continue with cli and forge configuration.
+
+You can re-run this assistant: git town init
+`
+)
+
+func CompleteMinimal(inputs dialogcomponents.Inputs, interactive configdomain.Interactive) (bool, dialogdomain.Exit, error) {
+	entries := list.Entries[bool]{
+		{
+			Data: true,
+			Text: `exit and save`,
+		},
+		{
+			Data: false,
+			Text: `continue to the cli and forge configuration`,
+		},
+	}
+	selection, exit, err := dialogcomponents.RadioList(entries, 0, completeMinimalTitle, completeMinimalHelp, inputs, interactive, "complete-minimal")
+	return selection, exit, err
+}
--- a/internal/cmd/init.go
+++ b/internal/cmd/init.go
@@ -94,11 +94,11 @@ Start:
 		goto Start
 	}
 	data.Config.NormalConfig.Interactive = interactive
-	userInput, exit, enterAll, err := setup.Enter(data, repo.ConfigDir)
+	userInput, exit, minimal, enterAll, err := setup.Enter(data, repo.ConfigDir)
 	if err != nil || exit {
 		return err
 	}
-	if err = setup.Save(userInput, repo.UnvalidatedConfig, data, enterAll, repo.Frontend); err != nil {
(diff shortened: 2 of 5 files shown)


kevgo:
Thanks for implementing this! The setup assistant is meant to help users get started with a meaningful basic configuration. I agree that it can be frustrating when new users are forced to deal with so many questions upfront.

I like your idea to give the users an additional option to exit with the absolute bare minimum configuration, so that `hack` and `sync` work.

I'll take a closer look at this in the next couple of days.

kevgo:
@txgruppi The setup assistant already supports the workflow this PR is aiming for: it exits as soon as the user has provided the minimum configuration Git Town needs to work well for a typical user. Right now that minimal setup is:

1. shortcuts for Git Town commands
2. the main branch
3. perennial branches
4. access to the forge API

We already agree that step 2 is essential.

I'd also consider step 3 essential. If perennial branches are configured incorrectly, Git Town will sync them the wrong way, which can create pretty serious issues.

Without step 4, Git Town is significantly less useful. If we remove it from the minimal setup, I think a large majority of users would end up re-running the setup assistant within a few hours once they hit features that depend on forge access. That feels like a worse experience overall.

Step 1 is the most debatable one. I currently keep it in the minimal setup because Git Town users typically run a lot of Git Town commands every day. Being able to run `git sync` instead of `git town sync` saves enough friction over time that I think it's worth asking once during setup. Having said that, I'm open to arguments for taking it out of the minimal setup.

What your PR introduces is effectively two different "minimal" setups. To me that adds complexity and makes the onboarding flow harder to reason about. I'd prefer to keep the distinction simple: one minimal setup and one complete setup.

If you use Git Town on a lot of repositories, where entering data into the setup assistant becomes repetitive and annoying, you can provide the minimal setup yourself so that Git Town doesn't show the setup assistant at all:

- Manually copy the [Git Town configuration file](https://www.git-town.com/configuration-file.html) from another repo that is already configured into the current repo. If the config file provides all the needed information, Git Town won't show the setup assistant. That's what I do when I set up Git Town on new repos.

- If creating a Git Town config file in your repos is not in the cards, you could configure the [main branch](https://www.git-town.com/preferences/main-branch.html#git-metadata) and the [perennial branches](https://www.git-town.com/preferences/perennial-branches.html#git-metadata) through Git metadata. Then Git Town won't show the setup assistant.

```sh
git config git-town.main-branch <value>
git config git-town.perennial-branches ""  # empty stirng configures no perennial branches
```

If you provide the `--global` flag to the two shell commands above, it automatically applies the setting to all repos on your machine.

Would any of these workarounds solve the problem you're trying to address?

txgruppi:
@kevgo I understand.

Feel free to close the PR if it doesn't fit your view of the future of git-town.

My intention was to have git-town working as fast as possible for my use case, and maybe for someone that uses it like me.

1. All my repos have only `master` and feature branches.
2. All commands I use are `sync`, `hack`, and `delete`.

I don't have use for any other commands, specially forge related.

I imagine I may be one of very few people that uses git-town like this.

txgruppi:
@kevgo I'm reading git-town's docs again to see which features I'm missing that could help me more with my work. Thanks.

kevgo:
Thanks for the replies. How do you merge finished feature branches into your master branch? If you merge them locally, then [ship](https://www.git-town.com/commands/ship.html) might be interesting to you. If you use a forge, then [propose](https://www.git-town.com/commands/propose.html) is better.

I also think you might find [switch](https://www.git-town.com/commands/switch.html) useful for switching to another branch.

Based on what you said, I'm going to close this PR since the intended functionality already exists. Please reach out if you have any other questions!

txgruppi:
@kevgo, everything after a push is done via the web ui, as it is easier for everyone in the team to work together. `git push` already displays the link we need to create the MR (Gitlab), it is just too easy to use already, I can't see how ship or propose can help me.

Switch is "fun", because of how it presents branches, but doesn't offer me anything I don't get with `git checkout`, as far as I could see. All my branch names follow well defined patterns, so it is really easy to switch to any of them when needed with the basic bash completions.

I use `hack`, `sync` and `delete` because they do the work of several commands in one go, and they know how to "undo" stuff in case it is needed. The other commands feel like "sweeter" versions of existing git commands, not something I really care about.

Well, this is just my view about the tool and its usage.

Thank you.

kevgo:
Thanks for sharing how you see it, this is helpful. How you use Git Town's core feature set makes sense for your use case. The other Git Town commands make more sense when using [stacked changes](https://www.git-town.com/stacked-changes.html).

PR #6203, opened 2026-05-17
chore: ensure_no_files_with_dashes.sh should not consider gitignored files

yaadata:


yaadata:
@kevgo going to merge this in the morning,

This is helpful for workflows where I have a local only gitignore that ignores certain files. The linter keeps failing on these ignored files on file with dashes, which it shouldn't because there is no intention to check in these files

kevgo (commented):
Thanks for fixing this! 🚀

PR #6212, opened 2026-05-20
fix(worktrees): don't treat bare-repo HEAD branch as active in another worktree

### LINKED ISSUE #6211: [worktrees] `main` treated as checked out by bare repo when in a worktree
### Steps to reproduce

Hello! I'm creating worktrees from bare repos locally to avoid the "oops `main` is checked out somewhere, can't sync" issue, but it seems Git Town is treating `main` as checked out by the bare worktree anyway -- even though it's possible to check it out and update it. 

I can work around this by checking out `main` in a worktree and manually pulling, but it'd be nice to have the Git Town worktree detection make this distinction. 

Similarly the `git town switch` UI (as well as the `set-parent` UI? perhaps wrongly, there) gray out `main` and don't allow me to pick it.

I'd be happy to put a PR up if you're open to it!

### Command output

```shell
from a line in `git town sync --verbose`: 

refname:refs/heads/main branchname:main sha:[REDACTED] head:N worktree:Y symref:N upstream:origin/main track:

`git worktree list` redacted output:

/Users/ben/[redacted]/worktrees/_bares/[a repo].git  (bare)
/Users/ben/[redacted]/worktrees/[task name]/[a repo]  {short sha} [bmo/{redacted}]
```

### Additional information

Main is disabled:
Image

and yet, `git town down` does switch to `main` correctly 👍 but `git town sync` will not update `main` first.

Sorry for the limited+redacted output but I'm using this at work 😅

benmosher:
## Summary

- When working from a linked worktree of a bare repository, `git for-each-ref`'s `%(worktreepath)` returns the bare repo path for whichever branch HEAD points to (typically `main`). Git Town was interpreting this as `SyncStatusOtherWorktree`, refusing to sync that branch and greying it out in `git town switch`.
- Fix: replace the Y/N worktree field in the `for-each-ref` format with the actual `%(worktreepath)` value, using NUL bytes (`%00`) as field delimiters to safely handle paths and track values that contain spaces.
- Add `isBareWorktreePath` helper that identifies bare repos by the absence of a `.git` entry inside the worktree path (with a fast-path check for the conventional `.git` suffix). Branches whose worktreepath is bare are marked available rather than locked.

Closes #6211

## Test plan

- [x] New unit test: `BranchesAvailableInCurrentWorktree` — "includes branch that is HEAD of a bare worktree"
- [x] New unit test: `BranchesSnapshot` — "branch is HEAD of a bare worktree" (verifies no `SyncStatusOtherWorktree`)
- [x] Existing test "excludes branch checked out in another worktree" still passes (non-bare linked worktrees remain correctly blocked)
- [x] Manual: `git town sync` from a linked worktree no longer skips `main`; `git town switch` shows `main` as selectable

🤖 Generated with [Claude Code](https://claude.com/claude-code)


benmosher:
please _do_ be harsh in your criticism -- I am not a Go dev so this is Claude work and I am personally blind to any faux pas it may have committed 😅

internal/git/commands.go
```diff
@@ -957,12 +960,15 @@ func branchesQuery(querier subshelldomain.Querier) (branchesQueryResults, error)
 	lines := stringslice.Lines(output.String())
 	result := make(branchesQueryResults, len(lines))
 	for l, line := range lines {
-		parts := strings.SplitN(line, " ", len(forEachRefFormats))
```
benmosher, line 960: I think `SplitN` is no longer needed given the change to null byte delimiter; could change to a raw `strings.Split` (assuming that exists and is the correct unbounded split function)

but: indistinguishable, practically speaking

internal/git/commands.go
```diff
@@ -980,6 +986,17 @@ func branchesQuery(querier subshelldomain.Querier) (branchesQueryResults, error)
 	return result, nil
 }
 
+// isBareWorktreePath reports whether the given worktree path is a bare repository.
+// Non-bare linked worktrees always contain a .git file; bare repos do not.
+// Conventionally bare repos are named with a ".git" suffix, checked first as a fast path.
+func isBareWorktreePath(path string) bool {
+	if strings.HasSuffix(path, ".git") {
+		return true
+	}
```
kevgo: It's totally reasonable that the LLM brings up that the path of bare repos usually end with `.git`, and proposes a fast path based on this. I love how LLMs excel at finding such patterns!

That said, it's pretty trivial to create a non-bare worktree that also ends in `.git`:

```sh
git clone git@github.com:git-town/git-town.git
cd git-town
git worktree add ../git-town-2.git
```

Git Town needs to behave correctly even in such unusual circumstances. The fastpath would result in buggy behavior in this situation: confusing a non-bare repo for a bare one. Let's therefore take the fast path out.

Doing so won't be a performance problem. Each Git Town command already performs hundreds of file operations (by itself and through Git). Adding one more file metadata check here won't be noticeable.

```suggestion
```

internal/git/commands.go
```diff
@@ -980,6 +986,17 @@ func branchesQuery(querier subshelldomain.Querier) (branchesQueryResults, error)
 	return result, nil
 }
 
+// isBareWorktreePath reports whether the given worktree path is a bare repository.
+// Non-bare linked worktrees always contain a .git file; bare repos do not.
+// Conventionally bare repos are named with a ".git" suffix, checked first as a fast path.
+func isBareWorktreePath(path string) bool {
```
kevgo: Suggestion to simplify the name:

```suggestion
func isBareWorktree(path string) bool {
```

internal/git/commands.go
```diff
@@ -980,6 +986,17 @@ func branchesQuery(querier subshelldomain.Querier) (branchesQueryResults, error)
 	return result, nil
 }
 
+// isBareWorktreePath reports whether the given worktree path is a bare repository.
+// Non-bare linked worktrees always contain a .git file; bare repos do not.
+// Conventionally bare repos are named with a ".git" suffix, checked first as a fast path.
```
kevgo: The function docstring should only summarize what the function does for human users, ideally in a single sentence.

Details around why it works in a certain way are useful, but only relevant to developers studying the source code of this function. A better place for such comments is therefore inside the body of the function, together with the code they document.


```suggestion
```

internal/git/commands.go
```diff
@@ -980,6 +986,17 @@ func branchesQuery(querier subshelldomain.Querier) (branchesQueryResults, error)
...
+func isBareWorktreePath(path string) bool {
+	if strings.HasSuffix(path, ".git") {
+		return true
+	}
+	_, err := os.Stat(filepath.Join(path, ".git"))
```
kevgo, line 992: Adding the comment removed from the function docstring above.

```suggestion
  // Non-bare worktrees always contain either a file or folder ".git",
  // bare worktrees do not. 
	_, err := os.Stat(filepath.Join(path, ".git"))
```

internal/git/commands_test.go
```diff
@@ -405,6 +405,25 @@ func TestBackendCommands(t *testing.T) {
...
+			bareDir := t.TempDir()
+			origin.MustRun("git", "clone", "--bare", origin.WorkingDir, bareDir)
+			worktreeDir := t.TempDir()
+			origin.MustRun("git", "-C", bareDir, "worktree", "add", "-b", "feature", worktreeDir)
+			linked := testruntime.New(worktreeDir, origin.HomeDir, origin.BinDir)
+			available := asserts.NoError1(linked.Git.BranchesAvailableInCurrentWorktree(linked))
+			must.SliceContainsAll(t, gitdomain.LocalBranchNames{
+				gitdomain.LocalBranchName("initial"), // bare HEAD — must be available
+				gitdomain.LocalBranchName("feature"), // current branch in linked worktree
+			}, available)
```
kevgo: Thanks for making this test match the structure of the other tests in this file. Since you check for the existence of the `initial` branch here (which the other tests intentionally ignore), you can verify the result directly: 

```suggestion
			must.Eq(t, available, gitdomain.LocalBranchNames{
				"feature", // current branch in linked worktree
				"initial", // bare HEAD — must be available
			})
```

kevgo (commented):
Thanks for adding support for bare worktrees, that's an important feature! 

Since you use AI, can you please try to let it generate end-to-end tests? They could go next to features/branch/worktree/regular.feature.

A few other comments below. Hopefully you have the AI hooked up to GitHub so that it can read and address the comments here.

No problem if the AI can't implement any of the feedback. In that case I'll merge this PR and add the changes myself. Just let me know.

kevgo:
To fix the linter errors, can you please run `make fix`?

internal/git/commands.go
benmosher: generous to point fingers at the agent but you can blame me for this 😅  but smart to throw it out 👍

benmosher:
@kevgo I believe I've addressed all your feedback -- ready for next round of review

internal/git/commands.go
kevgo: Agree, this could probably be simplified to `strings.Split`.

cucumber-sort.json
```diff
@@ -170,5 +171,6 @@
     "^the initial tags exist now$",
     "^the initial proposals exist now$",
     "^the proposals are now$"
-  ]
+  ],
+  "unknown-steps": []
```
kevgo:
```suggestion
  ]
```

kevgo (approved):
Very well done, thanks so much! 🙏 ❤️

PR #6227, opened 2026-06-03
FIX Bitbucket breadcrumb markdown bugs

OrHayat:
Bitbucket Cloud renders Markdown via Python-Markdown and its limited compared to other providers such as github/gitlab
that uses commonmarker.

- it fails to render <sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://github.com/git-town/git-town)</sup>  it render raw text `<sup> generated by  git town</sup>` 
- it doesnt render the breadcrumb stack properly because it uses[python-markdown 451](https://github.com/Python-Markdown/markdown/issues/451) need 4 spaces to use nested list indentation properly instead of 2


kevgo:
Thanks for fixing the footer on Bitbucket Cloud, this is awesome! :pray:

PR #6251, opened 2026-07-03
Redact secrets by default in "git town config", add --show-secrets flag

### LINKED ISSUE #6250: Mask the GitHub API token in `git town config` output
## Summary

`git town config` prints the resolved `github-token` in clear text. Even though the token is stored in `.git/config` or an env var anyway, printing it in the output of a routine command meaningfully increases accidental-exposure surface — `git town config` output gets pasted into bug reports, screenshots, CI logs, and chat far more readily than someone `cat`s `.git/config`.

## Suggestion

Redact/mask the token in `git town config` output — e.g. `ghp_••••••••last4` or `<redacted>` — as many CLIs do for secrets. This is a display-only, cross-platform change and doesn't affect how the token is stored or resolved.

Could optionally pair with a flag (e.g. `--show-secrets`) to opt back into showing the full value when needed.

## Note

git-town already supports sourcing the token from `$GITHUB_TOKEN` / `$GITHUB_AUTH_TOKEN` (which avoids at-rest plaintext in `.git/config`), but `git town config` still displays the resolved value regardless of source — so masking helps in all cases.

Indemnity83:
## Summary

`git town config` printed configured secrets (tokens, Bitbucket app password, git user email) in clear text and only hid them behind the opt-in `--redact` flag. As discussed in #6250, the default is backwards: this output routinely gets pasted into bug reports, screenshots, CI logs, and chat, so it should be safe by default.

This flips the behavior:

- redact secrets as `(configured)` **by default**
- add a `--show-secrets` flag to display them unredacted when needed
- remove the now-redundant `--redact` flag

Closes #6250.

## Behavior change

- `git town config` no longer prints secret values by default — use `git town config --show-secrets` to see them.
- The `--redact` flag is removed (its behavior is now the default). Anyone scripting `--redact` should drop the flag.

## Implementation

- `formatSecret` (renamed from `formatToken`, since it guards more than tokens) now redacts by default and reveals only when `--show-secrets` is set.
- `--redact` / `configdomain.Redact` replaced by `--show-secrets` / `configdomain.ShowSecrets`.

## Follow-up (not in this PR)

A couple of fields that are arguably PII are still printed in clear text because they don't flow through `formatSecret`: **git user name** and **Bitbucket username**. Happy to fold those in here or handle separately — whichever you prefer.

## Verification

- `go test ./internal/cmd/config`
- `git town config` → secrets show as `(configured)`
- `git town config --show-secrets` → secrets shown unredacted


kevgo:
Thanks for sending this fix! :pray: 

As we are discussing in #6250, this feature to redact secrets already exists in the form of the `--redact` flag, but the default behavior is probably wrong. If you agree with that, then the fix should be:

- redact secrets by default
- introduce a `--show-secrets` flag to optionally show secrets unredacted
- remove the `--redact` flag

What do you think?

Indemnity83:
Ah. I missed that option. You're right I think: hiding should be the default behavior. I will rework this PR later today to reflect that.

Indemnity83:
Reworked per your suggestion above — thanks for steering this.

**What changed from the original approach:**

- Instead of always masking just the GitHub token via a dedicated `ConfiguredStringerSetting` helper, secrets are now redacted **by default** through the existing `formatToken` path (which already covered the other tokens + git user email).
- Added a `--show-secrets` flag to reveal them unredacted.
- Removed the `--redact` flag (its behavior is now the default), along with `configdomain.Redact` → replaced by `configdomain.ShowSecrets`.
- Renamed `formatToken` → `formatSecret` in a separate commit, since it guards more than tokens and this makes the link to `--show-secrets` clearer. It's isolated so it's easy to drop if you'd rather not have it.

Updated the reworked `redact.feature` → `secrets.feature` (default masks + `--show-secrets` reveals), added a `formatSecret` unit test, and refreshed the config-view features whose default output now shows `(configured)`.

**Follow-up items for your call:**

1. **PII still shown in clear text.** `git user name` and `Bitbucket username` don't flow through `formatSecret`, so they're not redacted. Want those folded in here, or a separate issue?
2. **`origin hostname`** can leak a private/internal Git server address — not PII, but same "ends up in pasted output" concern. Optional.

**Note on the diff:** the branch still contains the original "mask GitHub token" commits followed by the rework, so reviewing commit-by-commit shows some add-then-remove churn. The net diff is clean, and since Git Town squash-merges it collapses to a single commit anyway — happy to squash the branch down if you'd prefer a tidier commit history to
review.

kevgo:
Thanks for implementing this excellent improvement! 🙌

kevgo:

PR #6252, opened 2026-07-03
Add gh autofill for proposal creation

### LINKED ISSUE #6224: git town propose --stack fails with the gh connector in non-interactive shells (no --title/--body/--fill passed to `gh pr create`)
### Steps to reproduce

1. Use the `gh` (GitHub CLI) connector:
     ```bash
     git config git-town.github-connector gh
     ```
  2. Create a two-branch stack off trunk, each with one commit:
     ```bash
     git town hack probe-1
     printf 'x\n' > probe-1.txt && git add -A && git commit -m "probe 1"

     git town append probe-2
     printf 'y\n' > probe-2.txt && git add -A && git commit -m "probe 2"
     ```
  3. Propose the stack non-interactively (no TTY — e.g. CI or an automation/agent harness):
     ```bash
     git town propose --stack --non-interactive --verbose
     ```

  Same failure occurs with `--no-browser` instead of `--non-interactive`, and with no flag at all.

### Command output

```shell
[tmp-gt-probe-1] gh pr create --head=tmp-gt-probe-1 --base=develop
  must provide `--title` and `--body` (or `--fill` or `fill-first` or `--fillverbose`) when not running interactively
  Usage:  gh pr create [flags]
  ...
  Error: exit status 1

  To continue after having resolved conflicts, run "git town continue".
  To go back to where you started, run "git town undo".
  To continue by skipping the current branch, run "git town skip".


No proposals are created for any branch; the command stops at the bottom of the stack and leaves an interrupted state (recoverable with `git town undo`).
```

### Additional information

**Environment**
  - Git Town: `23.0.1`
  - gh (GitHub CLI): `2.93.0`
  - `git-town.github-connector`: `gh`
  - Shell: non-interactive / no TTY

  **Diagnosis**
  - Git Town invokes `gh pr create --head=<branch> --base=<parent>` with **no** `--title`, `--body`, or `--fill`. In a non-TTY shell `gh` cannot prompt, so it exits with the usage error above.
  - `--non-interactive` does **not** change the generated `gh` argv — the failure is identical with and without it.
  - For `--stack`, the `--title`/`--body` flags are single values and cannot title N branches, so they are not a usable workaround for a multi-branch stack.

  **Expected behavior** (either is fine)
  1. In a non-TTY environment with the `gh` connector, pass `gh --fill` (or derive each branch's title from its commit subject) so `propose --stack` works unattended; or
  2. Fail fast with a clear Git Town-level message (e.g. "running non-interactively: pass `--title`/`--body` or ensure the connector can auto-fill") instead of surfacing a raw `gh` usage dump mid-stack.

  **Workaround**
  Create each PR manually, base = parent branch:
  ```bash
  gh pr create --base <parent-branch> --head <phase-branch> --title "…" --body "…"

Indemnity83:
Fixes #6224

## Summary
- add `--fill` to `gh pr create` when Git Town does not provide both title and body
- keep existing behavior unchanged when both title and body are provided
- add unit coverage for all title/body combinations in the gh connector

## Verification
- `go test ./internal/forge/gh`
- `go test ./internal/forge/...`
- `git diff --check`

## Notes
- `make cuke` failed in unrelated append/lang and append/propose scenarios around localized Git output and missing browser-open commands.
- `go test ./...` failed only in the Cucumber package with unrelated append/beam scenario diffs; regular Go packages passed.

### DIFF
--- a/internal/forge/gh/connector_test.go
+++ b/internal/forge/gh/connector_test.go
@@ -4,11 +4,83 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/git-town/git-town/v23/internal/browser/browserdomain"
+	"github.com/git-town/git-town/v23/internal/cli/print"
 	"github.com/git-town/git-town/v23/internal/forge/forgedomain"
 	"github.com/git-town/git-town/v23/internal/forge/gh"
+	"github.com/git-town/git-town/v23/internal/git/gitdomain"
+	"github.com/git-town/git-town/v23/internal/gohacks/stringss"
 	. "github.com/git-town/git-town/v23/pkg/prelude"
 	"github.com/shoenig/test/must"
 )
+
+func TestConnectorCreateProposal(t *testing.T) {
+	t.Parallel()
+
+	t.Run("autofills missing title and body", func(t *testing.T) {
+		t.Parallel()
+		frontend := &recordingRunner{}
+		connector := newTestConnector(frontend)
+		args := createProposalArgs()
+
+		err := connector.CreateProposal(args)
+
+		must.NoError(t, err)
+		must.Eq(t, [][]string{
+			{"gh", "pr", "create", "--head=feature", "--base=main", "--fill"},
+			{"gh", "pr", "view"},
+		}, frontend.calls)
+	})
+
+	t.Run("autofills missing body", func(t *testing.T) {
+		t.Parallel()
+		frontend := &recordingRunner{}
+		connector := newTestConnector(frontend)
+		args := createProposalArgs()
+		args.ProposalTitle = Some(gitdomain.ProposalTitle("my title"))
+
+		err := connector.CreateProposal(args)
+
+		must.NoError(t, err)
+		must.Eq(t, [][]string{
+			{"gh", "pr", "create", "--head=feature", "--base=main", "--title=my title", "--fill"},
+			{"gh", "pr", "view"},
+		}, frontend.calls)
+	})
+
+	t.Run("autofills missing title", func(t *testing.T) {
+		t.Parallel()
+		frontend := &recordingRunner{}
+		connector := newTestConnector(frontend)
+		args := createProposalArgs()
+		args.ProposalBody = Some(gitdomain.ProposalBody("my body"))
+
+		err := connector.CreateProposal(args)
+
+		must.NoError(t, err)
+		must.Eq(t, [][]string{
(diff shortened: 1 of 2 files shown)


internal/forge/gh/connector_test.go
```diff
@@ -4,12 +4,84 @@ import (
...
 
+func TestConnectorCreateProposal(t *testing.T) {
+	t.Parallel()
+
+	t.Run("autofills missing body", func(t *testing.T) {
+		t.Parallel()
+		frontend := &recordingRunner{}
+		connector := newTestConnector(frontend)
+		args := createProposalArgs()
+		args.ProposalTitle = Some(gitdomain.ProposalTitle("my title"))
```
kevgo: Some of the test data defined here, like branch names, proposal title, and empty proposal body, is asserted later on line 31. Can we make those values explicit in the fixture?

```suggestion
		args := forgedomain.CreateProposalArgs{
		  Branch:         "feature",
		  FrontendRunner: frontend,
		  MainBranch:     "main",
		  ParentBranch:   "main",
		  ProposalBody:   None[gitdomain.ProposalBody](),
		  ProposalTitle:  Some(gitdomain.ProposalTitle("my title")),
	}
```

kevgo (approved):
Thanks for implementing this, very nicely done! 🙌

internal/forge/gh/connector_test.go
Indemnity83: Good call — done in 586b52f6e. I took a slightly different shape than your suggestion, so flagging it:

Rather than inline the full `CreateProposalArgs` struct in each test, I kept the `createProposalArgs()` helper for the constant scaffolding (`Branch`, `MainBranch`, `ParentBranch`, `FrontendRunner`) and made each test set `ProposalTitle` and `ProposalBody` explicitly — including the empty `None` cases:

```go
args := createProposalArgs()
args.ProposalTitle = Some(gitdomain.ProposalTitle("my title"))
args.ProposalBody = None[gitdomain.ProposalBody]()
```

That way the title/body values the assertions depend on (especially the empty body that triggers `--fill`) are visible in each test, without repeating the four fields that never vary. Happy to switch to fully-inlined fixtures if you'd rather have everything explicit at the call site.

kevgo (approved):

kevgo:
Thanks so much for adding this!

PR #6253, opened 2026-07-03
ship: reject parent checked out in another worktree

### LINKED ISSUE #6223: ship: exit with error if the branch to ship into (parent) is checked out in another worktree
### Steps to reproduce

- have two worktrees, one with the feature branch and one with the main branch
- run `git town ship` on the feature branch

expected behavior: don't even attempt the ship, instead exit with error message "can't ship into <parent> because it is checked out in another worktree"

actual behavior: tries to check out the parent, which fails, then exits in the middle of the ship

### Command output

```shell
fatal: 'main' is already used by worktree at '<primary-worktree-path>'
Error: cannot check out branch main: exit status 128
```

### Additional information

_No response_

Indemnity83:
Fixes #6223

## Summary
- reject ship early when the target parent branch is checked out in another worktree
- add a regression Cucumber scenario for the fast-forward ship path

## Verification
- `go test ./internal/cmd/ship ./internal/messages`
- `git diff --check`
- focused Cucumber scenario via `make cukethis`

Note: full `make cuke` was attempted and hit unrelated existing append/propose and Spanish localization failures.


kevgo (approved):
Thanks so much! ❤️

kevgo:
@Indemnity83 Thanks for this contribution!

I sent you an invite to the Git Town organization. If you accept, you'll be able to create branches and pull requests directly in this repo, which can make working on Git Town a bit easier.

No pressure either way. You're also very welcome to stay outside the org and keep contributing the way you are now.

One way or another, welcome as a Git Town contributor!

kevgo:
This is now available as part of [Git Town 24](https://github.com/git-town/git-town/releases/tag/v24.0.0). Thanks again for contributing this! 🙏

Issue #6254, opened 2026-07-05
Inconsistent commit title/body handling across ship connectors when a message is supplied


Indemnity83:
### Summary

When shipping with a supplied commit message (`-m`/`--message` or `-f`/`--message-file`), connectors disagree on whether the message's first line becomes the squash-commit **title** or whether the whole message is dumped into the **body** with the PR title left untouched. Notably, the two GitHub connector types (`api` token vs. the `gh` CLI) behave differently for the *same forge*.

Surfaced while discussing #6249.

### Behavior today

**First line → commit title (message overrides the PR title):**

- GitHub API/token — `message.Parts()` → `CommitTitle` + `Body` (`internal/forge/github/api_connector.go:99`)
- Gitea — `Title` + `Message` (`internal/forge/gitea/api_connector.go:116`)

**Whole message → body only (PR title preserved):**

- GitHub `gh` CLI — `gh pr merge --squash --body=<full message>`, no `--title` (`internal/forge/gh/connector.go:147`)
- GitLab API — `SquashCommitMessage=<full message>` (`internal/forge/gitlab/api_connector.go:100`)
- GitLab `glab` CLI — `glab mr merge --squash --squash-message=<full message>` (`internal/forge/glab/connector.go:142`)
- Bitbucket Cloud — `Message=<full message>` (`internal/forge/bitbucketcloud/api_connector.go:161`)

### Why it matters

The same `git town ship -m "..."` produces a different squash-commit title depending on which connector is active. For GitHub specifically, switching between the `api` and `gh` connector types — which users otherwise treat as interchangeable — silently changes whether their message can override the commit title. This also interacts with "PR title must match rules" CI checks.

### Suggestion

Pick one consistent contract for how a supplied message maps to title/body, and apply it across all connectors (e.g. first line = title, remainder = body everywhere, matching the GitHub-API/Gitea behavior).

kevgo:
Thanks for investigating and reporting this! The different connectors were built at different times by different people, so it's not surprising they have different behaviors. I agree it would be better if they behaved consistently, ideally in the way you propose.

PR #6256, opened 2026-07-05
Merge with the forge's commit message by default when shipping via API

### LINKED ISSUE #6249: ship-strategy=api performs a local squash-merge instead of merging through the forge API
## Summary

With `ship-strategy = api` (github connector `gh`, valid token), `git town ship` performs a **local `git merge --squash`** into the parent branch rather than merging the proposal through the forge's API. Observable consequences versus a forge-side merge:

- The PR is recorded as **closed**, not **merged** — a local squash creates a new commit, so GitHub can't SHA-match it to the proposal.
- It opens the `SQUASH_MSG` editor and **requires a commit message**, even though the forge would supply one from the repository's squash-message settings.
- It is **not** the "presses the merge button for the proposal ... via an API call" behavior the documentation describes for the api strategy — the merge itself is a local `git merge`, not a forge API call.

## Observed

Dry-run (valid token), shipping a stacked branch:

```
Finding proposal from <branch> into <parent> ... #NNN
[<parent>] git merge --squash --ff <branch>
```

The API is used to find and retarget proposals (`gh pr edit …`), but the merge step is a local squash — there is no `gh pr merge` / forge merge call.

## Expected

With the api strategy, merge the proposal through the forge API (matching the documented behavior). This records the PR as **merged** and uses the **repository's configured default squash message** (which can be set to the PR title, or title + description) with no interactive prompt. Descendant branches are then restacked exactly as git-town already does. `-m` / `-f` work today but require manually entering a message the forge would otherwise provide.

## Confirmed working alternative (forge-side merge + normal restack)

The desired outcome is reproducible today with `gh`, and it behaves correctly — the shipped PR records as **merged**, the squash subject comes from the repo's squash-title setting, and stacked children restack with no conflicts. For a stack `main ← A ← B`, shipping `A`:

```bash
# 1. Forge-side squash-merge: records A as "merged" and uses the repo's squash-title
#    setting for the message. NO --delete-branch — keep A as the rebase boundary.
gh pr merge A --squash

# 2. Retarget the child, then rebase it onto the new main, dropping A's now-squashed
#    commits (<A-old-tip> is the parent tip B was based on).
git fetch origin
gh pr edit B --base main
git rebase --onto origin/main <A-old-tip> B
git push --force-with-lease origin B

# 3. Only now delete A's branch.
gh api --method DELETE repos/<owner>/<repo>/git/refs/heads/A
```

Step 2 is exactly the restack `git town ship` already performs. The only change from git-town's current behavior is **step 1** — merge through the forge API instead of a local `git merge --squash`, and defer branch deletion until after the restack. Verified on a real 3-branch stack: the shipped PR shows **merged**, the child retargets and rebases cleanly, and no interactive commit message is needed.

## Note

This supersedes the issue's original "invalid-token silent fallback" framing. Further investigation showed the local squash + message prompt happens **regardless of token validity** — it is inherent to the ship path, not a token-triggered fallback. The descendant restack is unaffected by the merge mechanism, so this is specifically about how the merge is performed.

## Environment

- Git Town 23.0.3
- macOS (Darwin 25.5.0)
- forge: github, `github-connector: gh`
- linked worktree

Indemnity83:
## ⚠️ BREAKING CHANGE

`git town ship` with the [`api` ship-strategy](https://www.git-town.com/preferences/ship-strategy.html) now merges the proposal using **the forge's own commit message** (the same message you'd get clicking "merge" in the web UI), instead of opening an editor for you to write one.

**Migration** — to keep the old behavior (enter the squash commit message yourself):

- per-invocation: `git town ship --enter-message`
- persistently: set `[ship] enter-message = true` (or `git config git-town.ship-enter-message true`)

This only affects the case where **no** message was supplied. `git town ship -m "…"` / `-f <file>` are unchanged.

## Why

As discussed in #6249, the previous default forced an interactive editor whenever you shipped without `-m`, which:

- can't run headless (CI, scripts, agentic loops), and
- makes you re-type a commit message even though "merge via the forge API" implies the forge should pick it — undercutting e.g. "PR title must match rules" checks.

"Merge with the forge's message, prompt only when asked" is the more intuitive default; the old behavior is preserved behind an explicit opt-in.

## What changed

- `git town ship` (no flags, `ship-strategy=api`) → merges via the forge API using the forge's commit message. No local squash commit, no editor.
- `git town ship --enter-message` → the previous behavior: a local squash merge lets you edit the message, which is then used for the API merge.
- New `[ship] enter-message` setting (config file, `git-town.ship-enter-message` git metadata, `GIT_TOWN_SHIP_ENTER_MESSAGE` env var), shown in `git town config`.
- `ProposalMerger.SquashMergeProposal` now takes an `Option[CommitMessage]`; `None` means "let the forge choose the message" (each connector omits the title/body/message it would otherwise send).

## Tests

- Cucumber (`features/ship/api/enter_message.feature`): the default path merges with no editor dance; `--enter-message` performs the squash/commit/reset dance before merging.
- Added `SquashMergeProposal` to the github mock connector so `ship-strategy=api` can complete a merge in end-to-end tests (it previously wasn't a `ProposalMerger`, so every `api/` scenario was an error case).
- gh connector unit test: `None` omits `--body`, `Some` includes it.

## Notes / follow-ups

- The mapping of a supplied `-m` message to commit **title vs. body** is left **unchanged per connector** in this PR — it matches today's behavior, which is inconsistent across connectors (e.g. the GitHub API connector splits first-line→title while the `gh` connector puts everything in `--body`). That inconsistency is tracked separately in #6254.

Closes #6249.


features/ship/api/enter_message.feature
```diff
@@ -0,0 +1,50 @@
...
+      | feature | git fetch --prune --tags                                          |
+      |         | Finding proposal from feature into main ... #1 (feature proposal) |
+      |         | git checkout main                                                 |
+      |         | GitHub API: merging PR #1 ... ok                                  |
+      | main    | git push origin :feature                                          |
+      |         | git branch -D feature                                             |
+    And the branches are now
+      | REPOSITORY    | BRANCHES |
+      | local, origin | main     |
+    And the initial proposals exist now
```
kevgo, line 31: Proposal #1 should technically be gone by this point. As written, it documents behavior that isn't actually correct.

This currently passes because the end-to-end tests run the Git Town executable in [test mode](https://github.com/git-town/git-town/blob/main/internal/subshell/doc.go#L17-L20). In test mode, Git Town uses the [mock GitHub connector](https://github.com/git-town/git-town/blob/main/internal/forge/github/mock_connector.go#L22) instead of the [real GitHub connector](https://github.com/git-town/git-town/blob/main/internal/forge/github/api_connector.go). 

The mock connector doesn't perform any API operations on github.com. It only prints output to the CLI.

We could update the mock connector to perform the actual squash-merge on the origin repo and delete the proposal, and then have this end-to-end test verify that the proposal is gone. But at that point, the test would mostly verify and document how the mock connector works, and not how the actual forge works. That doesn't seem useful.

```suggestion
```

internal/cli/flags/enter_message.go
```diff
@@ -0,0 +1,23 @@
...
+	. "github.com/git-town/git-town/v23/pkg/prelude"
+	"github.com/spf13/cobra"
+)
+
+const enterMessageLong = "enter-message"
+
+// type-safe access to the CLI arguments of type configdomain.ShipEnterMessage
+func EnterMessage() (AddFunc, ReadEnterMessageFlagFunc) {
+	addFlag := func(cmd *cobra.Command) {
+		cmd.Flags().Bool(enterMessageLong, false, "enter the commit message to use when merging the proposal")
```
kevgo: Let's keep this concise so that it fits within a single line on an 80 character terminal:

```suggestion
		cmd.Flags().Bool(enterMessageLong, false, "manually enter the commit message")
```

internal/forge/bitbucketcloud/api_connector.go
```diff
@@ -158,17 +158,21 @@ func (self APIConnector) SearchProposals(branch gitdomain.LocalBranchName) ([]fo
...
 		ID:       number.String(),
 		Owner:    self.Organization,
 		RepoSlug: self.Repository,
-		Message:  message.String(),
-	})
+	}
+	// When no commit message is given, Bitbucket determines the squash commit message.
+	if commitMessage, hasCommitMessage := message.Get(); hasCommitMessage {
+		options.Message = commitMessage.String()
+	}
```
kevgo, line 171: If we don't set the Message field explicitly, it still exists on the options struct and defaults to its zero value, which is an empty string.

So we can simplify this code:

```suggestion
	options := &bitbucket.PullRequestsOptions{
		ID:       number.String(),
		Owner:    self.Organization,
		RepoSlug: self.Repository,
		Message:  message.GetOrZero().String(),
	}
```

internal/forge/forgejo/api_connector.go
```diff
@@ -113,20 +113,26 @@ func (self *APIConnector) SearchProposals(branch gitdomain.LocalBranchName) ([]f
...
 		return errors.New(messages.ProposalNoNumberGiven)
 	}
-	commitMessageParts := message.Parts()
+	// When no commit message is given, Forgejo determines the squash commit message.
+	var commitTitle, commitBody string
+	if commitMessage, hasCommitMessage := message.Get(); hasCommitMessage {
+		commitMessageParts := commitMessage.Parts()
+		commitTitle = commitMessageParts.Title.String()
+		commitBody = commitMessageParts.Body
+	}
```
kevgo: Suggestion to simplify this. If you agree, you can restore lines 134-135.

```suggestion
commitMessageParts := message.GetOrZero().Parts()
```

internal/forge/gh/connector_test.go
```diff
@@ -14,6 +14,28 @@ import (
...
+func TestConnectorSquashMergeProposal(t *testing.T) {
+	t.Parallel()
+
+	t.Run("with a commit message sends it as the body", func(t *testing.T) {
+		t.Parallel()
+		runner := &recordingRunner{}
+		connector := gh.Connector{Frontend: runner}
+		err := connector.SquashMergeProposal(forgedomain.ProposalNumber(1), Some(gitdomain.CommitMessage("my message")))
+		must.NoError(t, err)
+		must.Eq(t, [][]string{{"gh", "pr", "merge", "--squash", "--body=my message", "1"}}, runner.calls)
```
kevgo, line 26: These are really useful tests. Thanks for adding them!

internal/forge/gitlab/api_connector.go
```diff
@@ -97,18 +97,23 @@ func (self APIConnector) SearchProposals(branch gitdomain.LocalBranchName) ([]fo
...
+func (self APIConnector) SquashMergeProposal(number forgedomain.ProposalNumber, message Option[gitdomain.CommitMessage]) error {
 	if number <= 0 {
 		return errors.New(messages.ProposalNoNumberGiven)
 	}
 	self.log.Start(messages.ForgeGitlabMergingViaAPI, number)
-	// the GitLab API wants the full commit message in the body
-	_, _, err := self.client.MergeRequests.AcceptMergeRequest(self.projectPath(), number.Int(), &gitlab.AcceptMergeRequestOptions{
-		SquashCommitMessage: new(message.String()),
-		Squash:              new(true),
+	options := &gitlab.AcceptMergeRequestOptions{
```
kevgo, line 105: You could use `message.GetOrZero()` here as well to simplify the code.

kevgo (approved):
Thanks for implementing this long overdue improvement of the default behavior! I only have a few smaller improvements.

features/ship/api/enter_message.feature
Indemnity83: Good catch on the semantics, but I hit a wall dropping the assertion outright: the Cucumber harness hard-exits when a scenario declares `And the proposals` but never verifies them ([`steps.go:87-90`](https://github.com/git-town/git-town/blob/main/internal/test/cucumber/steps.go#L87-L90) — `initialProposals.IsSome() && !proposalsChecked`). Since `ship` needs the proposal to exist in order to find and merge it, I can't remove the Background declaration either, so *some* proposal check is mandatory here.

That leaves two options:

1. Keep `And the initial proposals exist now` (what I've done for now). It's consistent with the rest of the suite, at the cost of documenting the mock's no-op behavior rather than the real forge's.
2. Have the mock connector's `SquashMergeProposal` actually remove the merged proposal, then assert the proposal is gone — i.e. the approach you mentioned. It'd make the assertion truthful, but as you noted it mostly documents the mock.

I've left (1) in place. Do you want me to go with (2) instead, or is keeping the existing assertion acceptable given the harness constraint?

internal/forge/gitlab/api_connector.go
Indemnity83: I applied `GetOrZero()` to the GitHub, Gitea, Forgejo, and Bitbucket connectors, but I'd like to leave this GitLab one as-is if that's alright.

Those other connectors pass the message into a plain `string` field, so `None` already results in an empty string being sent — `GetOrZero()` is a true no-op there. GitLab is different: `SquashCommitMessage` is a `*string`, so the current conditional leaves it `nil` (omitted from the request) when no message is given, whereas `GetOrZero()` would send an explicit `&""`.

Even if GitLab treats an empty `squash_commit_message` the same as an omitted one, that's an assumption about server-side behavior we're not testing, and this PR's whole point is that "no message given" should let the forge choose. Sending `nil` keeps that intent explicit and avoids an on-the-wire API change for a three-line savings. Happy to apply it if you'd still prefer the consistency — just wanted to flag the semantic difference first.

kevgo (approved):

internal/forge/gitlab/api_connector.go
kevgo: Sounds good, thanks for looking into it! 🙏

kevgo:
This is now available in [Git Town 24](https://github.com/git-town/git-town/releases/tag/v24.0.0). It took a while to release because this was a breaking change, and we try to batch planned breaking changes since we only bump the major version every few months. Thanks again for contributing this! 🙏

PR #6259, opened 2026-07-16
Migrate the Bitbucket integration from app passwords to API tokens

mdwint:
[Bitbucket Cloud app passwords are end-of-life and no longer accepted](https://www.atlassian.com/blog/bitbucket/bitbucket-cloud-transitions-to-api-tokens-enhancing-security-with-app-password-deprecation). This replaces the "git-town.bitbucket-app-password" setting with "git-town.bitbucket-api-token":

- automatically upgrade the deprecated setting in local and global Git config
- read GIT_TOWN_BITBUCKET_API_TOKEN, with the old env var as fallback for Bitbucket Data Center users whose HTTP access tokens remain valid
- update the setup assistant to ask for an API token with scopes and clarify that Bitbucket Cloud API tokens authenticate with the Atlassian account email as username
- document how to create a scoped API token and redirect the old documentation URL


kevgo:
Thanks for sending this, it's an important and well-executed change! 🙏

One thing to keep in mind is that Git Town also supports Bitbucket Data Center, down to pretty old versions. Will the new API tokens work there as well?

It looks like we may need to support both authentication methods:

* Bitbucket Cloud: username and API token
* Bitbucket Data Center: username and app password

Given that, I think it would be safer to keep the existing Bitbucket app-password setting and add a separate setting for the new Bitbucket API token.

The setup assistant then asks for the right credential type based on the identified forge type.

I'm suspicious of the automatic migrating of existing app password to the new API-token setting. My understanding is that API tokens and app passwords are different credentials, so reusing the old app password as the new API token would leave users with a non-working configuration.

Curious what you think.

mdwint:
Thanks for your feedback. 

> One thing to keep in mind is that Git Town also supports Bitbucket Data Center, down to pretty old versions. Will the new API tokens work there as well?

No, they won't work there. API tokens from id.atlassian.com only authenticate against Bitbucket Cloud. But the commit doesn't force them on Data Center users. Both connectors treat the setting as an opaque secret sent via HTTP basic auth (same as before). Data Center users keep using whatever they had in the old setting, just under a new name.

> It looks like we may need to support both authentication methods

Bitbucket Data Center never supported app passwords, those are a Cloud-only feature. Data Center users authenticate with HTTP access tokens (or account password on very old versions), which they were already storing in the misleadingly-named bitbucket-app-password setting. So the situation is:

- Bitbucket Cloud: username (email) + API token
- Bitbucket Data Center: username + HTTP access token

> I'm suspicious of the automatic migrating of existing app password to the new API-token setting. My understanding is that API tokens and app passwords are different credentials, so reusing the old app password as the new API token would leave users with a non-working configuration.

Right, passwords and API tokens are different credentials, but Bitbucket Cloud stopped accepting app passwords entirely on June 9, 2026 (that's what motivated this change). So for Cloud users the old value is dead no matter which key it sits under; migrating it doesn't break a working config, and they'll hit a 401 until they create an API token either way.

For Data Center users the opposite holds: their HTTP access tokens still work, so dropping the value during migration would break them. Since the migration can't tell Cloud from Data Center (a global config value can span repos on different forges), carrying the value over is the choice that breaks nobody who wasn't already broken. The env-var fallback (`GIT_TOWN_BITBUCKET_APP_PASSWORD`) follows the same logic.

> The setup assistant then asks for the right credential type based on the identified forge type.

Right, there _is_ a gap here. The setup assistant shows the same dialog for both forge types, and its help text points to id.atlassian.com and mentions "API token with scopes", which is wrong for a Data Center user. We could show Data Center-specific help text (HTTP access token, created in the instance's own UI) there. That gets you the "ask for the right credential type based on the identified forge type" behavior you described, without splitting the setting in two. We could also make them two settings, but that's redundant IMO, since forge type already tells us what the credentials is for. What do you think?

mdwint:
I added the Data-Center-specific help text. Let me know if you like this or if it needs adjusting.

Also did a small refactor to eliminate the duplication of forge connector arguments.

internal/execute/new_connector.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/git-town/git-town/v23/internal/cli/print"
+	"github.com/git-town/git-town/v23/internal/forge"
+	"github.com/git-town/git-town/v23/internal/forge/forgedomain"
+	"github.com/git-town/git-town/v23/internal/git/giturl"
+	. "github.com/git-town/git-town/v23/pkg/prelude"
+)
+
+// NewConnector provides the forge connector for this repo and the given remote URL.
+func (self OpenRepoResult) NewConnector(remoteURL Option[giturl.Parts]) (Option[forgedomain.Connector], Option[forgedomain.DetectedForgeType], error) {
```
kevgo, line 12: Creating this method is a pretty good idea. Thanks for cleaning up this repetitive code in all the commands!

kevgo (approved):

kevgo:
Thanks for the background, and the very nice implementation! Makes sense to me.

kevgo:
Can you please look at the failing `linters` CI job? The fix could be as easy as running `make fix` to sort the `And local Git setting` Cucumber steps alphabetically. Then this is good to merge. Thanks!

mdwint:
> Can you please look at the failing `linters` CI job? The fix could be as easy as running `make fix` to sort the `And local Git setting` Cucumber steps alphabetically. Then this is good to merge. Thanks!

Done!

kevgo:
Thanks for this!

kevgo:

Issue #4900, opened 2025-05-09
set-parent (changing parent branch in general) overwrites the title incorrectly in bitbucket PR for corresponding proposal


mw00120:
I have branch `bugfix/explicit-view-access-checks` which is based on `feature/upgrade-to-dotnet-9` which in turn is based on master.

I open a proposal which creates a PR in bitbucket from `bugfix/explicit-view-access-checks` -> `feature/upgrade-to-dotnet-9`.

Someone merges `feature/upgrade-to-dotnet-9` to master, but forgets to click the checkbox that retargets my PR to master. Now I go to my repo, call `set-parent master` on my bugfix branch and get the following output:

```
git town set-parent

Looking for proposal online ... #704
Selected parent branch for "bugfix/explicit-view-access-checks": master

Updating target branch of proposal #704 to master ... ok
```

However I go to bitbucket and my PR's title has been set to an empty text, the barring below meaning the title has been erased.

> Malcolm Watt edited the title: ~~Make view permission checks explicit~~ 9 minutes ago

I think the title should be left alone by these types of operations if possible, hopefully it's not a limitation of the bibucket API.. It seems to me like git town is probably using a PUT endpoint instead of a PATCH (which is _hopefully_ availlable).

This impacts the UX particularly badly in bitbucket because the clickable link in the pull requests list is _on the title text_ so people can't click directly on the PR untilI have to go manually change the title back. To get there I need to navigate to the PR using the pull request number.. Obviously this UX particularity is nothing to do with git town, but it can still serve as a motivation to fix this issue

mw00120:
I wonder if this should call get -> merge -> update instead of update directly? https://github.com/git-town/git-town/blob/6b322663e4896d3dffe46b1e6f54a70f448fbb62/internal/forge/bitbucketcloud/connector.go#L273

I don't actually know how to do that b/c I don't know go.. but hopefully that's useful analysis of the problem

kevgo:
Thanks for reaching out about this. Sounds like a bug in our Bitbucket driver. Which is pretty new, so there might be bugs.

fivemhostingserver:
This is a well-structured codebase. I appreciate the clean separation of concerns.

fivemserverhostings:
This is a well-structured codebase. I appreciate the clean separation of concerns.

fivemeyestore:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

scriptforfivem:
[...]

PR #6276, opened 2026-08-12
Fix reviewers getting removed on breadcrumb update and browser open command bitbucket

mw00120:
Hello! Thanks for maitaining git-town, it's an awesome tool, I use it all the time 👍 

Some context on my setup: I use windows and bitbucket cloud

I had two painpoints which were causing issues

1. When I would do git town propose on a stacked PR (i.e. not against main), the "browser open" would fail out saying `dest` is not a recognized program: (It tries to open `<url>&dest=target-branch`). This issue has existed forever, but I figured I would try to fix it since my second problem was much more serious.
2. I recently upgraded to 24.0 (from something ~21, there are some great new features!) and enabled the `breadcrumb` feature, which works fine however when I would do a sync on a stack of PRs, all my reviewers would get removed :( 

I'm not really a go programmer, so I tried some debugging and fixing with AI and was able to get both pain-points resolve with what seem like reasonable changes... It was having a lot of trouble figuring out why the escape of the `&` didn't work and decided the easiest fix was to use:

```
rundll32 url.dll,FileProtocolHandler url
```

Changing the program used here seems odd to me, but apparently this is equivalent to asking the OS for the "configured browser" and opening using that, and ships in all windows installs... If you think this is the wrong approach I can try to make the actual escaping work instead..

As for wiping the reviewer, it ended up being kind of similar to https://github.com/git-town/git-town/issues/4900 where the body being sent in the PUT to update the PR did not have all the information from the PR loaded into it and the PUT defaults to clearing out the fields that are unspecified.. The solution was to load the particular PR's data and populate the PUT accordingly!

I have a build of this working on my machine, and I haven't seen any other regressions so far.. Let me know if you have any feedback I'm happy to help make it work for other windows and bitbucket cloud users (maybe only me 😆 )

### DIFF
--- a/vendor/github.com/ktrysmt/go-bitbucket/pullrequests.go
+++ b/vendor/github.com/ktrysmt/go-bitbucket/pullrequests.go
@@ -218,15 +218,15 @@ func (p *PullRequests) Statuses(po *PullRequestsOptions) (interface{}, error) {
 
 func (p *PullRequests) buildPullRequestBody(po *PullRequestsOptions) (string, error) {
 	body := map[string]interface{}{}
-	body["source"] = map[string]interface{}{}
-	body["destination"] = map[string]interface{}{}
-	body["reviewers"] = []map[string]string{}
-	body["title"] = ""
-	body["description"] = ""
-	body["message"] = ""
-	body["close_source_branch"] = false
-
-	if n := len(po.Reviewers); n > 0 {
+	if n := len(po.ReviewerAccountIDs); n > 0 {
+		body["reviewers"] = make([]map[string]string, n)
+		for i, accountID := range po.ReviewerAccountIDs {
+			body["reviewers"].([]map[string]string)[i] = map[string]string{
+				"type":       "user",
+				"account_id": accountID,
+			}
+		}
+	} else if n := len(po.Reviewers); n > 0 {
 		body["reviewers"] = make([]map[string]string, n)
 		for i, uuid := range po.Reviewers {
 			body["reviewers"].([]map[string]string)[i] = map[string]string{"uuid": uuid}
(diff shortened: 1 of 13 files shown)


mw00120:
Actually I just found an edge case I think

When running `git town delete` on the branch I was using to test, I get the following log to finish:

```
Update body for #1683 ... 400 Bad Request
```

and the proposal isn't removed.. I will see if that's related to my changes or already in 24.0

kevgo:
Thanks for fixing these issues! This PR addresses several bugs. Since the changes are fairly complex and I’d like to verify each fix independently before merging, could you please submit each bug fix on a separate branch?

My Windows laptop just died, so I’ll need to wait a few days until I get a replacement before I can properly review and test these changes.

For bug 1, I’m hoping we can fix it by encoding the URL we pass to the browser more carefully, or possibly by escaping the special characters in it. If that's not possible, we can go with your solution.

Bug 2 looks like a legitimate bug in our Bitbucket connector.

mw00120:
Yep, no problem! Will try to do that soon here.. since you don't have the resources to test right now, I will check that out at the start of next week :) 

I agree with your take on the encoding issue.. I will see if I can fix that

I have marked this PR as a draft so I don't forget, and I will close it once I open the other ones

vendor/github.com/ktrysmt/go-bitbucket/pullrequests.go
```diff
@@ -218,35 +218,49 @@ func (p *PullRequests) Statuses(po *PullRequestsOptions) (interface{}, error) {
...
-	body["destination"] = map[string]interface{}{}
-	body["reviewers"] = []map[string]string{}
-	body["title"] = ""
-	body["description"] = ""
-	body["message"] = ""
-	body["close_source_branch"] = false
-
-	if n := len(po.Reviewers); n > 0 {
+	if n := len(po.ReviewerAccountIDs); n > 0 {
+		body["reviewers"] = make([]map[string]string, n)
```
mw00120, line 222: oh I didn`t even see that this is a vendored package.. do you want me to submit a patch there or are you ok with me patching directly here @kevgo

mw00120:
closed in favor of two targetted PRs

vendor/github.com/ktrysmt/go-bitbucket/pullrequests.go
kevgo: @mw00120 whoops I missed that too... Let's do the right thing and submit the patch at the upstream repo. I update vendored dependencies regularly, so any patches to the vendored libraries will get lost. Thanks!

PR #6282, opened 2026-08-19
Browser opening bug windows

mw00120:
Only the browser opening fixes for windows from https://github.com/git-town/git-town/pull/6276

The smaller and more obvious of the two changes! It ended up being possible to do escaping correctly like you suggested in the other PR, leading to a much smaller change :) 

Tested that this correctly opens proposal in bitbucket cloud

### DIFF
--- a/internal/subshell/frontend_runner.go
+++ b/internal/subshell/frontend_runner.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
-	"runtime"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


internal/subshell/frontend_runner.go
```diff
@@ -90,7 +90,14 @@ func (self *FrontendRunner) execute(env []string, cmd string, args ...string) er
 		PrintCommand(location, self.PrintBranchNames, env, cmd, args...)
 	}
 	if runtime.GOOS == "windows" && cmd == "start" {
-		args = append([]string{"/C", cmd}, args...)
+		// "cmd /C start <url>" causes CMD to re-parse everything after "/C" as a raw
```
mw00120, line 93: I think you'll like this much better @kevgo

PR #6287, opened 2026-09-12
diff-parent: don't print the executed Git commands when --name-only is given

### LINKED ISSUE #6284: diff-parent: if --name-only option is given, don't print the Git commands executed
Currently this happens:

```
~/d/tricorder [kg-pitstop-only-branch-changes] > git town diff-parent --name-only --non-interactive

[kg-pitstop-only-branch-changes] git diff --name-only --merge-base main kg-pitstop-only-branch-changes
file1.md
Cargo.lock
Cargo.toml
```

The `[branch] git command` line makes it hard for scripts to parse this.

It would be better to not print it if the --name-only option is given.

```
~/d/tricorder [kg-pitstop-only-branch-changes] > git town diff-parent --name-only --non-interactive

file1.md
Cargo.lock
Cargo.toml
```

LouisDeconinck:
## What

`diff-parent --name-only` now suppresses the `[branch] git diff ...` echo, so scripts can parse the file list directly.

## How

`execute.OpenRepo` already controls command/branch echo via `PrintCommands`/`PrintBranchNames` (the same mechanism `config get-parent` uses for machine-readable output). Both are now gated on `--name-only`, matching how the flag's existing doc describes the output.

The Cucumber scenarios in `features/diff_parent/name_only/enabled.feature` are updated to assert `runs no commands` instead of the echo table; the file-list assertions are unchanged.

## Testing

- Built binary in a scratch repo: `--name-only` prints exactly the file list; without the flag and with `--name-only=false` the echo is preserved; `--verbose` still prints backend queries.
- `go test ./internal/cmd/... ./internal/subshell/... ./internal/git/...` — pass
- `go test -- features/diff_parent` — 26/26 scenarios pass
- `gofmt -l`, `go vet` clean

Resolves #6284


kevgo (commented):
Hi @LouisDeconinck thanks so much for implementing this! 🙏 Great that you found a way to implement the improvement in such an elegant way! 

If you address the two linter messages, then this is good to go.

internal/cmd/diff_parent.go
```diff
@@ -80,8 +80,8 @@ Start:
 	repo, err := execute.OpenRepo(execute.OpenRepoArgs{
 		CliConfig:        cliConfig,
 		IgnoreUnknown:    false,
-		PrintBranchNames: true,
-		PrintCommands:    true,
+		PrintBranchNames: nameOnly.GetOr(false) == false,
```
kevgo, line 83:
```suggestion
		PrintBranchNames: !nameOnly.GetOr(false),
```

internal/cmd/diff_parent.go
```diff
@@ -80,8 +80,8 @@ Start:
...
+		PrintCommands:    nameOnly.GetOr(false) == false,
```
kevgo, line 84:
```suggestion
		PrintCommands:    !nameOnly.GetOr(false),
```

internal/cmd/diff_parent.go
```diff
@@ -80,8 +80,8 @@ Start:
 	repo, err := execute.OpenRepo(execute.OpenRepoArgs{
 		CliConfig:        cliConfig,
 		IgnoreUnknown:    false,
-		PrintBranchNames: true,
-		PrintCommands:    true,
+		PrintBranchNames: !nameOnly.GetOr(false),
```
kevgo:
```suggestion
		PrintBranchNames: nameOnly.GetOr(false) == false,
```

internal/cmd/diff_parent.go
```diff
@@ -80,8 +80,8 @@ Start:
 	repo, err := execute.OpenRepo(execute.OpenRepoArgs{
 		CliConfig:        cliConfig,
 		IgnoreUnknown:    false,
-		PrintBranchNames: true,
-		PrintCommands:    true,
+		PrintBranchNames: nameOnly.GetOr(false) == false,
+		PrintCommands:    !nameOnly.GetOr(false),
```
kevgo:
```suggestion
		PrintCommands:    nameOnly.GetOr(false) == false,
```

kevgo:
The linter warning was a false positive. Merging this in. Thanks again for this!

